PR #1, opened 2020-10-15
MTU testing

netoptimizer:
Simple TC-BPF program for testing different packets sizes that violate the MTU of the interface.

This is part of testing upstream kernel work, for removing the MTU limit in the BPF-helpers that change packet size, and instead add BPF-helper that can check or lookup interface MTU.


netoptimizer:
Going to merge this, lets see if I get around to writing a script for easier reload.

PR #4, opened 2020-12-22
MTU-tests: Add some external use of BPF-helper bpf_check_mtu

netoptimizer:
I'm developing new BPF-helper bpf_check_mtu() for the kernel, but it's not accepted upstream (yet).

Simple test showing how I can still test this, by manually defining the call signature.

Unsure how to handle when this gets accepted upstream, as there will be a compile conflict once the BPF-helper signature becomes avail via updating libbpf (via git-submodule).

### DIFF
--- a/MTU-tests/Makefile
+++ b/MTU-tests/Makefile
@@ -2,6 +2,7 @@
 
 USER_TARGETS :=
 BPF_TARGETS := tc_bpf_inc_pkt_size
+BPF_TARGETS += tc_mtu_enforce
 EXTRA_DEPS := encap.h
 
 LIB_DIR = ../lib
--- a/MTU-tests/tc_mtu_enforce.c
+++ b/MTU-tests/tc_mtu_enforce.c
@@ -1,0 +1,34 @@
+/* TC-BPF program that enforce MTU based on BPF-helper
+ */
+#include <linux/bpf.h>
+#include <bpf/bpf_helpers.h>
+
+/* Testing BPF-helper that is not upstream yet, requires us to define
+ * its call signature manually, which will only match our devel kernel.
+ *
+ * Generated signatur via command in kernel devel tree:
+ *   ./scripts/bpf_helpers_doc.py --header | grep bpf_check_mtu
+ */
+static int (*bpf_check_mtu)(void *ctx, __u32 ifindex, __u32 *mtu_len, __s32 len_diff, __u64 flags) = (void *) 163;
+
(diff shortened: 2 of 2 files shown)


tohojo:
./configure type check and -D when compiling? Or just manually make sure you remove the local def before updating the libbpf submodule? :)

netoptimizer:
I like the ./configure check idea and a define HAVE_HELPER_XXX that gets define in an include file, but it should be per example subdir to make this more obvious to people seeing this.

I also wonder if 'bpftool features' could be used?

tohojo:
Jesper Dangaard Brouer <notifications@github.com> writes:

> I like the ./configure check idea and a define HAVE_HELPER_XXX that
> gets define in an include file, but it should be per example subdir to
> make this more obvious to people seeing this.

We have one in xdp-tools:


```
check_perf_consume()
{
    cat >$TMPDIR/libbpftest.c <<EOF
#include <bpf/libbpf.h>
int main(int argc, char **argv) {
    perf_buffer__consume(NULL);
    return 0;
}
EOF
    libbpf_err=$($CC -o $TMPDIR/libbpftest $TMPDIR/libbpftest.c $LIBBPF_CFLAGS $LIBBPF_LDLIBS 2>&1)
    if [ "$?" -eq "0" ]; then
        echo "HAVE_LIBBPF_PERF_BUFFER__CONSUME:=y" >>"$CONFIG"
        echo "yes"
    else
        echo "HAVE_LIBBPF_PERF_BUFFER__CONSUME:=n" >>"$CONFIG"
        echo "no"
    fi
}
```

> I also wonder if 'bpftool features' could be used?

I think that probes the kernel, which is probably not what you want...

-Toke

netoptimizer:
> Toke writes:
> > I also wonder if 'bpftool features' could be used?
>
> I think that probes the kernel, which is probably not what you want...

You are right.
I have booted a kernel with (devel version of the) BPF-helper, but it is the libbpf version of `bpf/bpf_helpers.h` include file that is missing the BPF-helper signature. That is what is missing for my compile to work...

For others this header file is "installed" here when building libbpf git-submodule:
`./lib/libbpf/src/root/usr/include/bpf/bpf_helpers.h`

netoptimizer:
> Or just manually make sure you remove the local def before updating the libbpf submodule? :)

I will go with the option of taking care of this when updating libbpf submodule.
I need to move forward, and will merge

PR #7, opened 2021-01-12
Add pping

simosund:
This is the initial version of pping that I showed @tohojo and @netoptimizer yesterday (but with a little bit of cleanup). It only supports IPv4 at the moment (IPv6 support is on the TODO list).

Regarding the issue you saw with it reporting RTTs of around 100ms when adding just 30ms latency (on the veth0 interface in the test environment), I'm not sure what's causing it. After the first couple of packets it seems to approach 100ms for any delay <100ms, and the correct value for delays >100ms. I get the same issue when running [Pollere's pping](https://github.com/pollere/pping) as well as my [bpftrace pping](https://github.com/simosund/bpftrace_experiments/blob/main/network/pping.bt) implementation.


tohojo:
A few high-level comments:

- You should be able to use the existing functionality to reuse maps in libbpf instead of rolling your own. To do this, you'll need to switch to BTF-defined maps; then you can set the pinning attribute in the map definition, and all you'll have to do is set the `pin_root_path` attribute in `bpf_object_open_opts` when calling `bpf_object__open_file()`. See the XDP code file in the `traffic-pacing-edt` folder for an example of the BTF map definition format; you'll need to add `__uint(pinning, LIBBPF_PIN_BY_NAME);` to that to enable pinning.
- Please reformat the C code to the kernel style. See https://www.kernel.org/doc/html/latest/process/coding-style.html - there's a `.clang-format` file in the kernel source tree that may help; Emacs also has a `linux` C style mode.
- For loading XDP programs, you should probably use libxdp (in https://github.com/xdp-project/xdp-tools). That way this will work alongside other XDP programs, and for this use, loading the program should be quite simple as well. I guess we should add some code to this repository to make it simple to include libxdp; will look into that (so feel free to wait for that before switching).
- Before we merge this, you should rewrite the history to fix up the commit messages a bit. At a minimum, you should add a Signed-off-by tag by each commit to signify that you are submitting all this as open source (see https://developercertificate.org/). But if you want to improve commit messages in general and/or squash commits together to form logical units, that's totally fine as well. In general, getting into the habit of writing commits in the style where the first line is a short heading (that can be displayed in one line of 80-100 characters), and more text is in a separate paragraph below, is a good idea (notice how GitHub also tries to display commits like that but fails because you put the entire commit message on a single line in some places).
- Please add SPDX license identifier tags to the source files that don't have it.

simosund:
Thank you for your comments @tohojo. I will probably be busy with some other things the next couple of days, but after that I'll get back to addressing these points.

tohojo:
Alright, that's totally fine of course. You can just force-push to the existing branch to update the PR, but please leave a comment as well when you do that - Github doesn't always send out notifications on branch updates...

simosund:
Hi @tohojo,
I've made the switch over to BTF-defined maps now (for the XDP-program at least), and it is kinda working. I am however having issues getting the map fd from userspace (which I need for the periodic cleanup of the map). After just opening the object, the maps do not seem to have gotten a valid file descriptor yet (-1), and after loading the object, the name of the ts_start map disappears for some reason. This is the output I get from running the program currently (as of commit [c4b75a9](https://github.com/xdp-project/bpf-examples/pull/7/commits/c4b75a91c956e65fee416853a1d17d84c307122b)):
   
```
Maps after opening object
Map name: ts_start, fd: -1
Map name: rtt_events, fd: -1
 
Maps after loading object
Map name: , fd: 5
Map name: rtt_events, fd: 6
```

Any advise on how to handle this? My current workaround is to simply take the fd of the first map from the object, but that does not feel like a very good long term solution.

pping/pping.c
```diff
@@ -74,82 +74,16 @@ static int set_rlimit(long int lim)
...
 
 static int bpf_obj_open(struct bpf_object **obj, const char *obj_path,
-			enum bpf_prog_type prog_type)
+			char *map_path)
 {
-	struct bpf_object_open_attr attr = {
-		.prog_type = prog_type,
-		.file = obj_path,
+	struct bpf_object_open_opts opts = {
+		.sz = sizeof(struct bpf_object_open_opts),
```
tohojo, line 80: There's a macro for this:
```
DECLARE_LIBBPF_OPTS(bpf_object_open_opts, opts, .pin_root_path = map_path);
```

tohojo:
> Hi @tohojo,
> I've made the switch over to BTF-defined maps now (for the XDP-program at least), and it is kinda working. I am however having issues getting the map fd from userspace (which I need for the periodic cleanup of the map). After just opening the object, the maps do not seem to have gotten a valid file descriptor yet (-1), and after loading the object, the name of the ts_start map disappears for some reason. This is the output I get from running the program currently (as of commit [c4b75a9](https://github.com/xdp-project/bpf-examples/pull/7/commits/c4b75a91c956e65fee416853a1d17d84c307122b)):
> 
> ```
> Maps after opening object
> Map name: ts_start, fd: -1
> Map name: rtt_events, fd: -1
>  
> Maps after loading object
> Map name: , fd: 5
> Map name: rtt_events, fd: 6
> ```
> 
> Any advise on how to handle this? My current workaround is to simply take the fd of the first map from the object, but that does not feel like a very good long term solution.

libbpf will reload the name from the kernel when reusing the fd, so if there's no name that would be because your kernel doesn't have one? Which kernel version are you running? And what output does `bpftool map` show?

simosund:
> libbpf will reload the name from the kernel when reusing the fd, so if there's no name that would be because your kernel doesn't have one? Which kernel version are you running? And what output does `bpftool map` show?

I'm running kernel version 5.4.0-60 (under Ubuntu 20.04.1LTS), and the output of bpftool map is:
```
1: hash  flags 0x0
        key 16B  value 16B  max_entries 16384  memlock 1576960B
7: perf_event_array  name rtt_events  flags 0x0
        key 4B  value 4B  max_entries 8  memlock 4096B
```
Also tried it on a VM running Ubuntu 20.10 with kernel version 5.8.0-38, with the same result. 

I tc version `tc-utility, iproute2-ss200127` (and `tc-utility, iproute2-ss200602` on the VM) if that matters, as the map is pinned by tc.

tohojo:
Simon Sundberg <notifications@github.com> writes:

>  > libbpf will reload the name from the kernel when reusing the fd, so
>  > if there's no name that would be because your kernel doesn't have
>  > one? Which kernel version are you running? And what output does
>  > `bpftool map` show?
>
> I'm running kernel version 5.4.0-60 (under Ubuntu 20.04.1LTS), and the output of bpftool map is:
> ```
> 1: hash  flags 0x0
>         key 16B  value 16B  max_entries 16384  memlock 1576960B
> 7: perf_event_array  name rtt_events  flags 0x0
>         key 4B  value 4B  max_entries 8  memlock 4096B
> ```
> Also tried it on a VM running Ubuntu 20.10 with kernel version 5.8.0-38, with the same result. 
>
> I tc version `tc-utility, iproute2-ss200127` (and `tc-utility,
> iproute2-ss200602` on the VM) if that matters, as the map is pinned by
> tc.

Ah, 'tc' doesn't have BTF support so it can't read the name of a
BTF-defined map. With the changes you made it shouldn't be necessary to
pin it with TC, though. You can just call bpf_map__pin() and libbpf will
pin it in the same path as it would reuse it from. You can check with
bpf_map__is_pinned() if it's necessary to pin or if it already was.

Alternatively, you can build the latest iproute2-next from git which has
libbpf support - then you should get full BTF support in tc as well :)

simosund:
> Ah, 'tc' doesn't have BTF support so it can't read the name of a
BTF-defined map. With the changes you made it shouldn't be necessary to
pin it with TC, though. You can just call bpf_map__pin() and libbpf will
pin it in the same path as it would reuse it from. You can check with
bpf_map__is_pinned() if it's necessary to pin or if it already was.


No, my tc (or iproute2) version doesn't have libbpf support, so therefore the tc-bpf program doesn't use a BTF-defined map but rather `struct bpf_elf_map`. I guess I could do something similar as in [traffic-pacing-edt/edt_pacer_vlan.c](https://github.com/xdp-project/bpf-examples/blob/master/traffic-pacing-edt/edt_pacer_vlan.c#L61-L84) to use BTF-defined maps if they're supported. 

But you are completely right in that I don't need to pin the map with TC. I tried simply switching order so that the XDP-program which is loaded with libbpf is loaded first (and thus pins the map if it doesn't already exist), and then TC gets to reuse that map instead and the name is preserved. 

I'm not sure if I understand why I would need to "manually" pin the map with bpf_map__pin() though. Just leaving libbpf to handle that automatically based on the pinning attribute in the BTF-defined map seems to work fine? As far as I understand it, libbpf handles the automatic pinning similarly to iproute2, in that it will automatically reuse the map if it already exists, or otherwise pin it.

tohojo:
Simon Sundberg <notifications@github.com> writes:

>> Ah, 'tc' doesn't have BTF support so it can't read the name of a
> BTF-defined map. With the changes you made it shouldn't be necessary to
> pin it with TC, though. You can just call bpf_map__pin() and libbpf will
> pin it in the same path as it would reuse it from. You can check with
> bpf_map__is_pinned() if it's necessary to pin or if it already was.
>
>
> No, my tc (or iproute2) version doesn't have libbpf support, so
> therefore the tc-bpf program doesn't use a BTF-defined map but rather
> `struct bpf_elf_map`. I guess I could do something similar as in
> [traffic-pacing-edt/edt_pacer_vlan.c](https://github.com/xdp-project/bpf-examples/blob/master/traffic-pacing-edt/edt_pacer_vlan.c#L61-L84)
> to use BTF-defined maps if they're supported.
>
> But you are completely right in that I don't need to pin the map with
> TC. I tried simply switching order so that the XDP-program which is
> loaded with libbpf is loaded first (and thus pins the map if it
> doesn't already exist), and then TC gets to reuse that map instead and
> the name is preserved.

Yeah, just swapping the load order is probably the easiest :)

> I'm not sure if I understand why I would need to "manually" pin the
> map with bpf_map__pin() though. Just leaving libbpf to handle that
> automatically based on the pinning attribute in the BTF-defined map
> seems to work fine? As far as I understand it, libbpf handles the
> automatic pinning similarly to iproute2, in that it will automatically
> reuse the map if it already exists, or otherwise pin it.

No, you're right, of course, that was me misremembering how this works.
Just using the pinning attribute should be fine!

pping/pping.c
simosund: Wasn't aware of that, but seems like a very handy macro. Thanks! Fixed as of commit [5a40220](https://github.com/xdp-project/bpf-examples/pull/7/commits/5a40220d1aaeec64dd8e1d7279fc83d5a4e2bf9a).

simosund:
Hi again @tohojo. As of commit [5a40220](https://github.com/xdp-project/bpf-examples/pull/7/commits/5a40220d1aaeec64dd8e1d7279fc83d5a4e2bf9a) I think I've addressed all points from your initial high-level comments, with exception of using libxdp (think I'll wait with that until you've made it easy to include from this repository). So if you have the time, it would be great if you could take another look at this and see if there are any other issues. Otherwise I'll probably start working on adding IPv6 support.

pping/bpf_egress_loader.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+#
+# Author: Jesper Dangaaard Brouer <netoptimizer@brouer.com>
```
tohojo, line 3: You should probably add some indication here that you modified the script (which I'm assuming you did?); either adding yourself, or a comment that you modified it, or turning the 'author' into a series of copyright lines. @netoptimizer  any opinion?

pping/functions.sh
```diff
@@ -0,0 +1,64 @@
+#
+# Common functions used by scripts in this directory
+#  - Depending on bash 3 (or higher) syntax
+#
+# Author: Jesper Dangaaard Brouer <netoptimizer@brouer.com>
```
tohojo, line 5: As above

pping/parameters.sh
```diff
@@ -0,0 +1,90 @@
+#
+# Common parameter parsing used by scripts in this directory
+#  - Depending on bash 3 (or higher) syntax
+#
+# Author: Jesper Dangaaard Brouer <netoptimizer@brouer.com>
```
tohojo, line 5: And here

pping/pping.c
```diff
@@ -0,0 +1,357 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+#include <bpf/bpf.h>
+#include <bpf/libbpf.h>
+#include <linux/if_link.h>
+#include <net/if.h> // For if_nametoindex
+//#include <linux/err.h> // For IS_ERR_OR_NULL macro // use libbpf_get_error instead
```
tohojo, line 6: leftover?

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+#include <unistd.h>
+#include <stdbool.h>
+#include <signal.h> // For detecting Ctrl-C
+#include <sys/resource.h> // For setting rlmit
+#include <time.h>
+#include <pthread.h>
+#include "pping.h" //key and value structs for the ts_start map
+
+#define BILLION 1000000000UL
+#define MILLION 1000000UL
```
tohojo, line 22: These definitions don't really add any information (10**9 is a billion, of course, but defining it as such doesn't add anything where you're using it). If you're going to make a `#define`, make it meaningful in terms of the units in question. In this case that would be something like NS_PER_SECOND and US_PER_SECOND...

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+#define TCBPF_LOADER_SCRIPT "./bpf_egress_loader.sh"
+#define PINNED_DIR "/sys/fs/bpf/tc/globals"
+#define PPING_XDP_OBJ "pping_kern_xdp.o"
+#define XDP_PROG_SEC "xdp"
+#define PPING_TCBPF_OBJ "pping_kern_tc.o"
+#define TCBPF_PROG_SEC "pping_egress"
```
tohojo, line 28: Having defines like these is fine, but it would make more sense if you actually reused them between the BPF program and userspace; like above you're not really adding readability to the code where you use the define with just this definition at the top of the file, you're just adding another redirection when people have to figure out what's going on :)

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+{
+	struct bpf_program *prog;
+	int prog_fd;
+	int err;
+
+	if (sec)
+		prog = bpf_object__find_program_by_title(obj, sec);
+	else
+		prog = bpf_program__next(NULL, obj);
+	prog_fd = bpf_program__fd(prog);
```
tohojo, line 100: Using double-spacing to separate lines that "begin something new" is good practice for code readability (like paragraph breaks in text). You're mostly doing that, but there are a couple of exceptions, such as this one; I'd put a space before this line to visually separate the "fetch fd + handle error" operation.

There are a couple of other examples further down that I'm not going to comment on, just take this as a general point :)

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+		return err;
+	}
+	return 0;
+}
+
+static __u64 get_time_ns(clockid_t clockid)
+{
+	struct timespec t;
+	if (clock_gettime(clockid, &t) != 0) // CLOCK_BOOTTIME if using bpf_get_ktime_boot_ns
+		return 0;
```
tohojo, line 121: So is the calling code actually going to work if this returns 0? :)

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+
+static int remove_old_entries_from_map(int map_fd, __u64 max_age)
+{
+	int removed = 0, entries = 0;
+	struct ts_key key, prev_key = { 0 };
+	struct ts_timestamp value;
+	bool delete_prev = false;
+	__u64 now_nsec = get_time_ns(CLOCK_MONOTONIC);
+	if (now_nsec == 0)
+		return -errno;
```
tohojo, line 133: ...ah, it is, as long as it keeps working. I guess that's a reasonable assumption, carry on... :)

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+	err = bpf_set_link_xdp_fd(ifindex, prog_fd, xdp_flags);
+	if (err < 0) {
+		fprintf(stderr, "Failed loading xdp-program on interface %d\n",
+			ifindex);
+		return err;
+	}
+	return 0;
+}
+
+static __u64 get_time_ns(clockid_t clockid)
```
tohojo, line 117: Do the callers really need to select the clock? Bad things would happen if they disagree; and you're only ever calling this with the same argument, so maybe just remove it and put the constant into this function?

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+{
+	struct timespec t;
+	if (clock_gettime(clockid, &t) != 0) // CLOCK_BOOTTIME if using bpf_get_ktime_boot_ns
+		return 0;
+	return (__u64)t.tv_sec * BILLION + (__u64)t.tv_nsec;
+}
+
+static int remove_old_entries_from_map(int map_fd, __u64 max_age)
```
tohojo, line 125: `clean_map()`? shorter and sorta conveys the same :)

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+{
+	fprintf(stderr, "Lost %llu RTT events on CPU %d\n", lost_cnt, cpu);
+}
+
+int main(int argc, char *argv[])
+{
+	if (argc < 2) {
+		printf("Usage: ./pping_user <dev>\n");
+		return EXIT_FAILURE;
+	}
```
tohojo, line 196: Having code before variable declarations, while technically valid in modern C standards, is still a bit niche, so I'd prefer to just keep all the variable declarations at the top...

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+	
+	pthread_t tid;
+	struct map_cleanup_args clean_args;
+
+	struct perf_buffer *pb = NULL;
+	struct perf_buffer_opts pb_opts;
+
+	
+	// Increase rlimit
+	err = set_rlimit(RMEMLIM);
```
tohojo, line 217: Setting the rlimit to a static value is error-prone, as the limit applies to the total locked memory on the whole system. So either set it to infinity, or keep raising it in a loop until your program succeeds. The former is probably easier...

Thank goodness, the rlimit mechanism has been retired for BPF programs as of kernel 5.11 :)

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+	// Get map here to allow for unpinning at cleanup
+	map = bpf_object__find_map_by_name(obj, MAP_NAME);
+	err = libbpf_get_error(map);
+	if (err) {
+		fprintf(stderr, "Could not find map %s in %s: %s\n",
+			MAP_NAME, PPING_XDP_OBJ, strerror(err));
+		map = NULL;
+	}
+		
+
```
tohojo, line 250: Double space

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+		fprintf(stderr, "Failed attaching XDP program to %s: %s\n",
+			argv[1], strerror(-err));
+		goto cleanup;
+	}
+	xdp_attached = true;
+
+	//Load tc-bpf section on interface egress
+	snprintf(tc_cmd, MAX_COMMAND_LEN, "%s --dev %s --obj %s --sec %s",
+		 TCBPF_LOADER_SCRIPT, argv[1], PPING_TCBPF_OBJ, TCBPF_PROG_SEC);
+	err = system(tc_cmd);
```
tohojo, line 269: `system()` executes a shell, and since you're passing user-supplied data to it, this is basically carte blanche to execute anything the user feels like. Better hope this application is never ruin SUID root :)

A better alternative is to manually `fork()` + `exec()` with an explicit argv...

pping/pping_helpers.h
```diff
@@ -0,0 +1,66 @@
...
+					__u32 *tsval, __u32 *tsecr)
+{
+	if (tcph + 1 > data_end)
+		return -1;
+	int len = tcph->doff << 2;
+	if (len <= sizeof(struct tcphdr)) // No TCP options
+		return -1;
+	void *pos = (void *)(tcph + 1);
+	void *opt_end = ((void *)tcph + len);
+	__u8 i, opt, opt_size;
```
tohojo, line 33: See above about putting variable declarations at the top of the function. This example makes it especially difficult to read. You can assign pos and opt_end at the top of the function as long as you do the bounds checks before you dereference them...

pping/pping_helpers.h
```diff
@@ -0,0 +1,66 @@
...
+#pragma unroll
```
tohojo, line 34: With newer kernels it shouldn't be necessary to unroll the loop...

pping/pping_helpers.h
```diff
@@ -0,0 +1,66 @@
...
+	for (i = 0; i < MAX_TCP_OPTIONS; i++) {
+		if (pos + 1 > opt_end || pos + 1 > data_end)
+			return -1;
+		opt = *(__u8 *)pos; // Save value to avoid future data_end comparisons
```
tohojo, line 38: Don't think you need that comment; also, if you make pos __u8, you can save a cast :)

pping/pping_helpers.h
```diff
@@ -0,0 +1,66 @@
...
+		if (opt == 0) // Reached end of TCP options
+			return -1;
+		if (opt == 1) { // TCP NOP option - advance one byte
+			pos++;
+			continue;
+		}
+		// Option > 1, should have option size
+		if (pos + 2 > opt_end || pos + 2 > data_end)
+			return -1;
+		opt_size = *(__u8 *)(pos + 1); // Save value to avoid future data_end comparisons
```
tohojo, line 48: As above

pping/pping.c
```diff
@@ -0,0 +1,357 @@
...
+ *   if found looksup hashmap using reverse-flow+TSecr as key, 
+ *   and calculates RTT as different between now map value
+ * - Calculated RTTs are pushed to userspace 
+ *   (together with the related flow) and printed out 
+ */
+
+struct map_cleanup_args {
+	int map_fd;
+	__u64 max_age_ns;
+};
```
tohojo, line 56: Was wondering what this was for when I first read through this; maybe a comment? :)

tohojo (commented):
Overall in pretty good shape; these are mostly smallish nits and style issues :)

pping/pping.c
simosund: Yes, will clean that up.

pping/pping.c
simosund: I would argue that using these definitions do add something compared to using ex 1000000000 in that if you're somewhat number blind it's easier to read and leaves only one place where you can misstype the number of zeros. You are however completely right that NS_PER_SECOND and US_PER_SECOND would be more meaningful names and make it clearer what they're used for.

pping/pping.c
simosund: I honestly never even considered reusing them between files. The idea with defining them like that was mainly to only have to change them in one place (of that file at least) in case I decided to change names on something.

pping/pping.c
simosund: I realize that using 0 as an error-value is somewhat unconventional, but think my thought process was that I couldn't use negative values (unsigned) and 0 would be a very easy value to check for (and very unlikely to represent a real timer value as I doubt you would hit it even it this was the very first function you ran on boot-up). But is probably worth adding a documenting comment about the return value at least.

pping/pping.c
simosund: No, there's not really any good reason for the callers to select the clock in this case. Probably just a side effect of me having spent too long slicing data in Python, where parameterizing just about everything with default values makes it easier to experiment with different values.

pping/pping.c
simosund: 'clean_map()' does sound like a cleaner name :) I guess it could be considered slightly less informative in that it doesn't cover that the cleaning is based on time (how old the entries are), but that is more related to "how" rather than "what" the function does anyways.

pping/pping.c
simosund: Will admit that this was partly due to laziness on my part. Just wanted to get the program to work at that point and system seemed like the fewest lines of code to get the job done. In this early stage I figured that it was only going to be run on systems you already had root privilegies to anyways (and with the initial check that argv[1] is the name of an interface serving as a very poor security check), but is of course not production ready code.

pping/pping_helpers.h
simosund: Was thinking I would at least have to check if (tcph + 1 > data_end) before I set `len = tcph->doff << 2`, which is also used for opt_end, but apparently that's not required.

pping/pping_helpers.h
simosund: Thought it wouldn't hurt to support older kernels as well, although I haven't been keeping close track of the different BPF features I've used and their kernel requirements at this point, so there might as well be some other issues for older kernels anyways.

pping/pping_helpers.h
simosund: Thought using void for a pointer conceptually anything in the TCP-options part of the packet seemed more fitting, although it this case pos will only ever point at the option code field which is always 1 byte (and ex tsval and tsecr which are 4 bytes are handled with static offsets from pos) so guess I might as well have it point save that cast.

pping/pping.c
simosund: Could be arranged :)

pping/bpf_egress_loader.sh
simosund: I have added a comment here that I've made extended it by adding a "section" option (and have also changed a default fallback value, although this comment doesn't state that, but could be rectified).

pping/functions.sh
simosund: I have actually not modified this one at all, it's a straight copy of Jesper's (although before he added some ip functions).

pping/parameters.sh
simosund: This would be the same as for bpf_egress_loader.sh, where I've simply added the section (--sec) option.

pping/bpf_egress_loader.sh
netoptimizer: This is likely just a copy of my script.
My plan is to generalize this script and place it one level up.
It is ugly that we have several copies of the same script, but it is fine for this PR... I should be assigned to clean it up (maybe create a github issue to track this?)

pping/pping.c
```diff
@@ -14,28 +13,31 @@
...
-#define TCBPF_PROG_SEC "pping_egress"
+
 #define XDP_FLAGS XDP_FLAGS_UPDATE_IF_NOEXIST
-#define MAP_NAME "ts_start"
-#define MAP_CLEANUP_INTERVAL 1 * BILLION // Clean timestamp map once per second
-#define PERF_BUFFER_NAME "rtt_events"
+
+#define TS_MAP "ts_start"
+#define MAP_CLEANUP_INTERVAL 1 * NS_PER_SECOND // Clean timestamp map once per second
+#define TIMESTAMP_LIFETIME 10 * NS_PER_SECOND // Clear out entries from ts_start if they're over 10 seconds
```
tohojo, line 34: When you are defining constants as compound statements like this, it is good practice to enclose them in parenthesis (i.e., `(1 * NS_PER_SECOND)` instead of `1 * NS_PER_SECOND`. Otherwise, if you use the constants in an expression, they will just be replaced verbatim and be subject to operator precedence, which can lead to weird behaviour. You're not doing that with any of these right now, but you never know if that might change, and tracking down such a bug after the fact is a royal PITA, so better to get into the habit of always enclosing compound defines in `()` :)

pping/pping.c
```diff
@@ -114,21 +119,76 @@ static int xdp_attach(struct bpf_object *obj, const char *sec, int ifindex,
...
+{
+	int status;
+	int ret = -1;
+
+	pid_t pid = fork();
+
+	if (pid < 0)
+		return -errno;
+	if (pid == 0) {
+		execl(TCBPF_LOADER_SCRIPT, TCBPF_LOADER_SCRIPT,
```
tohojo, line 133: You could factor the actual fork() + exec() into a separate function and make these two wrappers around that; this becomes easier if you use execv() and just pass the argument list from the caller :)

pping/pping_helpers.h
tohojo: This is much nicer - but there's something odd going on with the whitespace used for indentation... Also, I'd suggest using size_t for len instead of int (to avoid signed-to-unsigned compare) :)

pping/bpf_egress_loader.sh
tohojo: Opened https://github.com/xdp-project/bpf-examples/issues/9 for this...

pping/pping.c
simosund: Didn't think of that, thanks for the heads up! Would indeed be very annoying to find if they had been used in division or something.

pping/pping.c
simosund: Sounds like a good idea, as 90% of the code in those functions is currently the same.

pping/pping_helpers.h
simosund: Hmm, yes not sure what is up with that. Seems like several actual tabs have gotten in there somehow. Must have messed something up in Emacs. Will let clang-format work over the files to see if it has affected any other files as well.

tohojo: This happens when `indent-tabs-mode` is not set correctly. I'm using dtrt-indent which tries (and usually succeeds) in guessing the right settings when opening a file: https://github.com/jscheid/dtrt-indent

pping/pping.c
```diff
@@ -0,0 +1,403 @@
...
+	char *const argv[] = { TCBPF_LOADER_SCRIPT, "--dev", interface,
+			       "--remove", NULL };
+	return run_program(TCBPF_LOADER_SCRIPT, argv);
+}
+
+/*
+ * Returns time of CLOCK_MONOTONIC as nanoseconds in a single __u64.
+ * On failure, the value 0 is returned (and errno will be set).
+ */
+static __u64 get_time_ns()
```
tohojo, line 162: Looks like you forgot to update the callers. It still compiles because a function declaration without any arguments really makes that function accept **any number** of arguments instead of **zero** arguments. So the correct declaration here would be `static __u64 get_time_ns(void)` - yay, esoteric language standards! :D

simosund: Oops, got a bit sloppy I see. Good catch. Will fix it later.

tohojo: Cool. Otherwise it looks like I'm running out of things to complain about - nice job :)

tohojo (changes requested):
Found a few runtime issues as well, now that I ran out of code style issues ;)

Also, a couple of points about commit messages: 

- The git formatter expects a completely blank line between the subject (first line) and the rest of the text. 
- Please break lines at 80 characters so it doesn't look weird when looking at 'git log' in a console.
- Try to keep the language imperative in the changes of the code ("change something" instead of "I changed something" or "this patch changes something")

So instead of:

```
Minor refactor and fixed weird spacing
Have refactored tc_bpf_load and tc_bpf_clear to use a common run_program function which does the fork+execv.
Have also fixed some weird spacing in pping_helpers.h, and ran clang-format with the kernel source tree .clang-format on the other source files which resultet in some minor changes.
    
Signed-off-by: Simon Sundberg <simon.sundberg@kau.se>
```

this could be something like:

```
Minor refactor and whitespace fixes
    
Refactor tc_bpf_load and tc_bpf_clear to use a common run_program
function which does the fork+execv.
    
Also fix some weird spacing in pping_helpers.h, and fix some formatting
issues, using clang-format with the kernel source tree .clang-format on the
whole tree.
    
Signed-off-by: Simon Sundberg <simon.sundberg@kau.se>
```

pping/pping.c
```diff
@@ -0,0 +1,403 @@
...
+	};
+
+	return !setrlimit(RLIMIT_MEMLOCK, &rlim) ? 0 : -errno;
+}
+
+static int bpf_obj_open(struct bpf_object **obj, const char *obj_path,
+			char *map_path)
+{
+	DECLARE_LIBBPF_OPTS(bpf_object_open_opts, opts,
+			    .pin_root_path = map_path);
```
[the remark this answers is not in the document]
simosund: Hmm, I guess I already had these folders from previously when tc was pinning the map. Long term libbpf should perhaps try to create the pinning folder recursively, or is that unreasonable? A temporary fix would indeed be to just hardcode it to create the sys/fs/bpf/tc folder, although feel like long term it would be nice if the program recursivly created the pinning directory no matter where it is.

pping/pping.c
```diff
@@ -0,0 +1,403 @@
...
+	struct perf_buffer_opts pb_opts;
+
+	// TODO - better argument parsing (more relevant as featureas are added)
+	if (argc < 2) {
+		printf("Usage: ./pping_user <dev>\n");
+		return EXIT_FAILURE;
+	}
+
+	// Increase rlimit
+	err = set_rlimit(RLIM_INFINITY);
```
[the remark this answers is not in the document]
simosund: Doesn't seem like too big of an ask. There are probably several other places where the error messages are less than optimal.

simosund:
> Found a few runtime issues as well, now that I ran out of code style issues ;)
> 
> Also, a couple of points about commit messages:
> 
> * The git formatter expects a completely blank line between the subject (first line) and the rest of the text.
> * Please break lines at 80 characters so it doesn't look weird when looking at 'git log' in a console.
> * Try to keep the language imperative in the changes of the code ("change something" instead of "I changed something" or "this patch changes something")
> 

So, would you like me to rebase in order to rewrite the commit messages in this style, or is messing with the history at this point an even worse sin?

tohojo:
> So, would you like me to rebase in order to rewrite the commit messages in this style, or is messing with the history at this point an even worse sin?

Yeah, please go ahead and rewrite history. As long as it is pre-merge this is fine; and in fact often encouraged. The final version of the commits is what will be preserved in the history for posterity, so the important thing is the readability of those.

For kernel submissions this also means folding fixes into the original patches. We're not as strict with that in this repository (as you can probably tell by looking at the history), so you don't have to do that for this PR. But fixing the commit messages is more important as that will go into the history, and right now `git log --oneline` looks... less than ideal... on this branch :)

If you do want to play with reshuffling commits and folding fixes into other patches (for this or for future series), I can recommend `stgit` for managing this: https://github.com/stacked-git/stgit

Oh, and if you're not already using `magit` in Emacs, I'd recommend taking a look at that as well; that also has a plugin for `stgit`...

simosund:
Hi @tohojo,
Have rewritten all the commit messages now, so they're hopefully ok (or at least better). Didn't think about it until now, but I see that for example the headers of all commit messages to ex. traffic-pacing-edt are prefixed with "traffic-pacing-edt:". Is that something I should have done (but with "pping:" then)?

Have also added the check for if the program is run as root or not (stealing your code snippet) and attempt to create the /sys/fs/bpf/tc folder before the map is pinned. 

So feel free to take another look at it and see if there's anything else that needs fixing (or if the fixes themselves need fixing...).

simosund:
Also, thanks for the tip about `magit`, seems pretty nice.

pping/pping.c
```diff
@@ -0,0 +1,428 @@
...
+	return !setrlimit(RLIMIT_MEMLOCK, &rlim) ? 0 : -errno;
+}
+
+static int mkdir_if_noexist(const char *path)
+{
+	struct stat st = {0};
+
+	if (stat(path, &st) != 0 || !S_ISDIR(st.st_mode)) {
+		return mkdir(path, 0700);
+	}
```
tohojo, line 87: `stat()`'ing the path first does seem like the obvious thing to do, but it's not strictly necessary with `mkdir()`. And the test you're doing doesn't really make sense (if `!S_ISDIR()` fails you still call mkdir() which is not going to work anyway).

So you could just make this whole function into:
```C
static int mkdir_if_noexist(const char *path)
{
	if (mkdir(path, 0700) && errno != EEXIST)
		return -errno;
	return 0
}
```

which will achieve basically the same thing and even save a syscall :)

tohojo:
> Have rewritten all the commit messages now, so they're hopefully ok (or at least better). Didn't think about it until now, but I see that for example the headers of all commit messages to ex. traffic-pacing-edt are prefixed with "traffic-pacing-edt:". Is that something I should have done (but with "pping:" then)?

Looks much better! But yeah, now that you mention it, prefixing the subjects with `pping:` would be good - makes it obvious which component a commit modifies when browsing the history of the whole repository. We haven't been completely consistent with this, but we really should :)

> So feel free to take another look at it and see if there's anything else that needs fixing (or if the fixes themselves need fixing...).

Other than the above and the comment on `mkdir()` I think we're done nitpicking this for now ;) Good job!

pping/pping.c
simosund: My though was that it would be possible to detect and differentiate between if the directory `path` exists or if ex. a file `path` exists. If the directory `path` exists, we don't need to create one and the function returns sucessfully. If a file `path` already exists, it will try to create a folder, but that should fail (and therefore the function should return unsucessfully).

If I'm not misstaken, your function (while a lot prettier) would in the case that `path` already exists, no matter if it's a directory or not, return successfully. So if the file `path` already exists, the function would return succesfully, and whatever the caller does assuming that the directory `path` exists would most likely fail (like creating a sub-directory).

simosund:
> Looks much better! But yeah, now that you mention it, prefixing the subjects with `pping:` would be good - makes it obvious which component a commit modifies when browsing the history of the whole repository. We haven't been completely consistent with this, but we really should :)

Can rewrite the commits prepending `pping:` to the header. Also learned that commit messages should normally keep lines below 72 characters (instead of 80), so that git can indent it a bit and still keep it below 80, so will probably redo the word-wrapping based on 72 instead of 80 characters.

pping/pping.c
tohojo: Hmm, yeah, you're right that my version relies on later failures if there's already a non-directory there. So if your intention is to fail if a regular file exists, how about making this explicit:

```C
ret = stat(path, &st);
if (ret) {
	if (errno != ENOENT)
		return -errno;

	return mkdir(path, 0700) ? -errno : 0;
}
return S_ISDIR(st.st_mode) ? 0 : -EEXIST;
```

This has the added benefit of unpacking the errno into an explicit return value (which means you'll need to handle in the caller, of course). Relying on errno propagating up the callstack is generally brittle as it can get overriden; for instance, adding a debug log somewhere can reset errno, and suddenly you have a heisenbug; much better to read out errno right after you call into libc and put the error value into the function return value of your own library functions :)

simosund: I agree that the intention in my version was quite unclear, and not really sure why I didn't return the error code (have done so in most other places), so will go with your new suggestion.

simosund:
So @tohojo, now I've finally gotten around to refactoring `mkdir_if_noexist()` (squshed it with the previous commit) and fixing up all the commit messages (prepending `pping` and wrapping at 72 instead of 80 characters). So hopefully it should be ready to go now, and I can move on to adding some new features instead.

tohojo (approved):
Yup, looks good - nice work! :)

PR #6, opened 2021-01-08
Cleanups for bpf-examples

netoptimizer:
Some cleanups that @tohojo should review, please.

PR #8, opened 2021-01-13
Cleanups for bpf-examples V2

netoptimizer:
This PR replace #6

### DIFF
--- a/MTU-tests/Makefile
+++ b/MTU-tests/Makefile
@@ -8,7 +8,8 @@ LIB_DIR = ../lib
 
 include $(LIB_DIR)/common.mk
 
-# The iproute2 'tc' tool doesn't understand BTF debug info
+ifndef HAVE_TC_LIBBPF
+# If the iproute2 'tc' tool doesn't understand BTF debug info
 # use llvm-strip to remove this debug info from object file
 #
 # *BUT* cannot strip everything as it removes ELF elems needed for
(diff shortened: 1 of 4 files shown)


netoptimizer:
Dropped the patch that @tohojo didn't like in #6

tohojo:
Hmm, okay, but if we're installing the libbpf headers to a different directory, shouldn't it be under headers/ instead of under lib/ ? :)

netoptimizer:
> Hmm, okay, but if we're installing the libbpf headers to a different directory, shouldn't it be under headers/ instead of under lib/ ? :)

No, I don't want it installed into headers/.
Because, I want it is be *very* clear that these files belongs to libbpf, when the system have libbpf installed, then they will not be located there.

I want to avoid other projects mispintreret and copy everything from headers/ directory.

You comment makes me realize that I've not been clear enough about the original intent with the 'headers/' directory.
All other/most projects have a direcory called 'include/' for their projects include/header files.
The intent with naming it 'headers/' was to keep the kernel headers in a separate directory, and then have an 'include/' directory with the projects own header files (in xdp-tutorial this was called 'common/', which should have been named include instead).

The intent was to have 'headers/' strictly contain the files need to compile BPF-progs, which the relevant headers copied from the kernel and the modified/compat headers like `compiler.h`, `err.h` and `perf-sys.h`. This would make it easier for other projects to understand or don't understand simply copy 'headers/' into their repos.

netoptimizer:
Merge based on agreement with Toke on video chat

PR #10, opened 2021-02-08
pping: Add IPv6 support

simosund:
Hi @tohojo,
Been working a little bit on adding IPv6 support to pping and now got it to a working stage. So thought I would start a pull request so you can provide some feedback on it.

I use `struct in6_addr` to store both IPv6 and IPv4 addresses, by mapping IPv4 addresses to the IPv6 range as we've talked about previously. I'm not entierly sure I do the mapping correctly, however as I'm also the one who does the "unmapping" it seems to work. Still, it would be great if you could take a look at that. 

I've also renamed some things to move away a bit from just TCP timestamps and be more general, so that pping can be extended to support different protocols in the future (as we've also talked about). There's still much more do be done in this area though, and my focus has mainly been on adding IPv6 support. I've therefore kept the code in `pping_kern*` quite similar to how it was for now, even if it means a fair bit of repetition.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -248,7 +248,7 @@ static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
 			     size_t size)
 {
 	if (af == AF_INET)
-		return inet_ntop(af, &(addr->s6_addr[12]),
+		return inet_ntop(af, &addr->s6_addr[12],
 				 buf, size) ? -errno : 0;
 	else if (af == AF_INET6)
 		return inet_ntop(af, addr, buf, size) ? -errno : 0;
@@ -261,8 +261,8 @@ static void handle_rtt_event(void *ctx, int cpu, void *data, __u32 data_size)
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
 
-	format_ip_address(e->flow.ipv, &(e->flow.saddr), saddr, sizeof(saddr));
-	format_ip_address(e->flow.ipv, &(e->flow.daddr), daddr, sizeof(daddr));
+	format_ip_address(e->flow.ipv, &e->flow.saddr, saddr, sizeof(saddr));
+	format_ip_address(e->flow.ipv, &e->flow.daddr, daddr, sizeof(daddr));
 
 	printf("%llu.%06llu ms %s:%d+%s:%d\n", e->rtt / NS_PER_MS,
 	       e->rtt % NS_PER_MS, saddr, ntohs(e->flow.sport), daddr,
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -187,8 +187,8 @@ static __u64 get_time_ns(void)
 static int clean_map(int map_fd, __u64 max_age)
 {
 	int removed = 0;
-	struct ts_key key, prev_key = { 0 };
-	struct ts_timestamp value;
+	struct packet_id key, prev_key = { 0 };
+	struct packet_timestamp value;
 	bool delete_prev = false;
 	__u64 now_nsec = get_time_ns();
 
@@ -240,17 +240,33 @@ static void *periodic_map_cleanup(void *args)
 	pthread_exit(NULL);
 }
 
+/*
+ * Wrapper around inet_ntop designed to handle the "bug" that mapped IPv4
+ * addresses are formated as IPv6 addresses for AF_INET6
+ */
+static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
+			     size_t size)
+{
+	if (af == AF_INET)
+		return inet_ntop(af, &(addr->s6_addr[12]),
+				 buf, size) ? -errno : 0;
+	else if (af == AF_INET6)
+		return inet_ntop(af, addr, buf, size) ? -errno : 0;
+	return -EINVAL;
+}
+
 static void handle_rtt_event(void *ctx, int cpu, void *data, __u32 data_size)
 {
 	const struct rtt_event *e = data;
-	struct in_addr saddr, daddr;
-	saddr.s_addr = e->flow.saddr;
-	daddr.s_addr = e->flow.daddr;
+	char saddr[INET6_ADDRSTRLEN];
+	char daddr[INET6_ADDRSTRLEN];
 
-	// inet_ntoa is deprecated, will switch to inet_ntop when adding IPv6 support
-	printf("%llu.%06llu ms %s:%d+", e->rtt / NS_PER_MS, e->rtt % NS_PER_MS,
-	       inet_ntoa(daddr), ntohs(e->flow.dport));
-	printf("%s:%d\n", inet_ntoa(saddr), ntohs(e->flow.sport));
+	format_ip_address(e->flow.ipv, &(e->flow.saddr), saddr, sizeof(saddr));
+	format_ip_address(e->flow.ipv, &(e->flow.daddr), daddr, sizeof(daddr));
+
+	printf("%llu.%06llu ms %s:%d+%s:%d\n", e->rtt / NS_PER_MS,
+	       e->rtt % NS_PER_MS, saddr, ntohs(e->flow.sport), daddr,
+	       ntohs(e->flow.dport));
 }
 
 static void handle_missed_rtt_event(void *ctx, int cpu, __u64 lost_cnt)
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -81,7 +81,7 @@ static int set_rlimit(long int lim)
 static int mkdir_if_noexist(const char *path)
 {
 	int ret;
-	struct stat st = {0};
+	struct stat st = { 0 };
 
 	ret = stat(path, &st);
 	if (ret) {
@@ -261,12 +261,12 @@ static void handle_rtt_event(void *ctx, int cpu, void *data, __u32 data_size)
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
 
-	format_ip_address(e->flow.ipv, &e->flow.saddr, saddr, sizeof(saddr));
-	format_ip_address(e->flow.ipv, &e->flow.daddr, daddr, sizeof(daddr));
+	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
+	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
 
 	printf("%llu.%06llu ms %s:%d+%s:%d\n", e->rtt / NS_PER_MS,
-	       e->rtt % NS_PER_MS, saddr, ntohs(e->flow.sport), daddr,
-	       ntohs(e->flow.dport));
+	       e->rtt % NS_PER_MS, saddr, ntohs(e->flow.saddr.port), daddr,
+	       ntohs(e->flow.daddr.port));
 }
 
 static void handle_missed_rtt_event(void *ctx, int cpu, __u64 lost_cnt)
--- a/pping/pping.h
+++ b/pping/pping.h
@@ -9,21 +9,29 @@
 #define TCBPF_PROG_SEC "pping_egress"
 
 /*
- * Struct to hold a full network tuple
+ * Struct that can hold the source or destination address for a flow (l3+l4).
  * Works for both IPv4 and IPv6, as IPv4 addresses can be mapped to IPv6 ones
- * based on RFC 4291 Section 2.5.5.2. The ipv member is technically not 
- * necessary, but makes it easier to determine if it is an IPv4 or IPv6 address
- * (don't need to look at the first 12 bytes of address).
- * The proto memeber is not currently used, but could be useful once pping
- * is extended to work for other protocols than TCP
+ * based on RFC 4291 Section 2.5.5.2.
+ */
+struct flow_address {
+	struct in6_addr ip;
+	__u16 port;
+	__u16 reserved;
+};
+
+/*
+ * Struct to hold a full network tuple
+ * The ipv member is technically not necessary, but makes it easier to 
+ * determine if saddr/daddr are IPv4 or IPv6 address (don't need to look at the
+ * first 12 bytes of address). The proto memeber is not currently used, but 
+ * could be useful once pping is extended to work for other protocols than TCP.
  */
 struct network_tuple {
-	struct in6_addr saddr;
-	struct in6_addr daddr;
-	__u16 sport;
-	__u16 dport;
+	struct flow_address saddr;
+	struct flow_address daddr;
 	__u16 proto; //IPPROTO_TCP, IPPROTO_ICMP, QUIC etc
-	__u16 ipv; //AF_INET or AF_INET6
+	__u8 ipv; //AF_INET or AF_INET6
+	__u8 reserved;
 };
 
 struct packet_id {
@@ -34,11 +42,13 @@ struct packet_id {
 struct packet_timestamp {
 	__u64 timestamp;
 	__u8 used;
+	__u8 reserved[7];
 };
 
 struct rtt_event {
 	__u64 rtt;
 	struct network_tuple flow;
+	__u32 reserved;
 };
 
 #endif
--- a/pping/pping.h
+++ b/pping/pping.h
@@ -3,30 +3,41 @@
 #define PPING_H
 
 #include <linux/types.h>
+#include <linux/in6.h>
 
 #define XDP_PROG_SEC "xdp"
 #define TCBPF_PROG_SEC "pping_egress"
 
-// TODO - change to support both IPv4 and IPv6 (IPv4 addresses can be mapped to IPv6 addresses)
-struct ipv4_flow {
-	__u32 saddr;
-	__u32 daddr;
+/*
+ * Struct to hold a full network tuple
+ * Works for both IPv4 and IPv6, as IPv4 addresses can be mapped to IPv6 ones
+ * based on RFC 4291 Section 2.5.5.2. The ipv member is technically not 
+ * necessary, but makes it easier to determine if it is an IPv4 or IPv6 address
+ * (don't need to look at the first 12 bytes of address).
+ * The proto memeber is not currently used, but could be useful once pping
+ * is extended to work for other protocols than TCP
+ */
+struct network_tuple {
+	__u8 ipv; //AF_INET or AF_INET6
+	struct in6_addr saddr;
+	struct in6_addr daddr;
 	__u16 sport;
 	__u16 dport;
+	__u16 proto; //IPPROTO_TCP, IPPROTO_ICMP, QUIC etc
 };
 
-struct ts_key {
-	struct ipv4_flow flow;
-	__u32 tsval;
+struct packet_id {
+	struct network_tuple flow;
+	__u32 identifier; //tsval for TCP packets
 };
 
-struct ts_timestamp {
+struct packet_timestamp {
 	__u64 timestamp;
 	__u8 used;
 };
 
 struct rtt_event {
-	struct ipv4_flow flow;
+	struct network_tuple flow;
 	__u64 rtt;
 };
 
--- a/pping/pping.h
+++ b/pping/pping.h
@@ -18,12 +18,12 @@
  * is extended to work for other protocols than TCP
  */
 struct network_tuple {
-	__u8 ipv; //AF_INET or AF_INET6
 	struct in6_addr saddr;
 	struct in6_addr daddr;
 	__u16 sport;
 	__u16 dport;
 	__u16 proto; //IPPROTO_TCP, IPPROTO_ICMP, QUIC etc
+	__u16 ipv; //AF_INET or AF_INET6
 };
 
 struct packet_id {
@@ -37,8 +37,8 @@ struct packet_timestamp {
 };
 
 struct rtt_event {
-	struct network_tuple flow;
 	__u64 rtt;
+	struct network_tuple flow;
 };
 
 #endif
--- a/pping/pping_helpers.h
+++ b/pping/pping_helpers.h
@@ -29,7 +29,7 @@ struct parsing_context {
 	void *data;           //Start of eth hdr
 	void *data_end;       //End of safe acessible area
 	struct hdr_cursor nh; //Position to parse next
-	__u32 len;            //Full packet length (headers+data)
+	__u32 pkt_len;        //Full packet length (headers+data)
 };
 
 /*
@@ -109,7 +109,7 @@ static int parse_tcp_identifier(struct parsing_context *ctx, bool is_egress,
 		return -1;
 
 	// Do not timestamp pure ACKs
-	if (is_egress && ctx->nh.pos - ctx->data >= ctx->len && !tcph->syn)
+	if (is_egress && ctx->nh.pos - ctx->data >= ctx->pkt_len && !tcph->syn)
 		return -1;
 
 	if (parse_tcp_ts(tcph, ctx->data_end, &tsval, &tsecr) < 0)
--- a/pping/pping_helpers.h
+++ b/pping/pping_helpers.h
@@ -7,6 +7,8 @@
 #include <string.h>
 #include "pping.h"
 
+#define AF_INET 2
+#define AF_INET6 10
 #define MAX_TCP_OPTIONS 10
 
 /*
@@ -15,14 +17,10 @@
 static __always_inline void map_ipv4_to_ipv6(__be32 ipv4, struct in6_addr *ipv6)
 {
 	/* __u16 ipv4_prefix[6] = {0x0, 0x0, 0x0, 0x0, 0x0, 0xffff}; */
-	/* memcpy(&(ipv6->in6_u.u6_addr8), ipv4_prefix, sizeof(ipv4_prefix)); */
-	memset(&(ipv6->in6_u.u6_addr8[0]), 0x00, 10);
-	memset(&(ipv6->in6_u.u6_addr8[10]), 0xff, 2);
-#if __UAPI_DEF_IN6_ADDR_ALT
+	/* memcpy(ipv6, ipv4_prefix, sizeof(ipv4_prefix)); // Won't load on TC */
+	memset(&ipv6->in6_u.u6_addr8[0], 0x00, 10);
+	memset(&ipv6->in6_u.u6_addr8[10], 0xff, 2);
 	ipv6->in6_u.u6_addr32[3] = ipv4;
-#else
-	memcpy(&(ipv6->in6_u.u6_addr8[12]), &ipv4, sizeof(ipv4));
-#endif
 }
 
 /*
--- a/pping/pping_helpers.h
+++ b/pping/pping_helpers.h
@@ -19,6 +19,21 @@
 #define MAX_TCP_OPTIONS 10
 
 /*
+ * This struct keeps track of the data and data_end pointers from the xdp_md or
+ * __skb_buff contexts, as well as a currently parsed to position kept in nh.
+ * The member data_end_end pointer which has the adress that would correspond
+ * to the logical end of the packet (for XDP, this is the same as data_end,
+ * but for TC this is skb->data + skb->len), which is useful for determining
+ * how much data a certain header encloses.
+ */
+struct parsing_context {
+	void *data;           //Start of eth hdr
+	void *data_end;       //End of safe acessible area
+	void *data_end_end;   //Logical end of packet
+	struct hdr_cursor nh; //Position to parse next
+};
+
+/*
  * Maps and IPv4 address into an IPv6 address according to RFC 4291 sec 2.5.5.2
  */
 static void map_ipv4_to_ipv6(__be32 ipv4, struct in6_addr *ipv6)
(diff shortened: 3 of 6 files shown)


pping/pping.h
```diff
@@ -3,30 +3,41 @@
...
+ * Struct to hold a full network tuple
+ * Works for both IPv4 and IPv6, as IPv4 addresses can be mapped to IPv6 ones
+ * based on RFC 4291 Section 2.5.5.2. The ipv member is technically not 
+ * necessary, but makes it easier to determine if it is an IPv4 or IPv6 address
+ * (don't need to look at the first 12 bytes of address).
+ * The proto memeber is not currently used, but could be useful once pping
+ * is extended to work for other protocols than TCP
+ */
+struct network_tuple {
+	__u8 ipv; //AF_INET or AF_INET6
```
tohojo, line 21: This will leave a 3-byte hole in your struct, because the compiler aligns fields to their size, adding holes in-between (because some architectures can more efficiently access memory if it's aligned).

This is inefficient, but more importantly it is not guaranteed that the compiler will zero out the holes when initialising the struct. And since you're using it as a key for your hashmap, those possibly uninitialised bytes will be part of the lookup leading to odd and hard-to-debug failures. So better pack it.

The `pahole` utility is really nice for seeing this (after compiling):
```C
$ pahole -C network_tuple pping
struct network_tuple {
	__u8                       ipv;                  /*     0     1 */

	/* XXX 3 bytes hole, try to pack */

	struct in6_addr            saddr;                /*     4    16 */
	struct in6_addr            daddr;                /*    20    16 */
	__u16                      sport;                /*    36     2 */
	__u16                      dport;                /*    38     2 */
	__u16                      proto;                /*    40     2 */

	/* size: 44, cachelines: 1, members: 6 */
	/* sum members: 39, holes: 1, sum holes: 3 */
	/* padding: 2 */
	/* last cacheline: 44 bytes */
};
$ pahole -C packet_id pping    
struct packet_id {
	struct network_tuple       flow;                 /*     0    44 */

	/* XXX last struct has 2 bytes of padding */

	__u32                      identifier;           /*    44     4 */

	/* size: 48, cachelines: 1, members: 2 */
	/* paddings: 1, sum paddings: 2 */
	/* last cacheline: 48 bytes */
};
```

You can use `__attribute__((packed))` to tell the compiler to pack the struct tightly, but in this case you might as well just move your 'ipv' member to the end and make it a u16. That gives you a nice divisible-by-four size of your struct, so when you put it into the packet_id struct afterwards, that too will be without holes.

Contrast after:

```C
$  pahole -C network_tuple pping
struct network_tuple {
	struct in6_addr            saddr;                /*     0    16 */
	struct in6_addr            daddr;                /*    16    16 */
	__u16                      sport;                /*    32     2 */
	__u16                      dport;                /*    34     2 */
	__u16                      proto;                /*    36     2 */
	__u16                      ipv;                  /*    38     2 */

	/* size: 40, cachelines: 1, members: 6 */
	/* last cacheline: 40 bytes */
};
$ pahole -C packet_id pping    
struct packet_id {
	struct network_tuple       flow;                 /*     0    40 */
	__u32                      identifier;           /*    40     4 */

	/* size: 44, cachelines: 1, members: 2 */
	/* last cacheline: 44 bytes */
};
```

pping/pping.h
```diff
@@ -3,30 +3,41 @@
...
-	__u32 tsval;
+struct packet_id {
+	struct network_tuple flow;
+	__u32 identifier; //tsval for TCP packets
 };
 
-struct ts_timestamp {
+struct packet_timestamp {
 	__u64 timestamp;
 	__u8 used;
```
tohojo, line 36: This has a 7-byte hole at the end. Not hugely important as you're not using it as a hash key, and it's smaller than a cacheline as well, so probably not much performance to be gained by packing it either. But something to be aware of if you extend it in the future.

pping/pping.h
```diff
@@ -3,30 +3,41 @@
...
 };
 
 struct rtt_event {
-	struct ipv4_flow flow;
+	struct network_tuple flow;
 	__u64 rtt;
```
tohojo, line 41: This is without holes because network_tuple happens to be 40 bytes long (after rearranging it per above). Otherwise there would be a hole. Moving the 'rtt' member up to the beginning of the struct would be a way to ensure you're not going to have holes if network_tuple does grow in the future.

pping/pping.h
tohojo: An alternative to making ipv be u16, is to add another u8 member beneath it ("reserved", say). This will allow you to re-use the space later without breaking the ABI. Not hugely important in this application, but just FYI (you sometimes see this in kernel structs that are exposed to userspace) :)

pping/pping_helpers.h
```diff
@@ -2,19 +2,27 @@
...
-	flow->saddr = saddr;
-	flow->daddr = daddr;
-	flow->sport = sport;
-	flow->dport = dport;
-	return 0;
+	/* __u16 ipv4_prefix[6] = {0x0, 0x0, 0x0, 0x0, 0x0, 0xffff}; */
+	/* memcpy(&(ipv6->in6_u.u6_addr8), ipv4_prefix, sizeof(ipv4_prefix)); */
+	memset(&(ipv6->in6_u.u6_addr8[0]), 0x00, 10);
+	memset(&(ipv6->in6_u.u6_addr8[10]), 0xff, 2);
+#if __UAPI_DEF_IN6_ADDR_ALT
```
tohojo, line 21: Double-underscore at the start of a name is usually a hint that something is considered 'private' (but C doesn't have a way to enforce that). So checking for such defines outside of the header file(s) that define them is not a good idea. For this particular define, it seems like you'll need to compile in strict ANSI mode to cause it to be unset, which I think it would be fine to just declare unsupported.

So in other words, just drop the `#if` check and the memcpy, and use the addr32 members unconditionally :)

pping/pping_helpers.h
```diff
@@ -2,19 +2,27 @@
 #ifndef PPING_HELPERS_H
 #define PPING_HELPERS_H
 
-#include "pping.h"
+#include <linux/in6.h>
 #include <linux/tcp.h>
+#include <string.h>
+#include "pping.h"
 
 #define MAX_TCP_OPTIONS 10
```
tohojo, line 10: This might be a good place to add:

```C
#define AF_INET 2
#define AF_INET6 10
```

(pping_kern_xdp.c doesn't compile without those defines, and pulling in the AF_* constants through BPF headers is a PITA, so you might as well just define them yourself)

pping/pping_helpers.h
```diff
@@ -2,19 +2,27 @@
...
 {
-	flow->saddr = saddr;
-	flow->daddr = daddr;
-	flow->sport = sport;
-	flow->dport = dport;
-	return 0;
+	/* __u16 ipv4_prefix[6] = {0x0, 0x0, 0x0, 0x0, 0x0, 0xffff}; */
+	/* memcpy(&(ipv6->in6_u.u6_addr8), ipv4_prefix, sizeof(ipv4_prefix)); */
+	memset(&(ipv6->in6_u.u6_addr8[0]), 0x00, 10);
+	memset(&(ipv6->in6_u.u6_addr8[10]), 0xff, 2);
```
tohojo, line 20: I personally think the commented-out version of these is the easiest to read. You can make it even easier by just using the `ipv6` pointer directly in the `memcpy()` instead of dereferencing the pointer, then taking the address of the first member. I.e.,:
```C
__u16 ipv4_prefix[6] = {0x0, 0x0, 0x0, 0x0, 0x0, 0xffff};
memcpy(ipv6, ipv4_prefix, sizeof(ipv4_prefix));
```

pping/pping_kern_tc.c
```diff
@@ -43,34 +45,51 @@ int tc_bpf_prog_egress(struct __sk_buff *skb)
...
 	struct iphdr *iph;
+	struct ipv6hdr *ip6h;
 	struct tcphdr *tcph;
 
-	proto = parse_ethhdr(&nh, data_end, &eth);
-	if (bpf_ntohs(proto) != ETH_P_IP)
+	struct packet_id p_id = { 0 };
+	struct packet_timestamp p_ts = { 0 };
+
+	proto = bpf_ntohs(parse_ethhdr(&nh, data_end, &eth));
```
tohojo, line 59: Micro-optimisation trick: When you're comparing to constants like this, you can apply `bpf_htons()` to the constant instead of applying `bpf_ntohs()` to the runtime value. That way, the compiler can just substitute in the converted constant, and you don't have to do any conversion at runtime. So drop the `bpf_ntohs()` here...

pping/pping_kern_tc.c
```diff
@@ -43,34 +45,51 @@ int tc_bpf_prog_egress(struct __sk_buff *skb)
...
+
+	// Parse IPv4/6 header
+	if (proto == ETH_P_IP) {
```
tohojo, line 62: ...and replace this with `if (proto == bpf_htons(ETH_P_IP)) {`

pping/pping_kern_xdp.c
```diff
@@ -37,46 +39,68 @@ int xdp_prog_ingress(struct xdp_md *ctx)
...
+	struct ipv6hdr *ip6h;
 	struct tcphdr *tcph;
 
-	proto = parse_ethhdr(&nh, data_end, &eth);
-	if (bpf_ntohs(proto) != ETH_P_IP)
+	struct packet_id p_id = { 0 };
+	struct packet_timestamp *p_ts;
+	struct rtt_event event = { 0 };
+
+	proto = bpf_ntohs(parse_ethhdr(&nh, data_end, &eth));
```
tohojo, line 54: As above...

Also, you're duplicating an awful lot of code here - maybe refactoring the code so you can reuse some of it between XDP and TC modes would be useful? :)

tohojo (changes requested):
Looks mostly good. The main thing that is likely to break is the struct packing issue, the rest is mostly nits, style and optimisations :)

pping/pping.c
```diff
@@ -240,17 +240,33 @@ static void *periodic_map_cleanup(void *args)
...
 
+/*
+ * Wrapper around inet_ntop designed to handle the "bug" that mapped IPv4
+ * addresses are formated as IPv6 addresses for AF_INET6
+ */
+static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
+			     size_t size)
+{
+	if (af == AF_INET)
+		return inet_ntop(af, &(addr->s6_addr[12]),
```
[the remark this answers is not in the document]
simosund: I've never actually checked up the operator precedence for `->`, simply assumed `&` would apply to the closest thing (because that's how my brain likes to parse it). But `->` having higher precedence of course makes sense when you think about it. Will go through and remove any unecessary parentheses of this type that I spot.

pping/pping.h
simosund: I actually ran into this problem first with the `packet_id` (or `ts_key` as it was named back then) a while back, and was first very confused what the verifier was unhappy with. While I did find some references regarding how you had to avoid holes in the struct (either by telling the compiler to pack it, or by manually padding the struct), someone claimed that initilizing the struct as `{0}` would initilize the whole struct (including holes) as 0. It seemed to work to me, so I simply ran with the latter back then as I figured I was less likely to screw that up =).

Will go with your suggestion here and try to keep it in mind going forward.

pping/pping.h
simosund: As per my previous reply, I will try to keep this in mind from now on. Would it be a bad idea to manually pad it with 7 bytes as the end (called `reserved`, or `padding` or something) to help remind my future self of this?

pping/pping_helpers.h
simosund: Not sure why I didn't think about the double underscore thing (also happens to be a very common practice in python which similary lacks means to make things private, although there it's also a common practice to use some of them anyways...). Just looked up the struct in the source and thought I would avoid potential issues, without really looking up what `__UAPA_DEF_IN6_ADDR_ALT` was for.

pping/pping_helpers.h
simosund: Yhea, was a bit unsure myself which was was better here, which is why I left the commented out version so you could see it. Guess the commented out version also saves a function call.

pping/pping_kern_tc.c
simosund: Hmm, wasn't aware compilers were smart enough to apply functions to constants at compile time. So actually thought I saved a call by applying it on proto instead of on the two different constants it will compare against. Those compiler folks are some clever people.

pping/pping_kern_xdp.c
simosund: Yes, I'm painfully aware of that (and mentioned it in both my original pull request as well as the commit message). I fully plan to refactor those parts a bit, but it was getting late so left that for another commit. 

Figured I would try and kill two birds with one stone when I refactor it, both avoiding (or at least significantly reducing) the elements of duplicated code, and also to make it easier to extend with different protocols. My thoughts there is to basically just have a function with a common interface for each protocol which returns an identifier (TSval or TSecr (depending on direction) for TCP).

pping/pping.h
tohojo: Ah, right, the verifier will also check and reject uninitialised padding; totally forgot about this. Still, I'm not sure how portable the `{0}` trick is (i.e., you may have problems on the userspace side), so better to just pack the structure (and slightly more efficient, too) :)

pping/pping.h
tohojo: As long as you use fields of the right size (i.e., in this case that would be __u8[7]), you can add the reserved member - it's already implicitly there. Or you can add a comment noting it; I have no strong preference either way :)

pping/pping_helpers.h
tohojo: Not necessarily - `memcmp()` and `memcpy()` are generally recognised by the compiler and may be replaced with other operations depending on the size and alignment of memory. It can be hard to predict, so you generally have to look at the generated code to know for sure. But it depends a little on compiler versions, which is why you may have to use `__builin_memcpy()` in BPF code sometimes...

pping/pping_kern_tc.c
tohojo: It's not always smart enough. For real functions the compiler will not generally optimise out the entire function (unless they are inlined; so again, not always predictable).

However, the BPF byte-order conversion functions are really macros; and they specifically handle constants for just this reason. See the definitions here:

https://github.com/libbpf/libbpf/blob/master/src/bpf_endian.h

pping/pping_kern_xdp.c
tohojo: Sure, sounds reasonable :)

pping/pping_helpers.h
simosund: Hmm, can't get the `memcpy()` version to work though. When I use that instead of the memset version I get:
```
$ sudo ./pping test123
ELF contains non-{map,call} related relo data in entry 0 pointing to section 5! Compiler bug?!
Error fetching program/map!
Unable to load program
ERROR: Exec error(1) occurred cmd: "/sbin/tc filter add dev test123 pref 2 handle 2 egress bpf da obj pping_kern_tc.o sec pping_egress"
Could not load section pping_egress of pping_kern_tc.o on interface test123: Unknown error -3
```

Any idea what this is about? The interesting part is that the XDP program is loaded first, and doesn't seem to run into any errors. Both the TC program and the XDP program use map_ipv4_to_ipv6 in essentially the same way (because of duplicated code...).

tohojo: Yeah, that's just what I was referring to above: It tries to make `memcpy()` into a real function call, which doesn't work in BPF. Try `__builtin_memcpy()` instead :)

pping/pping_kern_xdp.c
simosund: Been thinking a little about how to refactor these in a nice way. The parsing of the packets is essentially the same for both the TC and XDP program. The main differences are in which identifier is used (TSval on TC, TSecr on XDP) and that XDP reverses the order of source and destination (too match the one on TC). 

So will need some "reverse_flow" flag for the functions that parse the packet identifier (TSval or TSecr for TCP), but a bit undecided on how to best handle the reverse ordering of the network tuple. I currently see 3 ways to handle this.

1. Fill in the network_tuple in the "right" order (matching saddr in the packets with saddr in the tuple etc.) for both TC and XDP, and then have the XDP program swap those before the key lookup. This seems somewhat inefficent to me though.
2. Use the "reverse_flow" flag in if-switches that basically contain repeated code, but with oposite order of the saddr and daddr. This also seems like a somewhat clumsy solution to me that will also have to be repeated for different protocols (i.e. a function that for example parsed an identifier from a QUIC packet would also need a switch to fill in the port numbers correctly).
3. Reformat the network_tuple to something like this:
```
struct network_address {
	struct in6_addr ip;
	__u16 port;
}

struct network_tuple {
	struct network_address saddr;
	struct network_address daddr;
	__u16 proto;
	__u16 ipv;
};
```
Then the functions could simply have two pointers to `struct network_address` as parameters, and the TC and XDP program could simply pass those arguments in reverse order.

I'm currently leaning towards solution 3, but if you have any views of this I would like to hear them.

A separate question as well while I'm on it. If I've understood it correctly, BPF now supports function calls on modern kernels, so `__always_inline` isn't needed anymore. From what I can see from code in other folders of this repo you seem to mix a little bit between using `__always_inline` and not. Any rule of thumb for when one ought to use `__always_inline` or not? Use it for functions that are only called a single time (to avoid growing the instruction cache considerably)?

pping/pping_helpers.h
simosund: My bad, didn't refresh this page so didn't see your first reply until now.

pping/pping_kern_xdp.c
tohojo: Yeah, option 3 sounds reasonable (as long as you add a comment explaining why it's flipped at the call site). A "network address" is generally L3, though, so maybe call the struct `l4_address` or maybe `flow_address` ?

As for `__always_inline`, in kernel code the guidance is to generally let the compiler make inlining decisions and only override this if the compiler decision is demonstratively bad (e.g., if it's really performance-sensitive code where avoiding the CALL overhead is important). Generally, if you have a static function that is only called once, it will just get inlined automatically. At least for regular C code; not sure to what extent LLVM does this for BPF code...

pping/pping_helpers.h
simosund: Hmm, even with `__builtin_memcpy()` I get exactly the same error though.

tohojo: Works fine for me, with both versions of `memcpy()`. Which LLVM version are you using?

And what's the output of `llvm-objdump --reloc pping_kern_tc.o | fgrep -A 10 '[pping_egress]'` ?

simosund: I have LLVM version 11.0.0, and clang:
```
$ clang --version
Ubuntu clang version 11.0.0-2
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
If compiling the program with `map_ipv4_to_ipv6()` as (which results in previously shown error for me):
```
static __always_inline void map_ipv4_to_ipv6(__be32 ipv4, struct in6_addr *ipv6)
{
	__u16 ipv4_prefix[6] = {0x0, 0x0, 0x0, 0x0, 0x0, 0xffff};
	__builtin_memcpy(ipv6, ipv4_prefix, sizeof(ipv4_prefix)); // Won't load on TC
	/* memset(&ipv6->in6_u.u6_addr8[0], 0x00, 10); */
	/* memset(&ipv6->in6_u.u6_addr8[10], 0xff, 2); */
	ipv6->in6_u.u6_addr32[3] = ipv4;
}
```
... the output of the command is:
```
$ llvm-objdump --reloc pping_kern_tc.o | fgrep -A 10 '[pping_egress]'
RELOCATION RECORDS FOR [pping_egress]:
OFFSET           TYPE                     VALUE
00000000000002e0 R_BPF_64_64              .rodata
0000000000000d60 R_BPF_64_64              ts_start

RELOCATION RECORDS FOR [.debug_info]:
OFFSET           TYPE                     VALUE
0000000000000006 R_BPF_64_32              .debug_abbrev
000000000000000c R_BPF_64_32              
0000000000000012 R_BPF_64_32              
0000000000000016 R_BPF_64_32              .debug_line

```

simosund: Could it be related to me not having iproute with libbpf support again? Cilium issue [#7446](https://github.com/cilium/cilium/issues/7446) to get a similar error message and there the fix is apparently to use a different version of iproute.

tohojo: Well since you're using BTF-defined maps you can't load anything at all without a libbpf-based iproute2, can you? Doesn't work for me even on the master branch without it...

And indeed that error message seems to come from the iproute2 code, so yeah, that's most likely your issue - but I would expect it to happen regardless of what you put into map_ipv4_to_ipv6() then?

simosund: Well, I only use BTF-defined maps for the TC program in case `HAVE_TC_LIBBPF` is defined (which it should only be if the `configure` script detects that tc has libbpf support. Otherwise I fall back on old `struct bpf_elf_map`. And that part at least partly seems to work as I can run the program despite not having libbpf support in tc, and you can apparently run it with libbpf supported iproute (although for some reason it doesn't work for you without libbpf support?). 

So the memset version (both using `memset()` and `__builtin_memset()`) version works for me. But if I comment it out and replace with the memcpy version ´(either `memcpy()` or `__builtin_memcpy()`) instead, I get that error.

tohojo: Ah! Figured out what's going on (after re-running configure I can reproduce the same error you get):

```
$ sudo ./pping testns                        
ELF contains non-{map,call} related relo data in entry 0 pointing to section 5! Compiler bug?!
Error fetching program/map!
Unable to load program
ERROR: Exec error(1) occurred cmd: "/sbin/tc filter add dev testns pref 2 handle 2 egress bpf da obj pping_kern_tc.o sec pping_egress"
Could not load section pping_egress of pping_kern_tc.o on interface testns: Unknown error -3
$ llvm-objdump --section-headers pping_kern_tc.o       

pping_kern_tc.o:	file format elf64-bpf

Sections:
Idx Name             Size     VMA              Type
  0                  00000000 0000000000000000 
  1 .text            00000000 0000000000000000 TEXT
  2 pping_egress     00000d88 0000000000000000 TEXT
  3 license          00000004 0000000000000000 DATA
  4 maps             00000024 0000000000000000 DATA
  5 .rodata          0000000c 0000000000000000 DATA
  6 .debug_loc       00000aec 0000000000000000 
  7 .debug_abbrev    00000261 0000000000000000 
  8 .debug_info      00000fa1 0000000000000000 
  9 .debug_ranges    000002d0 0000000000000000 
 10 .debug_str       0000081d 0000000000000000 
 11 .BTF.ext         00000c30 0000000000000000 
 12 .debug_frame     00000028 0000000000000000 
 13 .debug_line      00000619 0000000000000000 
 14 .symtab          00001a88 0000000000000000 
 15 .relpping_egress 00000020 0000000000000000 
 16 .rel.debug_info  00001480 0000000000000000 
 17 .rel.BTF.ext     00000c00 0000000000000000 
 18 .rel.debug_frame 00000020 0000000000000000 
 19 .rel.debug_line  00000010 0000000000000000 
 20 .strtab          00000278 0000000000000000 
```

The error message is a hint - "pointing to section 5", which is .rodata. So this is not related to memcpy(); what's happening is that the constant (`{0x0, 0x0, 0x0, 0x0, 0x0, 0xffff}`) gets put into `.rodata` and copied from there. And the iproute2 BPF loader doesn't know how to deal with that, so it bugs out. Whereas `memset()` just gets compiled into a couple of instructions that encode the values directly in them, so it doesn't require relocations.

I guess you might as well keep the `memset()` implementation, then, for compatibility...

simosund: Ah, that makes sense I guess (although I admitdly don't really know that well in what ways the handling of BTF program differ between libbpf and iproute2, need to read up more on it). Was very confused why memset would work but not memcpy when they both wrote data to the same place, and completely overlooked the real problem (the place they wrote data from). But guess I'll go with memset implementation then and remove the commented out memcpy version.

Have also refactored the TC and XDP programs now, so when you have time feel free to have a look at that. I'm not entierly happy with `parse_packet_identifier` taking pointers to both the `packet_id` and to `saddr` and `daddr` (which are part of the `packet_id`), but that is how it ended up for now to make it convinent to swap place on `saddr` and `daddr`.

pping/pping_helpers.h
```diff
@@ -73,5 +80,80 @@ static __always_inline int parse_tcp_ts(struct tcphdr *tcph, void *data_end,
...
+ * Attempts to parse the packet limited by the data and data_end pointers,
+ * to retrieve a protocol dependent packet identifier. If sucessful, the
+ * ipv and identifier of p_id will be set, saddr and daddr (which may be part
+ * of p_id) will be filled with the source and destionation addresses of the
+ * packet, and 0 will be returned. On failure, -1 will be returned.
+ */
+static int parse_packet_identifier(void *data, void *data_end, bool is_egress,
+				   struct packet_id *p_id,
+				   struct flow_address *saddr,
+				   struct flow_address *daddr)
```
tohojo, line 118: You're passing in both the packet_id struct, and two members from it. Since you also have the bool explicitly indicating the direction you could just skip the flow_address pointers and access them through p_id (with the 'reverse' logic being in this function)?

Also, I think this only works because the compiler inlines the function; BPF function calls are limited to five parameters :)

pping/pping_helpers.h
```diff
@@ -73,5 +80,80 @@ static __always_inline int parse_tcp_ts(struct tcphdr *tcph, void *data_end,
...
+
+	// Parse IPv4/6 header
+	if (proto == bpf_htons(ETH_P_IP)) {
+		p_id->flow.ipv = AF_INET;
+		proto = parse_iphdr(&nh, data_end, &iph);
+	} else if (proto == bpf_htons(ETH_P_IPV6)) {
+		p_id->flow.ipv = AF_INET6;
+		proto = parse_ip6hdr(&nh, data_end, &ip6h);
+	} else
+		return -1;
```
tohojo, line 136: Nit: When you have an if/else structure like this where the first clauses have brackets it's easy to forget to add them if you ever do add something to the 'else' clause. So better to put brackets on the else from the beginning, even though it only has a single statement. Makes the code look more symmetrical, too :)

pping/pping_helpers.h
```diff
@@ -73,5 +80,80 @@ static __always_inline int parse_tcp_ts(struct tcphdr *tcph, void *data_end,
...
+
+	// Add new protocols here
+	if (proto == IPPROTO_TCP)
+		err = parse_tcp_identifier(&nh, data_end, is_egress, saddr,
+					   daddr, &p_id->identifier);
+	else
+		return -1;
+
+	if (err)
+		return -1;
```
tohojo, line 146: These two conditions with identical `return -1` look a bit odd. I think it's better to move the if(err) check up to right after the call to parse_tcp_identifier(). That makes it obvious which err you are reacting to, and makes it harder to accidentally move them further apart and override err when you change the code later.

tohojo (changes requested):
A few suggestions for improving the refactor, and a couple of style nits...

pping/pping_helpers.h
simosund: Yhea, as I mentioned I wasn't entierly happy with sending pointers to both the `packet_id` and the two `flow_address` members. Using `is_egress` for that actually makes a lot of sense. 

Ah right, knew that BPF function calls were limited to 5 parameters (had some run-ins with `bpf_printk()` about that), but didn't think about it when writing my own functions :).

pping/pping_helpers.h
simosund: Yhea, and is part of the linux kernel style guide as well I recall correctly.

pping/pping_helpers.h
simosund: Yhea, I agree that it looks a bit odd. My rationale was that as more protocols were added, it would be repetative to put an `if (err)` clause for each one. Although guess that I can put it up there for now, and then change that if and when new protocols actually are added.

simosund:
Hi again @tohojo.
I've addressed your previous points now, however I just noticed that `parse_tcp_identifier()` also has 6 parmaters (so probably only works due to compiler inlining). In this case I'm not really sure how to get rid of a parameter. 

1. I could do it in a similar manner as for `parse_packet_identifier()` and send in the whole `packet_identifier` (or at least the `network_tuple`), and then use the `is_egress` to perform the saddr/daddr swap. But then I fall back into having to replicate the swap-logic for every protocol which is what I wanted to get away from by beaking up the addresses into `flow_address` structs in the first place.
2. I could skip sending in a pointer to an identifier and instead have it as a return value. Then I however need to find a good value that would indicate error (and not be a valid identifier value for any protocol).
3. I could create another struct that just holds pointers to two ports, and have `parse_packet_identifier()` to fill in that, but that also feels a bit clumsy to me.
4. Declare this function as __always_inline to ensure compiler inlines it (and therefore not having to worry about number of parameters).

None of these solutions seem ideal to me, so would happily hear if you have any better ideas, or an opinion of which of these options are least horrible.

simosund:
I suppose a 5th option to get rid of a parameter is to use the `parsing_context` struct I've added in [this](https://github.com/simosund/bpf-examples/commit/219e962832367963b92cc27e7c8687db4048cfb7) commit (left in on a different branch so I can just wipe it of existance if it was a horrible idea). 

The main idea of the commit was to try and solve an issue with pure TCP ACKs being used as timestamp entries (as the "reply" to those may not represent the RTT). I introduced the `parsing_context` struct as I noticed that the `data_end` pointer in TC seems to point to the end of the packet headers (and not the end of the entire packet, as for XDP), which made it a bit problematic to detect if it was a pure ACK when parsing the TCP header from TC. So introduced a `data_end_end` pointer (brilliant name, I know). However, as a side effect of the `parsing_context` containing both the `hdr_cursor` and the `data_end` pointer, it frees up one parameter.

tohojo:
Well, the usual way of avoiding too many function parameters is to use a struct instead. I don't think the "parse context" idea is terrible; made a few comments on that commit... :)

If you want to return the identifier you could make the return type u64 and return -1 (which becomes 0xffffffffffffffff) on error?

pping/pping_kern_xdp.c
```diff
@@ -29,54 +21,39 @@ struct {
...
-	proto = parse_iphdr(&nh, data_end, &iph);
-	if (proto != IPPROTO_TCP)
-		goto end;
-	proto = parse_tcphdr(&nh, data_end, &tcph);
-	if (proto < 0)
-		goto end;
+	pctx.data = (void *)(long)ctx->data;
+	pctx.data_end = (void *)(long)ctx->data_end;
+	pctx.len = pctx.data_end - pctx.data;
+	pctx.nh.pos = pctx.data;
```
tohojo, line 36: If you initialise this with bracket-style initialisers you don't have to worry about forgetting to zero out any new members you may add in the future. I.e.,:

```C
struct parsing_context pctx = {
	.data = (void *)(long)ctx->data;
	.data_end = (void *)(long)ctx->data_end,
	.len = pctx.data_end - pctx.data,
	.nh = { .pos = pctx.data, },
};
```

also, pkt_len instead of len? :)

simosund: You're fast. Was just sitting writing a message to you :). But yhea, kinda missed your naming suggestions before, but think `pkt_len` is clearer (i.e it's not the length of the safe area to read or anything else). A bit more unsure about `pkt_start` vs `data` though. While `pkt_start` is a lot more descriptive in my world, `data` is sort of what it seems to be called everywhere else.

tohojo: Yeah, I can see your point about keeping data named the same; fine with me. Let's not bikeshed this any further ;)

tohojo (approved):
LGTM - let's merge and add the next features as a separate PR :)

PR #13, opened 2021-02-25
Add sampling to pping

simosund:
After some thinking and a bit of discussion with @abrunstrom and @sferlin, I've found that there are a number of potential issues that needs to be solved in order to get sampling to work well. Before I start actually implementing the sampling, I would like to get some feedback on if my plan for how to do the sampling makes sense. Therefore I've started documenting some of the challenges and how I plan to solve them. And instead of sending a bunch of mails back and fourth, I figured we could simply have that discussion in this PR instead.

So this PR doesn't contain any code changes (yet), only a markdown document (`SAMPLING_DESIGN.md`) which I hope can serve as the basis for some discussion. I still have some sections left which I plan to fill in tomorrow, and then we can maybe discuss some of these issues on the next meeting.


tohojo:
Immediate thoughts: The "keep track of the last seen identifier and sample the first one seen" was what immediately came to mind to me as well, so that makes sense!

As for the possibility of "missing" an event, one option could be to sample in a "window". I.e., when you figure out that it's time to do a new sample, you save the timestamps of the next (say) 3 identifiers. And when the replies come back you also rate limit the output so you'll end up only outputing one of those three samples when they come back (since they should be close together).

As for sampling frequency, I'd say start with the simple thing: A fixed interval that defaults to something sensible but which can be overridden at the command line. And then add dynamic RTT-based sampling as the next thing :)

simosund:
Thanks for your comments @tohojo. I've also been thinking about allowing a timestamping a few consecutive packets, however I didn't think about also rate-limiting the output. So that will probably be a good addition to make as well. 

Starting simple with the sampling frequency sounds like a good choice. Overall, I will probably start of with something pretty basic, and then add more of the features I discuss after hand (as you approve of the basics :)). But still wanted to outline the overall requirements for this whole sampling thing early on, so that I don't continue building on that basic thing, only to discover after a while that it has some fatal flaw and I have to start all over (guess there's still a risk for that, but can at least try to minimize it somewhat).

sferlin:
I was thinking a bit about the sampling as well (head sampling, i.e., sample some consecutive packets at the beginning of a flow), a time-based window sampling (this one is more intuitive and easier to overwrite per cmd line). About the output I have not thought about it. maybe I missed something, the benefit of it.

tohojo:
See also Kathie's visualisation tool for pping: https://github.com/pollere/ppviz - haven't played with it yet, but may be worth keeping in mind...

simosund:
Yhea, saw that visualization tool from the email-conversation you passed along. Haven't tried it out yet either, but might be useful to keep the same "machine readable" format as her pping does so one can use that tool with both pping implementations (or at least have it as an option, we might want to output some additional information, like protocol, in the future).

simosund:
Have pushed a very early implementation of the per-flow rate limitation now. This still lacks many key components that I will add on later (see commit message). However, first I plan to update the documentation with some of the outcomes from today's discussion. 

Also, before going further I also think I should attempt the fix @tohojo previously suggested to always load the TC program with libbpf. If I've understood it correctly, the idea is to let the current userspace program (which loads and attaches the XDP program) also load and pin the TC program using libbpf. Then it simply uses tc to attach the pinned BPF program. This way, it should have libbpf support regardless of if iproute has libbpf support or not, and I won't have to deal with the dual case iproute with/without libbpf support anymore (which in this commit leads to two definitions of the flow_state map).

pping/pping_kern_tc.c
```diff
@@ -6,44 +6,119 @@
...
-	struct packet_timestamp p_ts = { 0 };
+	__u64 p_ts;
 	struct parsing_context pctx = {
 		.data = (void *)(long)skb->data,
 		.data_end = (void *)(long)skb->data_end,
 		.pkt_len = skb->len,
 		.nh = { .pos = pctx.data },
 	};
+	struct flow_state *f_state;
+	struct flow_state new_state = { 0 }; // Rarely-ish used, but can't really dynamically allocate memory or?
```
tohojo, line 64: No, you can't. But allocating this on the heap would be inefficient anyway since it's such a small value :)

pping/pping_kern_tc.c
```diff
@@ -6,44 +6,119 @@
...
+	}
+
+	// Check if identfier is new
+	/* The gap between checking and updating last_id may cause concurrency
+	 * issues where multiple packets may simultaneously think they are the
+	 * first with a new identifier. As long as all of the identifiers are
+	 * the same though, only one should be able to create a timestamp entry.
+
+	 * A bigger issue is that older identifiers (for example due to
+         * out-of-order packets) may pass this check and update the current
```
tohojo, line 86: Something off with the indentation here...

pping/pping_kern_tc.c
```diff
@@ -6,44 +6,119 @@
...
 
-	p_ts.timestamp = bpf_ktime_get_ns(); // or bpf_ktime_get_boot_ns
+	// Check flow state
+	f_state = bpf_map_lookup_elem(&flow_state, &p_id.flow);
+	if (!f_state) { // No previous state - attempt to create it
+		bpf_map_update_elem(&flow_state, &p_id.flow, &new_state,
+				    BPF_NOEXIST);
+		f_state = bpf_map_lookup_elem(&flow_state, &p_id.flow);
+		if (!f_state)
+			goto end;
```
tohojo, line 76: The convention in the kernel is generally to use 'out' for these kinds of 'end of function' labels. You're being consistent with 'end', though, and obviously I haven't noticed before, but just thought I'd mention it since I did notice this time ;)

pping/pping_kern_tc.c
```diff
@@ -6,44 +6,119 @@
 #include "pping.h"
 #include "pping_helpers.h"
 
+#define RATE_LIMIT                                                             \
+	100000000UL // 100ms. Temporary solution, should be set by userspace
+#define BURST_DURATION                                                         \
+	4000000 // 4ms, duration for when it may burst packet timestamps
+#define BURST_SIZE                                                             \
+	3 // Number of packets it may create timestamps for in a burst
```
tohojo, line 14: Maybe don't define these until you're actually using them?

pping/pping_kern_tc.c
```diff
@@ -6,44 +6,119 @@
...
+
 char _license[] SEC("license") = "GPL";
 
 #ifdef HAVE_TC_LIBBPF /* detected by configure script in config.mk */
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(key_size, sizeof(struct packet_id));
-	__uint(value_size, sizeof(struct packet_timestamp));
+	__uint(value_size, sizeof(__u64));
```
tohojo, line 22: Oh, BTW, when you're using BTF-defined maps instead of value_size and key_size, you can use __type(key, struct packet_id) and __type(value, __u64) (IIRC that's the syntax), which will also encode the actual types into the BTF information that goes into the kernel with the map, which makes it possible for bpftool to print out the types, for instance...

pping/pping_kern_tc.c
simosund: Hmm, good catch. Not sure how it got there....thought I ran clang-format on the entire region I modified.

pping/pping_kern_tc.c
simosund: Ok, good to know. Have fixed it in the latest commit.

pping/pping_kern_tc.c
simosund: True, just got a bit ahead of myself.

pping/pping_kern_tc.c
simosund: Right, I remember seeing that somewhere (possibly in traffic-pacing-edt example), but forgot it once writing it myself. Have addressed it for all maps except the perf-buffer one which didn't seem too happy about it (guess __u32 might not be the correct type, just right size).

pping/pping.c
```diff
@@ -78,20 +78,20 @@ static int set_rlimit(long int lim)
...
+/* 			return -errno; */
 
-		return mkdir(path, 0700) ? -errno : 0;
-	}
-	return S_ISDIR(st.st_mode) ? 0 : -EEXIST;
-}
+/* 		return mkdir(path, 0700) ? -errno : 0; */
+/* 	} */
+/* 	return S_ISDIR(st.st_mode) ? 0 : -EEXIST; */
+/* } */
```
tohojo, line 94: Just remove code instead of commenting it out - it can always be resurrected from the git history...

pping/pping.c
```diff
@@ -419,29 +429,47 @@ int main(int argc, char *argv[])
...
+			fprintf(stderr,
+				"Failed unpinning tc program from %s: %s\n",
+				PINNED_DIR, strerror(-err));
+	}
+
+	/* 
+	 * Could use bpf_obj__unpin_maps(obj, PINNED_DIR) if it only tried
+	 * unpinning pinned maps. But as it also attempts (and fails) to unpin
+	 * maps that aren't pinned, will instead manually unpin the one pinned
+	 * map for now.
```
tohojo, line 461: Should be possible to call it with a NULL second argument, in which case it will only try to unpin the maps that are marked as pinned (so as long as it's the same bpf_object you originally pinned the maps on it should just work).

tohojo (changes requested):

pping/pping.c
simosund: Oops, forgot to clean that up. Often temporarily comment out some code while working on it, but missed removing it before the commit in this case.

pping/pping.c
simosund: Completely missed that possibility. In that case that would be a much nicer solution.

simosund:
Hi @tohojo,
Have pushed some new commits that attempts to add cleanup of the flow state map. The first commit adds the periodic userspace cleanup, which seems to work. The second commit contains an initial implementation of the BPF side cleanup, which attemps to delete flow state if the packet parsing detects that the flow is being closed (FIN or RST flags).

I couldn't get it to work on the XDP side however, as the verifier started complaining then. I'm still not very good at figuring out what the verifier finds problematic, and being tired/frustrated probably doesn't help. When the verifier also started to complain simply from adding a bpf_printk statement that didn't access packet data at all I decided to give up for the day. So if you have the time and energy, it would be great if you could take a look at it (I have commented out the section I have problem with in `pping_kern_xdp.c`, as well as the printk line in `pping_helpers.h/parse_tcp_identifier()`.

tohojo:
Simon Sundberg <notifications@github.com> writes:

> Hi @tohojo,

> Have pushed some new commits that attempts to add cleanup of the flow
> state map. The first commit adds the periodic userspace cleanup, which
> seems to work. The second commit contains an initial implementation of
> the BPF side cleanup, which attemps to delete flow state if the packet
> parsing detects that the flow is being closed (FIN or RST flags).
>
> I couldn't get it to work on the XDP side however, as the verifier
> started complaining then. I'm still not very good at figuring out what
> the verifier finds problematic, and being tired/frustrated probably
> doesn't help. When the verifier also started to complain simply from
> adding a bpf_printk statement that didn't access packet data at all I
> decided to give up for the day. So if you have the time and energy, it
> would be great if you could take a look at it (I have commented out
> the section I have problem with in `pping_kern_xdp.c`, as well as the
> printk line in `pping_helpers.h/parse_tcp_identifier()`.

Hmm, that's a really good question! I can see why you're frustrated...
As far as I can tell, the immediate cause of the trouble is that
including the bpf_printk() statement causes the compiler to allocate
registers differently, which means that some of the later code uses the
r7 register instead of the r1 register to store the packet offsets they
are loading from. And this means that the verifier suddenly thinks that
those packet reads are out of bounds.

Not sure if this is a verifier bug, or a compiler bug, or both. Or if
there is some other subtle bug lurking somewhere that just gets
triggered by this. But at least the above is an explanation :)

The bpf_map_delete() by itself seems to work fine for me if I leave only
the bpf_printk commented out...

simosund:
Hi again @tohojo,
Recently just pushed an attempt at a "fix" for the problem that I ran into (but that you apparently didn't?) when attempting to delete the flow-state from the XDP program (but without the prink-statement that breaks the program). I've found [this blog-post](https://blog.path.net/ebpf-xdp-and-network-security/) that seemed to run into a similar problem as me, which is apparently caused by the verifier to forget constraints for variables it retrieves from the stack. So if I understand it correctly, when `opt_size` (in `pping_helpers.h/parse_tcp_ts()`) is fetched from the stack at 297, the verifier no longer knows it's a u8, and think it might contain some negative value that puts us outside of the packet. 

So my attempt at a fix was to add a little extra check to try and convince the verifier that opt_size can't contain a negative value, (and limited the upper bound to 34 as that's the largest legal size for a TCP option as far as I can tell, simply putting it at 0xff will cause the compiler to skip the check as it's already a u8). But then I run into a new problem of the program suddenly being too large (over 1000001 instructions). I find this a bit strange as I don't feel like my program should have that many paths? 

Think I might somehow have gotten stuck in an infinite loop here. As just after pushing this commit I noticed that if I put `#pragma unroll` in front of the loop it works (as in the program isn't too large anymore, but if I un-comment the line that deletes the flow-state in the XDP program I still run into the same issue, so my fix didn't really fix anything). 

So if you have the time, could you check if this fix also causes the verifier to reject the program as being too big for you as well? And if so, any idea why the not-unrolled loop would become an infinite loop (or at least much larger than the unrolled version)? Finally, any idea how I should solve this whole mess?

simosund:
With last commit ([e04284a](https://github.com/xdp-project/bpf-examples/pull/13/commits/e04284a3ee9ab893611c6b171d6d5a571282c7bd)) I've added  a check (or rather a bitwise AND) that seems to work for convincing the verifier that `opt_size` must be positive. So now uncommenting the state-info cleanup from the XDP program works for me (even on my ancient 5.8 kernel).

Adding this does however cause the "infinite loop" problem we've talked about (at least for me), so until we've figured that out I've temporarily reverted to unrolling the loop. Will later on squash some of these commits to avoid this whole detour, but figured I'll leave them for now while they're still somewhat relevant to our discussion.

Also, as a note to myself, I must figure out what the `MAX_TCP_OPTIONS` value should really be, the 10 is just a temporary value I made up a while back and that seems to have worked alright. Need to sit down and figure out what the maximum number of possible TCP options can really be.

pping/pping_helpers.h
```diff
@@ -107,8 +107,17 @@ static int parse_tcp_ts(struct tcphdr *tcph, void *data_end, __u32 *tsval,
...
+		 * The &= is to make sure that verifier knows opt_size must
+		 * be positive, as verifier may not remeber u8 constraint if
+		 * it fetches variable from stack, see ex
+		 * https://blog.path.net/ebpf-xdp-and-network-security/.
+		 * Use 0x3f instead of 0xff to avoid compiler optimizing
+		 * away the check, and no option should be able to have a
+		 * size of above 63 anyways (maximum TCP header size is 64
+		 * bytes, and no more than 64-20=44 of them can by options).
+		 */
+		opt_size &= 0x3f;
```
simosund, line 120: Hmm, without this line it seems like the compiler unrolls the loop (even without the `#pragma unroll`). So I guess that partly explains why this line seems to cause an infinite loop, it is the line that apparently causes the compiler to create an actual loop.

That said, I still have no idea why it actually becomes an infinite loop (or why it's the breakpoint that causes the compiler to stop unrolling the loop).

simosund: The "workaround" from https://lore.kernel.org/bpf/20210226174800.2928132-1-yhs@fb.com/ seemed to just reduce the number of loop iterations. I just tried that as well to see what happens, and turns out that if I reduce `TCP_MAX_OPTIONS` to 9 the compiler unrolls the loop. So apparently I'm right on the edge of when the compiler decides to not unroll the loop. 

Still no clearer why the verifier thinks the not-unrolled loop is infinite though. If I understood the explanation in the above patch notes correctly, the issue was that the loop variable was saved to the stack, but as far as I can tell that doesn't appear to happen here. I have tried to check through the compiled bytecode, and don't see a clear issue there. But then I'm no expert at debugging BPF bytecode, and find it somewhat hard to keep track of what's being kept in each register at each step.

simosund:
Hi @tohojo,
Instead of getting stuck on the infinite loop thing, I've moved on a little bit with adding support for user-configurable rate-limit. So feel free to take a look at that.

Guess I'll have to get back to the infinite loop thing later, and start going through the 4.7 million lines of output the verifier spews out to see if I can make sense of any of it.

tohojo:

> Instead of getting stuck on the infinite loop thing, I've moved on a
> little bit with adding support for user-configurable rate-limit. So
> feel free to take a look at that.

Good idea! No need to block everything else on that one issue!

> Guess I'll have to get back to the infinite loop thing later, and
> start going through the 4.7 million lines of output the verifier spews
> out to see if I can make sense of any of it.

I am also planning to take a closer look at this, but I still have a few
other things I need to finish first. Hopefully I'll get around to it
sometime this week (along with looking at this latest round of changes) :)

pping/pping.c
```diff
@@ -125,6 +164,22 @@ static int xdp_attach(struct bpf_object *obj, const char *sec, int ifindex,
...
 }
 
+static int init_rodata(struct bpf_object *obj, void *src, size_t size)
+{
+	char map_name[16];
+	struct bpf_map *rodata_map = NULL;
+
+	strncpy(map_name, bpf_object__name(obj), 8);
+	map_name[8] = '\0'; // Ensure null-byte at truncation point
+	strcat(map_name, ".rodata");
```
tohojo, line 174: Constructing the name like this is a bit brittle, especially since there are plans to change the name generated by libbpf. I think it's better to loop through the maps with bpf_map__next() and checking if the name contains .rodata (with strstr)...

pping/pping.c
```diff
@@ -368,6 +461,13 @@ int main(int argc, char *argv[])
 		goto cleanup;
 	}
 
+	/* err = init_rodata(xdp_obj, &config, sizeof(config)); */
+	/* if (err) { */
+	/* 	fprintf(stderr, "Failed pushing user-configration to %s: %s\n", */
+	/* 		PPING_XDP_OBJ, strerror(-err)); */
+	/* 	goto cleanup; */
+	/* } */
```
tohojo, line 469: Why is this disabled?

tohojo (changes requested):
Basic approach is sound, just a few nits :)

pping/pping.c
simosund: Hmm, good suggestion, will do that. That should be a much more robust solution (as long as the new name generated by libbpf still at least contain's ".rodata" that is).

pping/pping.c
simosund: It's disabled because the XDP doesn't actually use the config variable yet. So the XDP program doesn't seem to contain a .rodata map right now and therefore that part will always fail. Spontaneously I find that somewhat strange as the config variable is declared as volatile so the compiler shouldn't be able to optimize it away. But guess libbpf figures out it never uses it and thus skips creating a .rodata map? 

So that code is some remains that I once again failed to clean up properly. In the process of figuring out how to initialize global variable from user-space I printed out the `config.rate_limit` value from both the XDP and TC programs (to see that it was set successfully in both). Once I removed the printout from the XDP program libbpf could no longer find any .rodata map for it. So once I get to implementing the burst capabilities I will likely re-enable it (as then the XDP program needs to know the rate-limit in order to implement the output rate-limit).

simosund:
Hi @tohojo,
Found that by simply declaring opt_size as volatile, it will always be read from stack as a u8 (i.e `rX = *(u8 *)(r10 - Y)`). So no longer need bitwise AND to convince the verifier that it's a u8. Although of course nothing is ever straight forward so instead then ran into an issue with `if (pos + opt_size > opt_end && pos + opt_size > data_end)`. When opt_size wasn't volatile the compiler would just translate the `pos + opt_size` to `pos + 10`. So simply changed to `pos + 10` myself to avoid the verifier crying about invalid packet access (as it due to its goldfish memory can't remember that opt_size has to be 10 there due to being inside an `if (opt_size == 10)` statement).

Although now after I've pushed this, I now start to wonder if you didn't have a better idea during a previous meeting. To simply change opt_size to a u16 (or u32 or u64 for that matter, as the registers are 64 bit anyways), and do the &=0xff (at which point the compiler shouldn't optimize it away). While that should force a bitwise AND operation, it might avoid constantly reading opt_size from the stack, and guessing a bitwise AND should be a lot faster than a memory access. Although it might also end up reading it from the stack anyways, in which case you just add the bitwise AND operation without any gain.

abrunstrom:
I can envision an interesting preface about your friendship with the verifier in your thesis :)

tohojo:
abrunstrom ***@***.***> writes:

> I can envision an interesting preface about your friendship with the
> verifier in your thesis :)

Heh, yeah. I've been beating my head against it for a couple of hours
now; think I'm just going to give up for the time being :/

One thing I did notice while doing this, though, is that I don't think
your code correctly handles the case where opt_size is 0 after you read
it from the packet (or at least it is only doing so incidentally by
exhausting the loop)?

simosund:
> One thing I did notice while doing this, though, is that I don't think your code correctly handles the case where opt_size is 0 after you read it from the packet (or at least it is only doing so incidentally by exhausting the loop)?

Yhea, you are right about that. While `opt_size` _should_ never be 0, you can of course never know what you get in a packet. So as you say, my code right now should just loop over the same place until it's exhausted the loop. Guess the correct way to handle it would instead be to just terminate the loop immediately.

tohojo:

> Yhea, you are right about that. While `opt_size` _should_ never be 0,
> you can of course never know what you get in a packet. So as you say,
> my code right now should just loop over the same place until it's
> exhausted the loop. Guess the correct way to handle it would instead
> be to just terminate the loop immediately.

Yup, that works :)

pping/pping.c
```diff
@@ -184,48 +227,94 @@ static __u64 get_time_ns(void)
...
+#ifdef DEBUG
 	duration = get_time_ns() - now_nsec;
-	printf("Gone through %d entries and removed %d of them in %llu.%09llu s\n",
-	       entries, removed, duration / NS_PER_SECOND,
+	printf("%d: Gone through %d entries and removed %d of them in %llu.%09llu s\n",
+	       map_fd, entries, removed, duration / NS_PER_SECOND,
 	       duration % NS_PER_SECOND);
+#endif
+cleanup:
+	if (key)
```
tohojo, line 303: It's perfectly fine to call free() on a NULL value, so you can skip these checks...

pping/pping.c
```diff
@@ -184,48 +227,94 @@ static __u64 get_time_ns(void)
...
+	if (value)
+		free(value);
 	return removed;
 }
 
+/*
+ * Periodically cleans out entries from both the packet timestamp map and the
+ * flow state map. Maybe better to split up the cleaning of the maps into two
+ * separate threads instead, to better utilize multi-threading and allow for
+ * maps to be cleaned up at different intervals?
```
tohojo, line 316: I very much doubt the map cleaning is going to be a bottleneck, so adding threading is just likely to complicate things. If you do want to optimise this in the future, the right way would be to use the in-kernel map iterator functionality... :)

pping/pping_helpers.h
```diff
@@ -76,11 +98,12 @@ static int parse_tcp_ts(struct tcphdr *tcph, void *data_end, __u32 *tsval,
...
 			return -1;
 		opt_size = *(pos + 1);
+		if (opt_size < 2) // Stop parsing options if opt_size has an invalid value
+			return -1;
 
 		// Option-kind is TCP timestap (yey!)
 		if (opt == 8 && opt_size == 10) {
-			if (pos + opt_size > opt_end ||
-			    pos + opt_size > data_end)
+			if (pos + 10 > opt_end || pos + 10 > data_end)
```
tohojo, line 106: I think that if you add a check before the loop for opt_end <= data_end, you can avoid these double checks everywhere (i.e., just check against opt_end)...

tohojo (changes requested):
A few nits - also, your main() function is getting quite long, may be worth thinking about splitting it. For the BPF code, now that you're compiling and loading everything with libbpf, you could move the two functions into the same source file and avoid some code duplications (and the need for the pping_helpers.h file).

pping/pping_helpers.h
simosund: I think I tried that before but the verifier wasn't convinced. That said, it was quite some time ago and have changed some things around since then, so can try that again and see if the verifier is smart enough this time.

simosund:
Hi @tohojo, think my latest [commit](https://github.com/xdp-project/bpf-examples/pull/13/commits/93b6c0eafa18d13d44466ae0b3e17b64f7ed0289) adress most of your points. There is however one major problem with it, the verifier rejects it due to hitting the 1 million instruction limit. I find this very confusing as:
- The BPF-program code is the same as before, simply moved into a single file
- The loop is unrolled, so it can't really be an infinite loop. That said, it still seems to be connected to the loop, as lowering the loop count to 5 will make it work. After having a look at the generated BPF bytecode for 5 and 6 loops, I didn't see any major differences between them other than an additional part of the "loop", and some registers being changed around. Although from my experience I guess some registers being changed around can actually be part of the problem.
- The generated BPF bytecode is about 600 instructions for the common functions and about 100 instructions each for the separate BPF programs. Even if I call some bpf-helpers, I have a hard time seeing how it can reach 1 million instructions.

pping/pping_helpers.h
simosund: Had another try at this, but still can't seem to convince the verifier that if opt_end <= data_end, then anything < opt_end must also < data_end. So think the double checks will have to be there for now. 

Not sure if it's just that the verifier is too stupid, or if the opt_end get's placed on the stack and then the verifier forgets about its limitations.

simosund:
Will also need to update the documentation a bit to reflect the changes from this commit, as incorrect documentation is worse than no documentation at all. That said, I want to hold off on it for a little while in case we need to do some other restructure to solve this "infinite loop" (or whatever that causes the verifier to think the program is so large) problem.

tohojo:
AFAICT what happens is that the compiler shuffles the code around in a way that includes a backwards jump even though the loop gets unrolled. And so the verifier still needs to trace all the jumps around, can't convince itself that the program exits, and ends up exiting due to the instruction limit.

I think we may be at the point where we need to post this on the upstream list and see if someone there knows how the verifier can be made smarter, or if there's a way to convince the compiler to emit code that passes. Do you think you could produce a minimal self-contained example suitable for posting upstream? :)

simosund:
> AFAICT what happens is that the compiler shuffles the code around in a way that includes a backwards jump even though the loop gets unrolled. And so the verifier still needs to trace all the jumps around, can't convince itself that the program exits, and ends up exiting due to the instruction limit.

Makes sense I guess, there are indeed some backward jumps in the bytecode. That said, those backjumps also exist for the 5 loop version, so guess it just it just gets too complex to track the jumps for the verifier after 6 loops? It also works fine for the XDP program (if I basically just comment out all the code in the tc program) which should hit the same backward jumps, so not sure what differeniates them that's relevant here.

> I think we may be at the point where we need to post this on the upstream list and see if someone there knows how the verifier can be made smarter, or if there's a way to convince the compiler to emit code that passes. Do you think you could produce a minimal self-contained example suitable for posting upstream? :)

I can try, but as I'm not sure exactly what's triggering the problem I don't really know how to recreate it in a minimal fashion. I guess some of the key ingredients are two programs in the same file sharing a common function with a loop, but there's also some secret sauce which I don't really know what it is.

Is there any way to get the verifier output even when it passes a program? Would be interesting to know how many instructions it exhausts on the XDP program, as well as when only using 5 loops. See if those are anywhere near 1 million (based on the much shorter loading times I suspect not).

tohojo:

>> AFAICT what happens is that the compiler shuffles the code around in
>> a way that includes a backwards jump even though the loop gets
>> unrolled. And so the verifier still needs to trace all the jumps
>> around, can't convince itself that the program exits, and ends up
>> exiting due to the instruction limit.
>
> Makes sense I guess, there are indeed some backward jumps in the
> bytecode. That said, those backjumps also exist for the 5 loop
> version, so guess it just it just gets too complex to track the jumps
> for the verifier after 6 loops? It also works fine for the XDP program
> (if I basically just comment out all the code in the tc program) which
> should hit the same backward jumps, so not sure what differeniates
> them that's relevant here.

Erm, no idea? I guess the context around the call to that function trips
the verifier for some reason?

>> I think we may be at the point where we need to post this on the
>> upstream list and see if someone there knows how the verifier can be
>> made smarter, or if there's a way to convince the compiler to emit
>> code that passes. Do you think you could produce a minimal
>> self-contained example suitable for posting upstream? :)
>
> I can try, but as I'm not sure exactly what's triggering the problem I
> don't really know how to recreate it in a minimal fashion. I guess
> some of the key ingredients are two programs in the same file sharing
> a common function with a loop, but there's also some secret sauce
> which I don't really know what it is.

"Minimal" doesn't have to mean "small". It should be as small as
possible, but no smaller. And it should be self-contained so you can put
it into a single email, which I guess means copying over the helpers
from parsing_helpers.h. So I'd start by doing this, and then delete as
much code as you can while still hitting the bug.

> Is there any way to get the verifier output even when it passes a
> program? Would be interesting to know how many instructions it
> exhausts on the XDP program, as well as when only using 5 loops. See
> if those are anywhere near 1 million (based on the much shorter
> loading times I suspect not).

Sure, the verifier will spit out the log in any case if requested to do
so. The trouble is that libbpf doesn't really expose this in a good way,
so you may have to hack the loading function. IIRC, it's enough to set
log_buf_size to a non-zero value at the top of load_program() in libbpf,
and set log_level in struct bpf_prog_load_attr (which means loading with
bpf_object__load_xattr() or bpf_prog_load_xattr().

simosund:
> Sure, the verifier will spit out the log in any case if requested to do
so. The trouble is that libbpf doesn't really expose this in a good way,
so you may have to hack the loading function. IIRC, it's enough to set
log_buf_size to a non-zero value at the top of load_program() in libbpf,
and set log_level in struct bpf_prog_load_attr (which means loading with
bpf_object__load_xattr() or bpf_prog_load_xattr().

Hmm, might toy around a bit with that later, as I guess that means I need to set the map pin-location in some different manner. For now though I simply put some of my bpftrace knowledge to use to fetch it from the verifier's `bpf_verifier_env` struct using a kprobe and kretprobe for `do_check()`.

tohojo:

>> Sure, the verifier will spit out the log in any case if requested to do
> so. The trouble is that libbpf doesn't really expose this in a good way,
> so you may have to hack the loading function. IIRC, it's enough to set
> log_buf_size to a non-zero value at the top of load_program() in libbpf,
> and set log_level in struct bpf_prog_load_attr (which means loading with
> bpf_object__load_xattr() or bpf_prog_load_xattr().
>
> Hmm, might toy around a bit with that later, as I guess that means I
> need to set the map pin-location in some different manner.

No, you can still open with open_opts() setting pin_root_path, and then
load with load_xattr... bpf_object__load() is literally just:

int bpf_object__load(struct bpf_object *obj)
{
	struct bpf_object_load_attr attr = {
		.obj = obj,
	};

	return bpf_object__load_xattr(&attr);
}


> For now though I simply put some of my bpftrace knowledge to use to
> fetch it from the verifier's `bpf_verifier_env` struct using a kprobe
> and kretprobe for `do_check()`.

That works too, I guess ;)

simosund:
> No, you can still open with open_opts() setting pin_root_path, and then
load with load_xattr... bpf_object__load() is literally just:

But for just getting the processed instructions (which was mainly what I was after), my bpftrace script seems to work fine as well. 

Anyways, seems like even before merged the two programs into a single part I wasn't too far away from hitting the verifier limit, as if comment out the entire XDP program (so the file only contains a single program) it still uses about 700k instructions to verify the tc program. So this is not really a new issue, in the sense that I would have hit it before as well if I just increased my loop count to say 13-15. When both the XDP and tc programs are present, it uses about 500k instructions to verify the XDP program and then hits the 1m limit for the tc program. And it's quite interesting how commenting out different parts of the code makes a differenence to the number of instructions the verifier needs to process.

For example, I've noted that if I remove the "switch" logic on saddr and daddr depending on if it's egress or ingress then the XDP and tc programs will use about the same amount of instructions (whereas with the switch logic the tc programs needs about twice the number of instructions processed). If I drop the IPv6 branch then falls down to just 8k for the XDP program and 16k for the tc program.

tohojo:

>> No, you can still open with open_opts() setting pin_root_path, and then
> load with load_xattr... bpf_object__load() is literally just:
>
> But for just getting the processed instructions (which was mainly what
> I was after), my bpftrace script seems to work fine as well.

Ah, you just wanted the number of processed instructions? Sure, bpftrace
will probably work better for that :)

> Anyways, seems like even before merged the two programs into a single
> part I wasn't too far away from hitting the verifier limit, as if
> comment out the entire XDP program (so the file only contains a single
> program) it still uses about 700k instructions to verify the tc
> program. So this is not really a new issue, in the sense that I would
> have hit it before as well if I just increased my loop count to say
> 13-15. When both the XDP and tc programs are present, it uses about
> 500k instructions to verify the XDP program and then hits the 1m limit
> for the tc program. And it's quite interesting how commenting out
> different parts of the code makes a differenence to the number of
> instructions the verifier needs to process.

Clever! :)

> For example, I've noted that if I remove the "switch" logic on saddr
> and daddr depending on if it's egress or ingress then the XDP and tc
> programs will use about the same amount of instructions (whereas with
> the switch logic the tc programs needs about twice the number of
> instructions processed). If I drop the IPv6 branch then falls down to
> just 8k for the XDP program and 16k for the tc program.

Ah! I guess you're running into some kind of combinatorial explosion
with the loop in skip_ip6hdrext(), then. You could try commenting that
out? Another thing to play with is removing the __always_inline to see
if you can get each function to actually end up as separate functions in
the BPF code. That *should* make the verifier look at each function in
isolation, which might also help :)

simosund:
> Ah! I guess you're running into some kind of combinatorial explosion 
with the loop in skip_ip6hdrext(), then.

Yep, I do indeed seem to be hitting some interesting combinatorial effect. 

> You could try commenting that out?

Commenting out or reducing the loop count for `skip_ip6hdrext()` greatly reduces the number of instruction the verifier needs to process. 

> Another thing to play with is removing the __always_inline to see
if you can get each function to actually end up as separate functions in
the BPF code. That *should* make the verifier look at each function in
isolation, which might also help :)

Unfortunately removing `__always_inline` seems to have no effect, the compiler still inlines everything (the generated BPF bytecode is identical). Would need a `__never_inline` instead =)

I have pushed a kinda minimal working example (or not working example I guess) to the `pping_mwe` folder on my branch [Create_minimal_example](https://github.com/simosund/bpf-examples/tree/Create_minimal_example). Feel free to have a look at that and let me know what you think. You can probably remove a few more parts of the code, but this was a bit of a compromise to still have a somewhat functioning semi-sensibe BPF program that I could experiment with. 

The question is what to do now? Should we still try to report something upstream, and in that case what? That having multiple (unrolled, non-nested) loops seems to cause the verifier to exhaust a very large amount of processed instructions? 

And how do we solve it in the meantime? Splitting it up into two separate programs would probably work for now, but based on my testing that still causes the verifier to process a rather large amount of instructions, so may still run into the same problem later on as I continue to add features to pping. Guess another option would be to lower the loop counts somewhat, as I think it's quite rare that packets would need to exhaust all of them anyways.

tohojo:
noinline does exist. You could try googling for the magical incantation for clang and trying that out?

simosund:
> noinline does exist. You could try googling for the magical incantation for clang and trying that out?

I've now tried `noinline` on the `skip_ip6hdrext()` and `parse_tcp_ts()` functions (the ones containing the loops), and it did help a little bit. But it was not enough to get under the 1 million instruction limit for the tc program (based on the instructions processed for the XDP program I'm guessing it would need about 1.14 million instructions). Could potentially try and apply it to other methods and see then, but not sure if it's a good idea for me to start telling the compiler to stop inlining a bunch of functions (that it would normally like to inline). 

Also tried a bit of the opposite, that is telling the compiler to inline everything (`__always_inline`). When the programs are kept in separate files it does inline everything, but when the programs are kept in the same file it calls `parse_packet_identifier()` as an actual function. So was hoping telling the compiler to simply inline it again would get me down below the 1 million limit, but seems like even when using `__always_inline` the compiler will still generate BPF bytecode where it's called like a function. I've compiled some of the options I've tried and the resulting number of instructions the verifier processed [here](https://github.com/simosund/bpf-examples/tree/Create_minimal_example/pping_mwe) if it is of interest. 

After some discussion with my supervisors we decided it would be good if I didn't get stuck on this too long, so thinking of implementing a bit of a quickfix for now and we can eventually come back to this in the future. I'm planning on just decreasing the maximum number of IPv6 header extensions parsed (they are apparently not very common). Some other options would be to split it up into two files again (largley undoing the last commit) or decreasing the maximum number of parsed TCP options, but think those come with more drawbacks. But if you have a different opinion I'm all ears.

Regarding reporting it upstream, I'm not sure if there's anything in particular to report anymore (that the verifier doesn't like multiple loops?). Another less formal alternative could be to just post it on the EBPF slack channel and see if people have any good ideas on how to get around this.

tohojo:

>> noinline does exist. You could try googling for the magical incantation for clang and trying that out?
>
> I've now tried `noinline` on the `skip_ip6hdrext()` and
> `parse_tcp_ts()` functions (the ones containing the loops), and it did
> help a little bit. But it was not enough to get under the 1 million
> instruction limit for the tc program (based on the instructions
> processed for the XDP program I'm guessing it would need about 1.14
> million instructions). Could potentially try and apply it to other
> methods and see then, but not sure if it's a good idea for me to start
> telling the compiler to stop inlining a bunch of functions (that it
> would normally like to inline).

Ah, too bad - was hoping that would be a quick fix ;)

> Also tried a bit of the opposite, that is telling the compiler to
> inline everything (`__always_inline`). When the programs are kept in
> separate files it does inline everything, but when the programs are
> kept in the same file it calls `parse_packet_identifier()` as an
> actual function. So was hoping telling the compiler to simply inline
> it again would get me down below the 1 million limit, but seems like
> even when using `__always_inline` the compiler will still generate BPF
> bytecode where it's called like a function. I've compiled some of the
> options I've tried and the resulting number of instructions the
> verifier processed
> [here](https://github.com/simosund/bpf-examples/tree/Create_minimal_example/pping_mwe)
> if it is of interest.

Wow, nice writeup! :)
Adding the kernel and compiler versions you're testing this on is the
only thing I can see that's missing for this to be a useful reference
for asking people; see below.

> After some discussion with my supervisors we decided it would be good
> if I didn't get stuck on this too long, so thinking of implementing a
> bit of a quickfix for now and we can eventually come back to this in
> the future. I'm planning on just decreasing the maximum number of IPv6
> header extensions parsed (they are apparently not very common). Some
> other options would be to split it up into two files again (largley
> undoing the last commit) or decreasing the maximum number of parsed
> TCP options, but think those come with more drawbacks. But if you have
> a different opinion I'm all ears.

Totally agreed, don't get held up on this; just lower one or both loop
counts and carry on!

> Regarding reporting it upstream, I'm not sure if there's anything in
> particular to report anymore (that the verifier doesn't like multiple
> loops?). Another less formal alternative could be to just post it on
> the EBPF slack channel and see if people have any good ideas on how to
> get around this.

Given that you've already created such a comprehensive writeup, I
definitely think you should at least ask upstream! You could do this
either on Slack or on the mailing list; or both? On Slack you can
probably get away with just putting in the link above, but on the
mailing list you probably need to do a little bit of a writeup. You
could just edit the existing README file a bit and turn it into an
email, then send that to ***@***.*** and
***@***.***

I guess posting the link on Slack is easiest, so try that first, then
send the email afterwards if you don't get any replies? :)

abrunstrom:
Nice summary Simon! I agree, send it on the Slack channel and let's move on to the output :)

simosund:
Hi again @tohojo. I've now reduced the number of IPv6 extensions parsed, so that the program loads again. So if you could have a final look at it we could perhaps get it merged in. I think I've addressed all of your requests (minus double checks for opt_end and data_end, which I can't seem to manage without), but there could of course be something more to do as well.

Now that we load both the BPF programs from the same object-file, I guess we don't need to pin the maps anymore? So guess I could also remove the code related to that (not that much, since libbpf takes care of most of it).

tohojo:
Yeah, good point about the pinning! I agree, let's get rid of it. We may need to bring it back later for when you enable background monitoring (in that case there won't necessarily have to be a long-running process at all if a cron job can just ping the maps), but for now it's just complicating stuff and incurring a dependency on bpffs...

Couldn't spot anything else to complain about, so I'm fine with merging that once you've removed the pinning :)

simosund:
> Yeah, good point about the pinning! I agree, let's get rid of it. We may need to bring it back later for when you enable background monitoring (in that case there won't necessarily have to be a long-running process at all if a cron job can just ping the maps), but for now it's just complicating stuff and incurring a dependency on bpffs...

Have now removed the code related to pinning the maps (as stated, not many lines). That said, we don't really get rid of the dependency on bpffs as we still pin the tc program (so we don't need to rely on tc loading the program, only attaching it).

tohojo:
Right, but libbpf will hopefully gain support for attaching the tc progs soon, so that can go away as well :)

PR #15, opened 2021-03-14
Added an eBPF Qdisc classifier example

freysteinn:
This example demonstrates how to write a simple eBPF Qdisc classifier
that classifies flows depending on their TCP port. The example comes
with two scripts, run-u32.sh and run-bpf.sh. Both scripts implement the
same functionality to make it easy to compare them. The scripts both
create two network namespaces called Left and Right, representing two
different hosts. The Left namespace loads a Qdisc classifier that rate
limits depending on the destination TCP port. It also starts iperf3
clients that connect to the Right namespace's iperf3 servers to
demonstrate the rate-limiting. The Qdisc classifiers give TCP ports 8080
and 8081 a high rate limit, while TCP port 8082 represents all other
traffic capped at 20 Mbps.

Signed-off-by: Frey Alfredsson <freysteinn@freysteinn.com>


bpf-filter/filter.c
```diff
@@ -0,0 +1,70 @@
...
+		goto out;
+	}
+
+	/* Classify TCP ports 8080 and 8081 */
+	tcp_type = parse_tcphdr(&nh, data_end, &tcphdr);
+	if (tcphdr + 1 > data_end) {
+		goto out;
+	}
+
+	switch (bpf_ntohs(tcphdr->dest)) {
```
tohojo, line 58: This incurs a runtime conversion - better to match on the raw header field and use bpf_htons() on the cases...

bpf-filter/run-bpf.sh
```diff
@@ -0,0 +1,76 @@
...
+$TC -netns "$L_NS" qdisc add dev "$L_DEV" root handle 1:0 htb default 30
+TC_CLASS_ADD="$TC -netns $L_NS class add dev $L_DEV parent"
+$TC_CLASS_ADD 1:0 classid 1:1 htb rate "$LIMIT"
+$TC_CLASS_ADD 1:1 classid 1:10 htb rate "$START_RATE" ceil "$P8080_LIMIT"
+$TC_CLASS_ADD 1:1 classid 1:20 htb rate "$START_RATE" ceil "$P8081_LIMIT"
+$TC_CLASS_ADD 1:1 classid 1:30 htb rate "$START_RATE" ceil "$DEFAULT_LIMIT"
+
+# Setup filters
+$TC -netns $L_NS filter add dev $L_DEV protocol ip parent 1:0 \
+		bpf obj filter.o classid 1: direct-action
```
tohojo, line 66: This is the only thing that is different between the two scripts - so do we really need two? Just passing 'bpf' or 'u32' as an argument to the script and varying the filter setup based on that would save the duplication...

tohojo (changes requested):
Found a few nits, but otherwise looks good :)

bpf-filter/run-bpf.sh
freysteinn: You are right. I will merge the scripts and add parameters.

bpf-filter/filter.c
freysteinn: That makes sense. I will change that and send in a new pull-request.

tohojo:
Frey Alfredsson ***@***.***> writes:

> @freysteinn commented on this pull request.
>
>
>
>> +	}
> +	else if (eth_type == bpf_htons(ETH_P_IPV6)) {
> +		ip_type = parse_ip6hdr(&nh, data_end, &ipv6hdr);
> +		if (ip_type != IPPROTO_TCP)
> +			goto out;
> +	} else {
> +		goto out;
> +	}
> +
> +	/* Classify TCP ports 8080 and 8081 */
> +	tcp_type = parse_tcphdr(&nh, data_end, &tcphdr);
> +	if (tcphdr + 1 > data_end) {
> +		goto out;
> +	}
> +
> +	switch (bpf_ntohs(tcphdr->dest)) {
>
> That makes sense. I will change that and send in a new pull-request.

If you just force-push to the branch this PR will update itself :)

freysteinn:
> Frey Alfredsson ***@***.***> writes:
> @freysteinn commented on this pull request. > + } + else if (eth_type == bpf_htons(ETH_P_IPV6)) { + ip_type = parse_ip6hdr(&nh, data_end, &ipv6hdr); + if (ip_type != IPPROTO_TCP) + goto out; + } else { + goto out; + } + + /* Classify TCP ports 8080 and 8081 */ + tcp_type = parse_tcphdr(&nh, data_end, &tcphdr); + if (tcphdr + 1 > data_end) { + goto out; + } + + switch (bpf_ntohs(tcphdr->dest)) { That makes sense. I will change that and send in a new pull-request.
> If you just force-push to the branch this PR will update itself :)

Excellent, I will do that :)

netoptimizer:
I find the directory name 'bpf-filter' misleading.

tohojo:
Jesper Dangaard Brouer ***@***.***> writes:

> I find the directory name 'bpf-filter' misleading.

Good point, didn't even notice that :)

Maybe 'tc-bpf-basic' or something like that instead?

freysteinn:
Yes, that's a good point.
I think tc-bpf-basic sounds good. However, if we later create an example that uses actions instead, would it be better to call it tc-bpf-basic-classifier? What do you think?

tohojo:
Frey Alfredsson ***@***.***> writes:

> Yes, that's a good point.
> I think tc-bpf-basic sounds good. However, if we later create an
> example that uses actions instead, would it be better to call it
> tc-bpf-basic-classifier? What do you think?

Sure, makes sense; or maybe drop 'bpf', and just go with
'tc-basic-classifier' ('bpf' is in the repo name, after all)

tc-basic-classifier/filter.c
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	/* Classify TCP ports 8080 and 8081 */
+	tcp_type = parse_tcphdr(&nh, data_end, &tcphdr);
+	if (tcphdr + 1 > data_end) {
+		goto out;
+	}
+
+	switch (tcphdr->dest) {
+	case bpf_ntohs(8080):
```
tohojo, line 59: This should be bpf_htons()

tc-basic-classifier/filter.c
```diff
@@ -0,0 +1,70 @@
...
+		skb->tc_classid = 0x10; /* Handles are always in hex */
+		break;
+	case bpf_ntohs(8081):
```
tohojo, line 62: And this as well

tc-basic-classifier/runner.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0
+# Copyright 2021 Frey Alfredsson <freysteinn@freysteinn.com>
+
+#set -o errexit
```
tohojo, line 5: ?

tohojo (changes requested):
A few more nits :)

tc-basic-classifier/runner.sh
freysteinn: Ahh, I was debugging an error related to the $1 variable not being set when I had no parameters. I forgot to remove the comment

tc-basic-classifier/filter.c
freysteinn: Of course. I blindly took the previous call and moved it. Unfortunately, it didn't fail, and therefore I didn't think much of it

tc-basic-classifier/runner.sh
tohojo: Thought that might be the case :)

tc-basic-classifier/filter.c
tohojo: Yup, figured - and those 'h' and 'n's do look way too much like each other ;)

tohojo (approved):
Right, LGTM :)

PR #16, opened 2021-04-30
Pping better output

simosund:
These commits update the standard output format to be equivalent with Kathie's pping standard format, which means it adds a timestamp (pushed from the XDP program) and min-RTT (adds in-kernel tracking of minimum RTT for each flow). Furthermore, I've added a simple JSON format as well as Kathie's "machine friendly" format (used by ppviz).

TODOs:
- Currently the program my report incorrect minimum-RTT if the tc program deletes the flow-information before the XDP program can calculate the RTT of a packet from that flow. Couple of options for fixes would be to only attempt to delete flow-information from the XDP program, or include the current min-RTT into the stored packet timestamp. 
- Update `README` and `TODO`

### DIFF
--- a/pping/Makefile
+++ b/pping/Makefile
@@ -3,7 +3,7 @@
 USER_TARGETS   := pping
 BPF_TARGETS    := pping_kern
 
-LDFLAGS    += -pthread
+LDLIBS     += -pthread
 EXTRA_DEPS += pping.h
 
 LIB_DIR = ../lib
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -74,7 +74,7 @@ struct pping_config {
 	char *pin_dir;
 	char *packet_map;
 	char *flow_map;
-	char *rtt_map;
+	char *event_map;
 	int xdp_flags;
 	int ifindex;
 	char ifname[IF_NAMESIZE];
@@ -377,8 +377,10 @@ static bool packet_ts_timeout(void *val_ptr, __u64 now)
 static bool flow_timeout(void *val_ptr, __u64 now)
 {
 	__u64 ts = ((struct flow_state *)val_ptr)->last_timestamp;
-	if (now > ts && now - ts > FLOW_LIFETIME)
+	if (now > ts && now - ts > FLOW_LIFETIME) {
+		//TODO - create and "push" flow-closing event
 		return true;
+	}
 	return false;
 }
 
@@ -517,33 +519,83 @@ static const char *proto_to_str(__u16 proto)
 	}
 }
 
-static void print_rtt_event_standard(void *ctx, int cpu, void *data,
-				     __u32 data_size)
+static const char *flowevent_to_str(enum flow_event_type fe)
 {
-	const struct rtt_event *e = data;
+	switch (fe) {
+	case FLOW_EVENT_UNSPECIFIED:
+		return "unspecified event";
+	case FLOW_EVENT_OPENING:
+		return "opening";
+	case FLOW_EVENT_CLOSING:
+		return "closing";
+	default:
+		return "unknown";
+	}
+}
+
+static const char *eventreason_to_str(enum flow_event_reason er)
+{
+	switch (er) {
+	case EVENT_REASON_SYN:
+		return "SYN";
+	case EVENT_REASON_SYN_ACK:
+		return "SYN-ACK";
+	case EVENT_REASON_FIRST_OBS_PCKT:
+		return "first observed packet";
+	case EVENT_REASON_FIN:
+		return "FIN";
+	case EVENT_REASON_FIN_ACK:
+		return "FIN-ACK";
+	case EVENT_REASON_RST:
+		return "RST";
+	case EVENT_REASON_FLOW_TIMEOUT:
+		return "flow timeout";
+	default:
+		return "unknown";
+	}
+}
+
+static void print_event_standard(void *ctx, int cpu, void *data,
+				 __u32 data_size)
+{
+	const struct flow_event *fe = data;
+	const struct rtt_event *re = data;
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
 	char timestr[9];
-	__u64 ts = convert_monotonic_to_realtime(e->timestamp);
+	__u64 ts = convert_monotonic_to_realtime(fe->timestamp);
 	time_t ts_s = ts / NS_PER_SECOND;
 
-	format_ip_address(saddr, sizeof(saddr), e->flow.ipv, &e->flow.saddr.ip);
-	format_ip_address(daddr, sizeof(daddr), e->flow.ipv, &e->flow.daddr.ip);
+	format_ip_address(saddr, sizeof(saddr), fe->flow.ipv,
+			  &fe->flow.saddr.ip);
+	format_ip_address(daddr, sizeof(daddr), fe->flow.ipv,
+			  &fe->flow.daddr.ip);
 	strftime(timestr, sizeof(timestr), "%H:%M:%S", localtime(&ts_s));
 
-	printf("%s.%09llu %llu.%06llu ms %llu.%06llu ms %s:%d+%s:%d\n", timestr,
-	       ts % NS_PER_SECOND, e->rtt / NS_PER_MS, e->rtt % NS_PER_MS,
-	       e->min_rtt / NS_PER_MS, e->min_rtt % NS_PER_MS, saddr,
-	       ntohs(e->flow.saddr.port), daddr, ntohs(e->flow.daddr.port));
+	if (fe->event_type == EVENT_TYPE_RTT)
+		printf("%s.%09llu %llu.%06llu ms %llu.%06llu ms %s:%d+%s:%d\n",
+		       timestr, ts % NS_PER_SECOND, re->rtt / NS_PER_MS,
+		       re->rtt % NS_PER_MS, re->min_rtt / NS_PER_MS,
+		       re->min_rtt % NS_PER_MS, saddr, ntohs(re->flow.saddr.port),
+		       daddr, ntohs(re->flow.daddr.port));
+	else if (fe->event_type == EVENT_TYPE_FLOW)
+		printf("%s.%09llu %s:%d+%s:%d flow %s due to %s from %s\n",
+		       timestr, ts & NS_PER_SECOND, saddr,
+		       ntohs(fe->flow.saddr.port), daddr,
+		       ntohs(fe->flow.daddr.port), flowevent_to_str(fe->event),
+		       eventreason_to_str(fe->reason),
+		       fe->from_egress ? "src" : "dest");
 }
 
-static void print_rtt_event_ppviz(void *ctx, int cpu, void *data,
-				  __u32 data_size)
+static void print_event_ppviz(void *ctx, int cpu, void *data, __u32 data_size)
 {
 	const struct rtt_event *e = data;
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
 	__u64 time = convert_monotonic_to_realtime(e->timestamp);
+
+	if (e->event_type != EVENT_TYPE_RTT)
+		return;
 
 	format_ip_address(saddr, sizeof(saddr), e->flow.ipv, &e->flow.saddr.ip);
 	format_ip_address(daddr, sizeof(daddr), e->flow.ipv, &e->flow.daddr.ip);
@@ -555,16 +607,22 @@ static void print_rtt_event_ppviz(void *ctx, int cpu, void *data,
 	       ntohs(e->flow.saddr.port), daddr, ntohs(e->flow.daddr.port));
 }
 
-static void print_rtt_event_json(void *ctx, int cpu, void *data,
-				 __u32 data_size)
+static void print_event_json(void *ctx, int cpu, void *data, __u32 data_size)
 {
-	const struct rtt_event *e = data;
+	const struct flow_event *fe = data;
+	const struct rtt_event *re = data;
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
-	__u64 time = convert_monotonic_to_realtime(e->timestamp);
+	__u64 time = convert_monotonic_to_realtime(fe->timestamp);
 
-	format_ip_address(saddr, sizeof(saddr), e->flow.ipv, &e->flow.saddr.ip);
-	format_ip_address(daddr, sizeof(daddr), e->flow.ipv, &e->flow.daddr.ip);
+	if (fe->event_type != EVENT_TYPE_RTT &&
+	    fe->event_type != EVENT_TYPE_FLOW)
+		return;
+
+	format_ip_address(saddr, sizeof(saddr), fe->flow.ipv,
+			  &fe->flow.saddr.ip);
+	format_ip_address(daddr, sizeof(daddr), fe->flow.ipv,
+			  &fe->flow.daddr.ip);
 
 	if (!json_ctx) {
 		json_ctx = jsonw_new(stdout);
@@ -573,17 +631,29 @@ static void print_rtt_event_json(void *ctx, int cpu, void *data,
 
 	jsonw_start_object(json_ctx);
 	jsonw_u64_field(json_ctx, "timestamp", time);
-	jsonw_u64_field(json_ctx, "rtt", e->rtt);
-	jsonw_u64_field(json_ctx, "min_rtt", e->min_rtt);
 	jsonw_string_field(json_ctx, "src_ip", saddr);
-	jsonw_hu_field(json_ctx, "src_port", ntohs(e->flow.saddr.port));
+	jsonw_hu_field(json_ctx, "src_port", ntohs(fe->flow.saddr.port));
 	jsonw_string_field(json_ctx, "dest_ip", daddr);
-	jsonw_hu_field(json_ctx, "dest_port", ntohs(e->flow.daddr.port));
-	jsonw_string_field(json_ctx, "protocol", proto_to_str(e->flow.proto));
-	jsonw_u64_field(json_ctx, "sent_packets", e->sent_pkts);
-	jsonw_u64_field(json_ctx, "sent_bytes", e->sent_bytes);
-	jsonw_u64_field(json_ctx, "rec_packets", e->rec_pkts);
-	jsonw_u64_field(json_ctx, "rec_bytes", e->rec_bytes);
+	jsonw_hu_field(json_ctx, "dest_port", ntohs(fe->flow.daddr.port));
+	jsonw_string_field(json_ctx, "protocol", proto_to_str(fe->flow.proto));
+
+	if (fe->event_type == EVENT_TYPE_RTT) {
+		jsonw_u64_field(json_ctx, "rtt", re->rtt);
+		jsonw_u64_field(json_ctx, "min_rtt", re->min_rtt);
+		jsonw_u64_field(json_ctx, "sent_packets", re->sent_pkts);
+		jsonw_u64_field(json_ctx, "sent_bytes", re->sent_bytes);
+		jsonw_u64_field(json_ctx, "rec_packets", re->rec_pkts);
+		jsonw_u64_field(json_ctx, "rec_bytes", re->rec_bytes);
+
+	} else if (fe->event_type == EVENT_TYPE_FLOW) {
+		jsonw_string_field(json_ctx, "flow_event",
+				   flowevent_to_str(fe->event));
+		jsonw_string_field(json_ctx, "reason",
+				   eventreason_to_str(fe->reason));
+		jsonw_string_field(json_ctx, "from",
+				   fe->from_egress ? "src" : "dest");
+	}
+
 	jsonw_end_object(json_ctx);
 }
 
@@ -710,13 +780,13 @@ int main(int argc, char *argv[])
 		.pin_dir = "/sys/fs/bpf/pping",
 		.packet_map = "packet_ts",
 		.flow_map = "flow_state",
-		.rtt_map = "rtt_events",
+		.event_map = "events",
 		.xdp_flags = XDP_FLAGS_UPDATE_IF_NOEXIST,
 	};
 
 	struct perf_buffer *pb = NULL;
 	struct perf_buffer_opts pb_opts = {
-		.sample_cb = print_rtt_event_standard,
+		.sample_cb = print_event_standard,
 		.lost_cb = handle_missed_rtt_event,
 	};
 
@@ -743,9 +813,9 @@ int main(int argc, char *argv[])
 	}
 
 	if (config.json_format)
-		pb_opts.sample_cb = print_rtt_event_json;
+		pb_opts.sample_cb = print_event_json;
 	else if (config.ppviz_format)
-		pb_opts.sample_cb = print_rtt_event_ppviz;
+		pb_opts.sample_cb = print_event_ppviz;
 
 	err = load_attach_bpfprogs(&obj, &config, &tc_attached, &xdp_attached);
 	if (err) {
@@ -764,13 +834,13 @@ int main(int argc, char *argv[])
 
 	// Set up perf buffer
 	pb = perf_buffer__new(bpf_object__find_map_fd_by_name(obj,
-							      config.rtt_map),
+							      config.event_map),
 			      PERF_BUFFER_PAGES, &pb_opts);
 	err = libbpf_get_error(pb);
 	if (err) {
 		pb = NULL;
 		fprintf(stderr, "Failed to open perf buffer %s: %s\n",
-			config.rtt_map, strerror(err));
+			config.event_map, strerror(err));
 		goto cleanup;
 	}
 
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -488,8 +488,8 @@ static __u64 convert_monotonic_to_realtime(__u64 monotonic_time)
  * Wrapper around inet_ntop designed to handle the "bug" that mapped IPv4
  * addresses are formated as IPv6 addresses for AF_INET6
  */
-static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
-			     size_t size)
+static int format_ip_address(char *buf, size_t size, int af,
+			     const struct in6_addr *addr)
 {
 	if (af == AF_INET)
 		return inet_ntop(af, &addr->s6_addr[12],
@@ -526,8 +526,8 @@ static void print_rtt_event_standard(void *ctx, int cpu, void *data,
 	__u64 ts = convert_monotonic_to_realtime(e->timestamp);
 	time_t ts_s = ts / NS_PER_SECOND;
 
-	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
-	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
+	format_ip_address(saddr, sizeof(saddr), e->flow.ipv, &e->flow.saddr.ip);
+	format_ip_address(daddr, sizeof(daddr), e->flow.ipv, &e->flow.daddr.ip);
 	strftime(timestr, sizeof(timestr), "%H:%M:%S", localtime(&ts_s));
 
 	printf("%s.%09llu %llu.%06llu ms %llu.%06llu ms %s:%d+%s:%d\n", timestr,
@@ -544,8 +544,8 @@ static void print_rtt_event_ppviz(void *ctx, int cpu, void *data,
 	char daddr[INET6_ADDRSTRLEN];
 	__u64 time = convert_monotonic_to_realtime(e->timestamp);
 
-	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
-	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
+	format_ip_address(saddr, sizeof(saddr), e->flow.ipv, &e->flow.saddr.ip);
+	format_ip_address(daddr, sizeof(daddr), e->flow.ipv, &e->flow.daddr.ip);
 
 	printf("%llu.%09llu %llu.%09llu %llu.%09llu %s:%d+%s:%d\n",
 	       time / NS_PER_SECOND, time % NS_PER_SECOND,
@@ -562,8 +562,8 @@ static void print_rtt_event_json(void *ctx, int cpu, void *data,
 	char daddr[INET6_ADDRSTRLEN];
 	__u64 time = convert_monotonic_to_realtime(e->timestamp);
 
-	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
-	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
+	format_ip_address(saddr, sizeof(saddr), e->flow.ipv, &e->flow.saddr.ip);
+	format_ip_address(daddr, sizeof(daddr), e->flow.ipv, &e->flow.daddr.ip);
 
 	if (json_started) {
 		printf(",");
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -478,23 +478,22 @@ static void *periodic_map_cleanup(void *args)
 
 static __u64 convert_monotonic_to_realtime(__u64 monotonic_time)
 {
-	__u64 now_mon, now_rt;
 	static __u64 offset = 0;
 	static __u64 offset_updated = 0;
+	__u64 now_mon = get_time_ns(CLOCK_MONOTONIC);
+	__u64 now_rt;
 
-	now_mon = get_time_ns(CLOCK_MONOTONIC);
 	if (offset == 0 ||
 	    (now_mon > offset_updated &&
 	     now_mon - offset_updated > MON_TO_REAL_UPDATE_FREQ)) {
 		now_mon = get_time_ns(CLOCK_MONOTONIC);
 		now_rt = get_time_ns(CLOCK_REALTIME);
+
 		if (now_rt < now_mon)
 			return 0;
-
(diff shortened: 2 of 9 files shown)


pping/pping.c
```diff
@@ -40,6 +40,11 @@ static const char *__doc__ =
 
 #define MAX_PATH_LEN 1024
 
+#define MON_TO_REAL_UPDATE_FREQ                                                \
+	(1 * NS_PER_SECOND) // Update offset between CLOCK_MONOTONIC and CLOCK_REALTIME once per second
+#define TIME_STR_LEN                                                           \
+	20 // Length of string containing timestamp "HH:MM:SS"
```
tohojo, line 46: No it isn't? ;)

("HH:MM:SS" is only 9 bytes long include null-byte)... Also, you can use sizeof() on a string constant, so you don't really need the define...

pping/pping.c
```diff
@@ -461,12 +488,17 @@ static void handle_rtt_event(void *ctx, int cpu, void *data, __u32 data_size)
 	const struct rtt_event *e = data;
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
+	char timestr[TIME_STR_LEN];
+	time_t time =
+		convert_monotonic_to_realtime(e->timestamp) / NS_PER_SECOND;
 
 	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
 	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
+	strftime(timestr, TIME_STR_LEN, "%H:%M:%S", localtime(&time));
```
tohojo, line 510: Full-second granularity is too long here (you'll get several data points for each second). Let's at least go down to ms accuracy (so "HH:MM:SS.ms")...

pping/pping_kern.c
```diff
@@ -332,30 +333,40 @@ int pping_ingress(struct xdp_md *ctx)
...
+	event.rtt = now - *p_ts;
+	event.timestamp = now;
 	/*
 	 * Attempt to delete timestamp entry as soon as RTT is calculated.
 	 * But could have potential concurrency issue where multiple packets
 	 * manage to match against the identifier before it can be deleted.
 	 */
 	bpf_map_delete_elem(&packet_ts, &p_id);
 
+	// Update flow's min-RTT, may have concurrency issues
```
tohojo, line 355: May have? :)

pping/pping_kern.c
```diff
@@ -332,30 +333,40 @@ int pping_ingress(struct xdp_md *ctx)
...
+	f_state = bpf_map_lookup_elem(&flow_state, &p_id.flow);
+	if (f_state && (f_state->min_rtt == 0 || event.rtt < f_state->min_rtt))
+		f_state->min_rtt = event.rtt;
+
+	event.min_rtt = f_state ? f_state->min_rtt : event.rtt;
```
tohojo, line 360: Under what circumstances can this fail while the timestamp lookup above succeeds? And if it does, is outputing an event really the right thing to do?

pping/pping.c
```diff
@@ -69,27 +69,30 @@ struct map_cleanup_args {
...
 static volatile int keep_running = 1;
+static bool json_started = false;
 
 static const struct option long_options[] = {
 	{ "help",             no_argument,       NULL, 'h' },
 	{ "interface",        required_argument, NULL, 'i' }, // Name of interface to run on
 	{ "rate-limit",       required_argument, NULL, 'r' }, // Sampling rate-limit in ms
 	{ "force",            no_argument,       NULL, 'f' }, // Detach any existing XDP program on interface
 	{ "cleanup-interval", required_argument, NULL, 'c' }, // Map cleaning interval in s
+	{ "json",             no_argument,       NULL, 'j' }, // Output in JSON format
```
tohojo, line 96: How about making this a `--format` (`-F`?) argument instead? So you can do use `--format json` - makes it easier to add more formats without blowing through a lot of flags (that the user also has to remember).

pping/pping.c
```diff
@@ -502,6 +510,39 @@ static void handle_rtt_event(void *ctx, int cpu, void *data, __u32 data_size)
...
+	} else {
+		printf("[");
+		json_started = true;
+	}
+
+	printf("\n{\"timestamp\":%llu.%09llu, \"rtt\":%llu.%09llu, \"min_rtt\":%llu.%09llu, \"src_ip\":\"%s\", \"src_port\":%d, \"dest_ip\":\"%s\", \"dest_port\":%d}",
+	       time / NS_PER_SECOND, time % NS_PER_SECOND,
+	       e->rtt / NS_PER_SECOND, e->rtt % NS_PER_SECOND,
+	       e->min_rtt / NS_PER_SECOND, e->min_rtt % NS_PER_SECOND, saddr,
+	       ntohs(e->flow.saddr.port), daddr, ntohs(e->flow.daddr.port));
```
tohojo, line 557: I am mildly horrified at this way of writing JSON ;)
But I don't *really* see any good reason why it won't do, at least for now...

pping/pping.c
```diff
@@ -93,6 +94,7 @@ static const struct option long_options[] = {
 	{ "force",            no_argument,       NULL, 'f' }, // Detach any existing XDP program on interface
 	{ "cleanup-interval", required_argument, NULL, 'c' }, // Map cleaning interval in s
 	{ "json",             no_argument,       NULL, 'j' }, // Output in JSON format
+	{ "machine-friendly", no_argument,       NULL, 'm' }, // Ouput in Kathie's "machine friendly" format
```
tohojo, line 97: And then this could just be adding another possible argument to `--format`. Maybe even call it `ppviz` instead of "machine friendly"?

tohojo (changes requested):
A few comments...

pping/pping.c
simosund: Yhea, originally used the format string "YYYY-MM-DD HH:MM:SS" (which is 20 characters including the null-byte) but thought it became a bit long. So did the somewhat unusual misstake of updating the documentation but not the code =). As you say though, probably much better to just have the format string and use `sizeof` on that to avoid these types of problems alltogether.

pping/pping.c
simosund: Agree, but I suspect that just as Kathie I got a bit lazy when strftime didn't support anything lower by itself.

pping/pping_kern.c
simosund: Well, can in theory of course end up with the situation where multiple packets for the flow is discovered to have the lowest RTT concurrently, and basically get a lost update. So similar problems as most of the other code that operates on these non-per-CPU maps, but got a bit tired of commenting about it as my tc program is mainly comments about possible concurrency issues at this point.

From my understanding from some previous discussions however these types of concurrency issues aren't too likely in reality as packets from the same flow are generally processed on the same core.

pping/pping_kern.c
simosund: The timestamp lookup of course has to succeed, but the flow-state lookup (which keeps the per-flow minimum RTT) doesn't have to. That flow information can potentially be deleted by the tc program if it detects that the flow is about to close before the XDP program has had time to get a response and compute an RTT. 

So this is basically what I refered to as the issue where the wrong minimum-RTT may be reported (mainly a possibility for the final RTT calculation for the flow). Have some changes in the works for this, which includes a bit less aggressive deletion of flow information and not outputting the RTT in case no relevant flow information can be found.

pping/pping.c
simosund: Seems like a good idea. Don't currently plan on adding a whole lot of additional formats, but can never know what the future has in store.

pping/pping.c
simosund: Can understand that, but for the relativly simple JSON format I use I thought it was sufficent. If we start using a more nested JSON structure or similar I guess it makes sense to move towards some kinda of JSON-builder, although not sure how you feel about additional depedencies for the bpf-examples.

pping/pping.c
simosund: Sure. Simply kept the name that Kathie used for it, but it's primary use case is pretty much `ppviz`, so that's probably more descriptive/telling.

pping/pping_kern.c
tohojo: Yup, totally agree that leaving comments about concurrency issues sprinkled throughout code is just tedious, especially if they're vague ones like this one. Which was kinda the point of my comment ;)

Ideally this is something that will be handled at the design level. Which would in this case mean documenting the assumptions in one place, and they either referring to that in the code, or just leaving out comments like this one entirely :)

pping/pping_kern.c
tohojo: Right, but my point was more that if the flow state is not present (because it has been garbage-collected), is it really appropriate to output a data point? The even says it contains the "min rtt" of the flow, so if we don't know what that is, shouldn't we either say so (min_rtt = 0), or abort entirely (just skip the event)?

simosund: Yhea, my plan for now is to skip the event entierly in those instances.

tohojo: OK, SGTM :)

pping/pping.c
abrunstrom: Thinking also about more local use case, like latency between VMs, even ms granularity may perhaps be too course grained.

pping/pping_kern.c
abrunstrom: Losing a minRTT update does not seem very critical. When sampling, the minRTT is anyway an estimate and if you had sampled a different packet you would anyway have missed it.

pping/pping.c
simosund: > Thinking also about more local use case, like latency between VMs, even ms granularity may perhaps be too course grained.

Well, we're talking about the timestamp here and not the RTT (which is printed in ms with 6 decimals, so nanosecond resolution), so not sure if I see why it matters if it's local or not (in some sense I guess the timestamp when we recieve the packet the RTT is calculated for is always local). For the JSON and "machine friendly"/ppviz format the timestamp is printed with nanosecond resolution, but for the "standard" console output I'm not sure that's of much use to the user. If events are printed out so fast they can't even be differentiated with ms-resolution the "standard" format is probably not what you should be using to begin with.

abrunstrom: My thought with local was that then we may have very short RTTs with a possibility and need for a much higher sampling rate, so having a fine granular timestamp may be needed to separate the samples. But if we have it in the timestamp for the JSON and ppviz format and that is what we will use for the actual measurements and this format is only for live visual inspection I agree it is not useful. But if you want to make a short test run, pipe the output to a file and then make some simple post processing or visually inspect the result it may still be good to make sure all the timestamps are unique?

simosund: Hmm, guess that's sort of a valid point. Have added the full nano-second resolution of the timestamp in the default output mode now as well. Initially wanted to try and prevent the output from getting so wide, but as I don't do the adaptive us/ns unit selection and rounding of RTTs that Kathie does my output is already a fair bit wider anyways. 

Even with no sampling at all however you're still not very likely to get output at a higher rate than 1000 Hz within a single flow due to TCP timestamps only being updated at 1000Hz.

abrunstrom: That is true, i did not think about the TCP timestamp update rate. Maybe with another protocol... Pick what you feel is best.

tohojo:
abrunstrom ***@***.***> writes:

> @abrunstrom commented on this pull request.
>
>
>
>>  
>  	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
>  	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
> +	strftime(timestr, TIME_STR_LEN, "%H:%M:%S", localtime(&time));
>
> That is true, i did not think about the TCP timestamp update rate.
> Maybe with another protocol... Pick what you feel is best.

Just because a flow doesn't *emit* new timestamps closer together than
1ms, doesn't guarantee that they don't *arrive* within the same ms,
though. Say if two consecutive timestamps are caught up in the same
queue and then released together. Middlebox shenanigans can also play
into this (ACK thinning, for one).

simosund:
> Just because a flow doesn't *emit* new timestamps closer together than
1ms, doesn't guarantee that they don't *arrive* within the same ms,
though. Say if two consecutive timestamps are caught up in the same
queue and then released together. Middlebox shenanigans can also play
into this (ACK thinning, for one).

Good point. Anyways, as of commit [f6cf8ed](https://github.com/xdp-project/bpf-examples/pull/16/commits/f6cf8edd96243f1f5a6d44e6ce66eaae2ac73a3a) I now output timestamps in all formats (including the standard one) with nanosecond resolution. The accuracy is probably a bit lower though, partly because I don't actually take the timestamp before I've parsed the packet, and partly because having to convert between `CLOCK_MONOTONIC` and `CLOCK_REALTIME`)

tohojo:

> Hi @tohojo, just realized that in its current form you can only run pping on a single interface, as if you try to run it on another it will fail pinning the tc program (as it will try to pin it to the same place as the tc program for the other interface). 
>
> Thinking we could just add some random suffix to the pinned location (so instead of pinning to `/sys/fs/bpf/pping/classifier` it pins it to `sys/fs/bpf/pping/classifier_<random_suffix>`) that makes it rare enough that they "should" never crash. Another option I guess would be to add the interface name as a suffix, but that might not work if you want to attach them inside virtual environments (could for example then have multiple veth0 interfaces). 
>
> Or do you have any other ideas on how to best solve it?

Why is the pinning staying around? You can unlink the pinned file as
soon as you've loaded the program with TC, no? in which case the race
window is quite small, so you could just retry if you hit it on the
second run...

simosund:
> Why is the pinning staying around? You can unlink the pinned file as
soon as you've loaded the program with TC, no?

Hmm, for some reason I just assumed that it wouldn't work. But of course that works and is overall a much nicer solution than having a bunch of pinned programs staying around with different names.

pping/pping.c
```diff
@@ -456,18 +499,98 @@ static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
...
+	if (json_started) {
+		printf(",");
+	} else {
+		printf("[");
+		json_started = true;
+	}
+
+	printf("\n{\"timestamp\":%llu.%09llu, \"rtt\":%llu.%09llu, "
+	       "\"min_rtt\":%llu.%09llu, \"src_ip\":\"%s\", \"src_port\":%d, "
+	       "\"dest_ip\":\"%s\", \"dest_port\":%d, \"protocol\":\"%s\", "
```
tohojo, line 577: Including the protocol is a good idea. It may be worth thinking about what the format is supposed to be if it's not TCP in more detail, though. For instance, ICMP has no port numbers; so what will src_port and dest_port contain if it's an ICMP flow? Will they be omitted? or null? or something else?

pping/pping.c
```diff
@@ -456,18 +499,98 @@ static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
...
+{
+	static char buf[8];
+
+	switch (proto) {
+	case IPPROTO_TCP:
+		return "TCP";
+	case IPPROTO_ICMP:
+		return "ICMP";
+	case IPPROTO_ICMPV6:
+		return "ICMPv6";
```
tohojo, line 512: No love for UDP? ;)

pping/pping.c
```diff
@@ -456,18 +499,98 @@ static int format_ip_address(int af, const struct in6_addr *addr, char *buf,
...
+static void print_rtt_event_json(void *ctx, int cpu, void *data,
+				 __u32 data_size)
 {
 	const struct rtt_event *e = data;
 	char saddr[INET6_ADDRSTRLEN];
 	char daddr[INET6_ADDRSTRLEN];
+	__u64 time = convert_monotonic_to_realtime(e->timestamp);
 
 	format_ip_address(e->flow.ipv, &e->flow.saddr.ip, saddr, sizeof(saddr));
 	format_ip_address(e->flow.ipv, &e->flow.daddr.ip, daddr, sizeof(daddr));
```
tohojo, line 566: It just occurred to me that most other "print" type functions (e.g., sprintf) has the target buffer as the first argument; so this looks a little odd...

tohojo (changes requested):
Looks mostly good now, but may be worth it to think a little more about extensibility of the JSON format ahead of time... :)

pping/pping.c
simosund: Well, guess I could just change order to: (buffer, buffer-size, ip-version, ip-address) then to follow convention a bit better.

pping/pping.c
```diff
@@ -646,6 +647,13 @@ static int load_attach_bpfprogs(struct bpf_object **obj,
 	}
 	*tc_attached = true;
 
+	err = bpf_obj_unpin_program(*obj, config->egress_sec, config->pin_dir);
+	if (err) {
+		fprintf(stderr, "Failed unpinning tc program from %s: %s\n",
+			config->pin_dir, strerror(-err));
+		return err;
```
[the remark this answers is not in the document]
tohojo: Ah, right, there's a separate bit of cleanup in the main function. It would be better to put that into this function; the way it is now, you have to keep the state of the entire program in your head to know that things are being cleaned up correctly; making each function clean up properly after itself makes it way easier to reason about the behaviour...

simosund:
Hi @tohojo, I've now tried to refactor the code to accommodate some of your requests, and also moved to using the json-writer from iproute. For now I've simply copied the files directly to the `pping` folder, but guessing they're maybe supposed to go into a `lib/util` folder instead?

Have limited experience with larger projects in general, and the Makefiles for this project are a bit more complicated than the trivial ones I've had to work with during my education. So its not clear to me exactly how additional dependencies are supposed to be added here.

pping/pping.c
```diff
@@ -531,11 +652,39 @@ static int load_attach_bpfprogs(struct bpf_object **obj,
...
 			config->force ? "" : ", ensure no other XDP program is already running on interface",
 			strerror(-err));
-		return err;
+		goto cleanup;
 	}
-	*xdp_attached = true;
+	xdp_attached = true;
 
-	return 0;
+cleanup:
```
tohojo, line 659: Instead of having one label and multiple conditionals in the cleanup path, try using different labels for each error. This makes the code more readable and also forces you to do the cleanup in the opposite order as when you were setting things up, which is good practice. Have a look at the error path in __dev_map_alloc_node() as an example:

https://elixir.bootlin.com/linux/latest/source/kernel/bpf/devmap.c#L583

pping/Makefile
```diff
@@ -3,9 +3,14 @@
...
 BPF_TARGETS    := pping_kern
 
-LDFLAGS    += -pthread
+LDLIBS     += -pthread
 EXTRA_DEPS += pping.h
 
+# Based on common.mk I would have guessed this should be json_writer.o,
+# but that causes infinite recursion so sticking to .c for now even if it
+# results in a "doesn't match target pattern" warning
+LIB_OBJS += json_writer.c
```
tohojo, line 12: This is not meant to be set at all from the program Makefiles. I pushed some updates to common.mk that adds the lib/util directory, so you should be able to rebase and then just drop the json_writer files into there (and update util.mk of course).

tohojo (changes requested):
Better, but a few more comments.

I also pushed an update to the libbpf submodule so you should be able to switch to using the libbpf TC API methods instead of shelling out now :)

pping/pping.c
simosund: Hmm, seems reasonable. Got into my head that `goto` is the root of all evil that will instantly lead to spaghetti code and therefore tend to avoid it. But avoiding it definitly leads to some rather clumsy conditionals at times.

pping/Makefile
simosund: Thanks, was guessing it was supposed to go into `lib/util`, but as there was no previous example to go by there I felt a bit unsure and figured it would be easier to just ask you =).

pping/pping.c
tohojo: Oh you mean like this? https://xkcd.com/292/

When used wrong, gotos can indeed lead to spaghetti code, but used correctly they can improve readability. In particular the "labels for error handling" pattern is extremely common in kernel code and it does help I think.

For another example that's a bit more questionable (but still IMO arguably acceptable due to its very specialised use case), feast your eyes upon this efficient memcpy implementation from the Cilium code base: https://github.com/cilium/cilium/blob/master/bpf/include/bpf/builtins.h#L163

pping/Makefile
tohojo: Yup, good thinking :D

pping/pping.c
simosund: > Oh you mean like this? https://xkcd.com/292/

Haha, yhea pretty much. Will constantly be looking over my shoulder to check for velociraptors now =).

tohojo: Don't worry, they are extinct in most of the world - just stay away from secluded tropical islands owned by crazy billionaires and you should be fine ;)

simosund: Hmm, not sure if my brain is just a bit slow today, but have to admit I'm struggeling somewhat to not get eaten by velociraptors right now when trying to apply this "goto err"-pattern. My problem is that the code for unpinning the tc program should run both on success and errors (as long as the tc program has first been pinned). Furthermore, unpinning the tc-program may also fail (and should in that case also detach the tc and xdp programs if they've previously been attached). 

So when I try to follow this structure I land in something like this (pseudo-code):

```
int load_attach_progs()
{
       // Some code of loading the obj-file

        err = pin_tc_prog();
        if (err)
                return err;
        
        err = attach_tc_prog();
        if (err)
                goto err_tc;
        
        err = attach_xdp_prog()
               goto err_xdp;

        err = unpin_tc_prog()
        if (err)
               goto unpin_err;
        return 0;

unpin_err:
        detach_xdp_prog()
xdp_err:
        detach_tc_prog();
tc_err:
        unpin_tc_prog();  // Duplicate code, may run unpin_tc_prog() twice
        return err;
}
```

This is not particularly pretty as I then have to duplicate the code for unpinning the tc program (once in the normal path, and once in the error-handling path), and if the initial unpinning of the tc-program fails (unpin_err), then it will attempt to unpin it again (which I would expect would just fail again, and print an additional error message about the same issue).

I could of course use an additional jump in the normal path to avoid duplicating the unpin_tc_prog() code, something similar to this:
```
int load_attach_progs()
{
       // Some code of loading the obj-file

        err = pin_tc_prog();
        if (err)
                return err;
        
        err = attach_tc_prog();
        if (err)
                goto err_tc;
        
        err = attach_xdp_prog()
               goto err_xdp;

        goto tc_err; // Not an acutal error, just want to unpin the tc prog

unpin_err:
        detach_xdp_prog()
xdp_err:
        detach_tc_prog();
tc_err:
        err = unpin_tc_prog();
        if (err)
                goto unpin_err; // Causes infinite loop if unpin_tc_prog() fails, will incorrectly try and detach xdp and tc programs if tc_attach_prog() failed.
        return err;
}
```
But this is obviously very broken. Could fix that by making the jump to unpin_err conditional, but at that point I'm back to having conditions in my cleanup code again as well as having jumps, making the whole thing a more confusing version of the original code IMO. Ex.

```
int load_attach_progs()
{
       // Some code of loading the obj-file

        bool do_back_jmp = false;

        err = pin_tc_prog();
        if (err)
                return err;
        
        err = attach_tc_prog();
        if (err)
                goto err_tc;
        
        err = attach_xdp_prog()
               goto err_xdp;

        do_back_jmp = true;
        goto tc_err; // Not an acutal error, just want to unpin the tc prog

unpin_err:
        detach_xdp_prog()
        do_back_jmp = false;
xdp_err:
        detach_tc_prog();
tc_err:
        err = unpin_tc_prog();
        if (err && do_back_jmp)
                goto unpin_err;
        return err;
}
```

tohojo: So when error handling ends up being too gnarly that is usually a good sign that the code needs refactoring :)

In this case, logically pinning is part of TC attach. Also, if unpinning fails after attaching the TC prog there is no way to avoid ending up in an inconsistent state (since no matter what you do you'll end up with the program still being pinned). So arguably it's enough to just warn and continue - so how about:

```C
int attach_tc_prog()
{
        err = pin_tc_prog();
        if (err)
                return err;

        err = __attach_tc_prog(); // do the actual attach

        /* we need to unpin regardless of whether attach succeeded, and
         * we can't really do anything about it if unpinning fails, so just warn
         * and continue if it does fail.
         */
        unpin_err = unpin_tc_prog();
        if (unpin_err)
             warning("couldn't unpin TC prog: %s", strerror(unpin_err));
        return err;
}

int load_attach_progs()
{
       // Some code of loading the obj-file

        err = attach_tc_prog();
        if (err)
                return err;
        
        err = attach_xdp_prog();
        if (err)
               goto err_xdp;
        return 0;

err_xdp:
        detach_tc_prog();
        return err;
}
```

This has the added benefit of making it really straight-forward to replace the code with the libbpf TC API later :)

simosund: Should have been able to figure that out myself... Anyways, thanks for the help!

simosund:
Hi again @tohojo. Hopefully the cleanup code looks a bit better now.

While I don't think you need to do another review yet (although if you want to I of course welcome the feedback), I am wondering what you think of my overall approach to implementing start/end of flow messages.

I didn't want to add a bunch of additional fields to the existing `rtt_event` struct (which I pass to userspace through the perf-buffer when an RTT is computed) that wouldn't be used for the RTT-messages. Likewise many of the existing fields in the RTT event wouldn't be used for the start/end of flow messages.

Therefore, I've instead added a new type of `flow_event` struct. To be able to send both of these through the same perf-buffer I added a `event_type` member to the beginning of them (which is only 8 bytes for the sake of struct alignment, for now a single bit would be sufficent otherwise) that can be used to differentiate between the two when receiving them. I've also reorderd the fields a bit so that all the fields that both event types have in common come first (and in the same order) to make the use of the common part a bit easier to use.

But now I've started to wonder if perhaps it would be better to just utilize a single event struct, and then use a union to allow for different interpretation of the (non-common) data fields. The drawback of this is that such a struct would be unecessarily large  for the "flow events" (which at least so far are a fair bit smaller than the RTT events due to not having so many 8-byte members). But thinking it would be somewhat easier to maintain (don't need to keep two different structs in sync), and for the XDP program it would be possible to reuse its existing rtt_event for another flow_event to save a little bit of stack-space and computation.

As for my current implementation, I'm also not a huge fan of how I on the BPF side directly access and fill in different values of the `flow_event` struct from different functions in a way where they all sort of have to be aware of eachtoher for it to work properly. Guess I could maybe package the `event` and `reason` memebers into their own struct to get a bit better logical seperation for `parse_packet_identifier()` at least.

But to summarize, what do you think of the overall approach? Is it reasonable, or should I switch to the "single struct with union" approach I mentioned, or something different entirely?

tohojo:
How about multiple structs with a union? :)

I.e., keep the two structs and define a union like:

```C
struct flow_event;
struct rtt_event;

union pping_events {
  __u64 event_type;
  struct flow_event;
  struct rtt_event;
};
```

Then code reading events can do something like:

```C
union pping_events *evt = get_event();
if (evt->event_type == EVENT_TYPE_FLOW) 
  process_flow_event(&evt->flow_event);
else if (evt->event_type == EVENT_TYPE_RTT)
  process_rtt_event(&evt->rtt_event);
```

Another thing to consider when designing a protocol like this (because at this point, you've ventured into protocol design; congratulations! ;)), is whether it needs to be extensible. I.e., with a design like this, the reader has to know the entire definition to be able to do anything sensible with the data; and if any of the structs ever get extended, the reader won't know which version is being used. For a perf buffer event, the reader can get the event size from there and infer the version of the struct; but if, say, you ever decide to stick two events into the same buffer, the length also needs to be encoded along with the type (AKA TLV encoding).

This is not a problem if BPF code and userspace code are always upgraded together; but it's worth thinking about anyway, because that might change (say if there's a long-running program loaded into the kernel where various userspace implementations need to read the data out).

simosund:
> How about multiple structs with a union? :)

Hmm, I guess that's slightly prettier than my current solution. As both the flow-events and rtt-events have a timestamp + flow-tuple, I'll probably have some common part for printing those out which will pretend that it is ex. a rtt-event regardless of if it is an rtt-event or flow-event.

> Another thing to consider when designing a protocol like this (because at this point, you've ventured into protocol design; congratulations! ;)), is whether it needs to be extensible. 

Why can nothing ever be simple? =) Guess we can discuss this a bit during the meeting tomorrow.

simosund:
Hi @tohojo, 
I've now tried switching to using libbpf to attach the tc program. I'm not entirely sure I've understood the API correctly though, so would be great if you could have a look at that (commit [9a4fa42](https://github.com/xdp-project/bpf-examples/pull/16/commits/9a4fa4288713258f206ebb02639bd5db0fb2b64a)). 

The API-description in the [commit](https://github.com/libbpf/libbpf/commit/d71ff87a2dd7b92787719aab233767e9c74fbd48) that added the TC-BPF API mentioned that you could check if a qdisc had any filters attached before deleting it, but couldn't figure out how to do that correctly. So right now the cleanup isn't 100% working as it still leaves the clsact qdisc. As I mentioned in the commit message I guess I could just use the previous solution of simply deleting the qdisc regardless if it's used by anything else or not, but figured that's probably a bad idea in case it actually is used by something else.

tohojo:
Hmm, the tc code went through quite a few revisions, so not sure what happened to the wildcard query; maybe it got dropped somewhere on the way? @kkdwivedi should know...

kkdwivedi:
Yeah, multi query/wildcard query was dropped, but looks I forgot to remove its description from the commit message (ouch).

So that example is pretty much broken and doesn't work. Right now you should just leave the clsact qdisc as is, since even what that example suggests (probing for zero filters) is inherently racy, as some else could install a filter between check and deletion.

pping/pping.c
```diff
@@ -292,72 +268,119 @@ static int init_rodata(struct bpf_object *obj, void *src, size_t size)
...
-	char *const argv[] = { TCBPF_LOADER_SCRIPT, "--dev",   interface,
-			       "--pinned",	    prog_path, NULL };
+	int err;
+	int prog_fd;
+	DECLARE_LIBBPF_OPTS(bpf_tc_hook, hook, .ifindex = ifindex,
+			    .attach_point = attach_point);
 
-	if (snprintf(prog_path, sizeof(prog_path), "%s/%s", pin_dir, section) < 0)
-		return -EINVAL;
+	err = bpf_tc_hook_create(&hook);
```
kkdwivedi, line 311: You would want to handle EEXIST here. This can happen when there is clsact, or the ingress qdisc installed on the interface. Also maybe avoid hook_destroy (or atleast only do it when you created the qdisc).

simosund:
> Yeah, multi query/wildcard query was dropped, but looks I forgot to remove its description from the commit message (ouch).

> So that example is pretty much broken and doesn't work. Right now you should just leave the clsact qdisc as is, since even what that example suggests (probing for zero filters) is inherently racy, as some else could install a filter between check and deletion.

Thanks for clearing that up! In that case I'll simply drop the `del_tc_hook_if_empty()` function for now and let the clsact qdisc hang around (guess it shouldn't cause much harm sitting there unused?).

pping/pping.c
simosund: Sounds like a good idea. However, I've been unable to get `bpf_tc_hook_create()` to actually return `EEXIST`. If the clsact qdisc already exists it still seems to return 0 and chug along nicely (and if it doesn't exist, it of course creates it and returns 0). If an ingress qdisc is installed on the interface it instead seems to return -`EINVAL`.

kkdwivedi: I guess that would be because of missing this... https://lore.kernel.org/bpf/20210612023502.1283837-1-memxor@gmail.com/T/#m7174e4ca28fe08d6ffd204696f9968085f96cdf8

kkdwivedi: So I guess you can still handle it, since it would break when this is applied.

simosund: Ah, ok. Will do.

tohojo:
So this PR is turning into a grab-back of random improvements, not just output improvements. Could you please restructure it so that it actually does what it says on the tin (i.e., contains only the output stuff)? Then we can merge it and you can do the other fixes in a separate PR.

Also, please squash all the small "fixes" commits so that the commit messages are coherent and explain the changes; as a totally random sample, commit 537c9c5b9cd5d93f97384f69b8798206453de9e1 is a candidate for squashing, for instance...

simosund:
> So this PR is turning into a grab-back of random improvements, not just output improvements. Could you please restructure it so that it actually does what it says on the tin (i.e., contains only the output stuff)? Then we can merge it and you can do the other fixes in a separate PR.

You are right about that, think most of the output related stuff have been finished at this point, but have just continued pushing other fixes here as well after that (and also a bit in the middle I guess...). As we discussed in the meeting, I guess we can add a "miscellaneous fixes" PR and move those unrelated fixes there.

So guess the main thing to break out from this PR is using libbpf's tc-attach API together with my most recent fix checking the ACK-flag before parsing TSecr. The refactoring of the cleanup code is also not really related to the output, although it does address one of your requested changes in this PR. I can still try to break that out (and probably squash those two refactoring commits together) and hopefully not break anything too badly in the process.

Adding flow events and packet/byte counts I think is related to improving the output (even if it of course entails some other changes in order to collect the data needed for that improved output).

I guess the commit that adds information about concurrency issues to the README (and removes concurrency comments from the code) isn't exactly related to the output. However as many updates to the README also contains various minor spelling fixes, updates to the concurrency section etc. I think it might be tricky to break that out at this point. 

> Also, please squash all the small "fixes" commits so that the commit messages are coherent and explain the changes; as a totally random sample, commit 537c9c5 is a candidate for squashing, for instance...

I guess that goes for small fixes that are related to, and can relativly easily be included in, previous commits as parts of a larger change. For example, my latest commit about avoiding TSecr if the ACK flag is a small fix, but isn't set isn't really related to anything else in any remotely recent commit (and should as previously stated not even be in this PR).

simosund:
Hi @tohojo, I have rewritten the history a bit now so it should hopefully be a bit easier to follow, and no longer include a bunch of stuff not related to improving the output (or at least reducing it to a much smaller degree). So please have a look at it and let me know if there's anything else you would like me to change before merging.

simosund:
With the recent force-push I just added some fixups that updated the TODO for the "Add timestamp..." and "Add machine friendly format" commits.

PR #17, opened 2021-05-06
Discussing measurement study

simosund:
This pull request is not really for adding any code (at least not for now), but rather just to get a discussion going on how we should use pping to gather measurements and ultimately publish the results in a paper. I've added a document (`MEASUREMENT_STUDY.md`) where I've written down some thoughts around it. 

I've clearly not gotten very far in this process, so the document is mainly open questions for now, but as we decide on how to go about things I also intend to write down those decisions in the document.


simosund:
Direct link to document to make it easier to find: https://github.com/simosund/bpf-examples/blob/Measurement_study/pping/MEASUREMENT_STUDY.md

pping/MEASUREMENT_STUDY.md
```diff
@@ -15,7 +15,32 @@ integrate it with some database or Kafka stream or something? I have very little
...
+bytes per RTT event (depending a bit on type of event, IPv4 or IPv6, values
+etc.). For IPv4 addresses I would expect the "typical" length to be
+approximately 190 bytes for flow-events and 240 bytes for RTT-events.
+
+How much data is written of course depends on the amount and type of
+traffic. For a simple example, let's assume 1000 concurrent non-sparse flows,
+each flow lasting about 1 second on average. Using the default rate limit of 100
+ms and the "typical" values above, that would then generate 10 RTT-events + 2
+flow-events per flow per second, which translates to writing down about 2.8
+MB/s, or 10 GB/h.
```
tohojo, line 32: JSON compresses quite well, so maybe re-check this with compression enabled?

pping/MEASUREMENT_STUDY.md
```diff
@@ -15,7 +15,32 @@ integrate it with some database or Kafka stream or something? I have very little
...
+
+With this in mind, we should probably aim to collect a relatively short trace
+(in the order of hours rather than days/weeks). If we want to collect data over
+a longer time, an alternative could be to increase the rate-limit to ex. 1 s
+(match typical ping), but then we lose a lot of granularity in the data, making
+it hard to analyze shorter flows. It may in fact be more interesting to go in
+the other direction, decreasing the sampling rate to say ex. 10 ms to get more
+fine-grained data, but then the trace will probably have to be in the order of
+minutes.
```
tohojo, line 41: Ideally the sample rate should auto-tune to be similar to the flow RTT I suppose? :)

pping/MEASUREMENT_STUDY.md
```diff
@@ -53,19 +78,166 @@ should that be done independently from the IP-addresses (so that one can still
...
+  would not be suitable for our analysis as we can't isolate single flows
+  anymore, and depending on topology the anonymity may be questionable. This
+  method is for example used by [Google Analytics][google-anonym].
+
+- Crypto-PAn seems to be fairly common way to create a 1-to-1 mapping of
+  addresses and is prefix-preserving (i.e. if the first X bits of two addresses
+  are the same, then the first X bits of their anonymized addresses will also be
+  the same). Not sure this is a property that's particularly useful for our
+  analysis however, and which may weaken the anonymization. This scheme is for
+  example used by CAIDA.
```
tohojo, line 108: For the client IPs it's probably not, but if we're interested in seeing things like whether particular providers suffer from bloat, correlating at the prefix level may be more interesting than on the IP level...

pping/MEASUREMENT_STUDY.md
```diff
@@ -53,19 +78,166 @@ should that be done independently from the IP-addresses (so that one can still
...
+
+It may also be a good idea to obscure the timestamps a bit, for example by
+subtracting the lowest timestamp from all timestamps. I.e. instead of having a
+UNIX-timestamp that tells exactly when the measurements took place, you simple
+have a timestamp that tells you when events occurred relative to the start of
+the measurements. This may make it harder to attempt to combine the measurements
+with external data sources in order to de-anonymize users (ex. using a log from
+a web-server to trace visitors global surfing habits). This probably doesn't add
+as strong protection as lowering resolution or adding noise would, but these
+latter options may pose an issue for the analysis.
```
tohojo, line 130: Even if we want to look at time-of-day type things we could still do something like this. I.e., turning all timestamps into "time since midnight" still gives you time-of-day data without showing which day it was. Same thing for "time since Sunday" if we want weekly data...

pping/MEASUREMENT_STUDY.md
```diff
@@ -53,19 +78,166 @@ should that be done independently from the IP-addresses (so that one can still
...
+
+Furthermore, while I think data size (packet and byte counts) can be useful for
+the analysis, it may be worth to make these optional to collect (partly due to
+data minimization, partly due to storage size for logged data, and partly
+because it may impact performance of pping). In case we in advance decide that
+these are not useful for the analysis, we should not collect them at all.
```
tohojo, line 136: One thing that may be interesting is to correlate latency variation with flow rate, to see if the RTT increases along with the rate (which may indicate bloat)

pping/MEASUREMENT_STUDY.md
```diff
@@ -53,19 +78,166 @@ should that be done independently from the IP-addresses (so that one can still
...
+router and web-server (time between router forwarding client data until server
+ACKs it).
+
+There are a number of other things that one could also use the traffic to
+analyze, but that should maybe not be the focus.
+
+- Traffic types based on ports, including occurrence of HTTP vs HTTPs
+
+- To what degree IPv6 is used (unless we anonymize away the difference between
+  IPv4 and IPv6 addresses)
```
tohojo, line 185: We could explicitly save the address type even if we obscure the actual addresses... Also, seeing whether there's a latency difference between v4/v6 may be interesting?

pping/MEASUREMENT_STUDY.md
```diff
@@ -53,19 +78,166 @@ should that be done independently from the IP-addresses (so that one can still
...
+
+The performance measurements should look at CPU and memory usage for pping
+(i.e. how much resources does original and eBPF version of pping use), as well
+as overall effect on throughput (what is the maximum line rate the pping
+versions can manage).
+
+For the eBPF version of pping it may be of interest to try and time the
+execution time of BPF-programs, as it may be hard to directly measure their CPU
+usage (not being a userspace process). May also just run with pping on/off to
+observe difference.
```
tohojo, line 225: We can measure CPU usage compared to a baseline forwarding scenario. Or, we can see what impact is has on the machine's total forwarding capacity

pping/MEASUREMENT_STUDY.md
```diff
@@ -53,19 +78,166 @@ should that be done independently from the IP-addresses (so that one can still
...
+
+Setting up traffic generators to reach sufficiently high packet rates can
+apparently be challenging. Suggestions for packet generators:
+
+- [TRex](https://trex-tgn.cisco.com/)
```
tohojo, line 230: The in-kernel 'pktgen' can be used to generate base load, but I don't think it supports TCP. But maybe we'd need both? It's certainly easier to use than T-rex IMO :)

pping/MEASUREMENT_STUDY.md
simosund: Hmm, adding an RTT-based sampling interval is on the TODO-list. The question is if it's a feature that will end up being done by the time we run these measurements or not. If it is implemented by then and we run at rate of one RTT-sample per RTT, I guess that will in general result in a higher rate than the default rate limit of 100 ms (as I sure hope most RTTs are below 100 ms). 

But this would probably allow us to overall data in a more useful manner, collecting at a high granularity when it's most likely more relevant. The only potential downsides I see with this is that it would make it harder to predict how much data will be collected (as it then to an even larger extent depends on the traffic), and I guess some types of timeseries analysis may become more tricky?

pping/MEASUREMENT_STUDY.md
simosund: Will update the document after I've done a more thourough check, but when simply running pping with a single 10s iperf3 flow, piping the output through xz seems to reduce the data size by a factor of 10 (from about 25 kB to 2.5 kB).

Of course, then we also need to start to consider the overhead of the compression, so we don't put a too high compression level and eat up a lot of CPU power just compressing our data.

pping/MEASUREMENT_STUDY.md
simosund: Hmm, good point. That would actually be a good usecase for using a prefix-preserving scheme (like CryptoPan).

pping/MEASUREMENT_STUDY.md
simosund: Good idea. Granted it would then be slightly easier to "sync" an external log with the dataset (can check for specific timepoints over a number of days to find a match). Granted even when just using "time since first measurement" it would still be possible to "sync" external datasets by simply looking for matching traffic patterns, so the difference in protection isn't that big.

pping/MEASUREMENT_STUDY.md
simosund: Yes, I also belive it would be useful to keep the packet and bytecounts for this reason. What I meant was rather that it may be a good idea to make them optional metrics, so we can easily decide to not collect them in case we later for whatever reason decide that they're actually not interesting for the analysis.

pping/MEASUREMENT_STUDY.md
simosund: Well, if we expclicitly save the IPv4/IPv6 address type, then we haven't really anonymized away the difference have we? =)

But yhea, don't see a strong case for why we should anonymize away the difference between them, and if the IP-anonymization algorithm itself does that (ex. by using 128-bit encryption) we can of course as you point out explicitly mark the entry with IP-type to keep that information.

Looking at difference in RTT between the traffic types could of course be interesting. Although I guess it might be hard to directly compare them with the IPv4 ones when we don't know which hosts they belong to. Still, guess we could look at ex. the difference in RTT variation within flows etc.

pping/MEASUREMENT_STUDY.md
simosund: Yhea, (unless I missinterpret you), that's pretty much what I meant with running with pping on/off (off being the baseline, and Kathie's pping as comparison).

pping/MEASUREMENT_STUDY.md
simosund: Well, I basically have 0 experience using traffic generators, so very open for suggestions here. Simone has also mentioned MoonGen. So will have to look at some different alternatives here and find what seems most suitable.

We would of course need to be able to generate TCP traffic (with a working implementation of the TCP timestamp option) for the measurements to be particularly relevant. Guess the overhead when pping doesn't do anything useful (i.e. just partially parsing traffic it can't find identifiers for) can also be of some interest, but think we should mainly focus on the overhead for traffic it can generate RTTs for (which should be significantly higher than for other types of traffic).

pping/MEASUREMENT_STUDY.md
simosund: Another aspect for the performance analysis I guess would be to look at which point we start missing RTTs due to the packet-timesamp map being full (or missing entire flows, although that one should be rather 1-to-1 with the size of the flow-state map).

simosund:
Closing this as it's not something I intend to ever merge in the first place. 

Was initially intended as a place for discussion regarding a possible measurement study, but think we have other channels for that these days. Right now it just serves as a place where I've pushed a bunch of not-so-pretty experiment and plotting scripts, and slightly cleaned up versions of the relevant ones (for the PAM paper) are available at https://doi.org/10.5281/zenodo.7590225.

PR #18, opened 2021-05-31
configure: remove requirements on M4 and Emacs

vincentbernat:
They are not used.


tohojo:
Yeah, good point, seems these got carried over from xdp-tools. Thanks for the fix! :)

PR #22, opened 2021-06-10
Add DHCP relay program

sachints123:
Initial working version of DHCP relay using XDP is created.Currently, this code
has user program and a xdp ebpf program. User program takes network interface 
and dhcp relay server IP as inputs and store it in a map. XDP program filters 
the incoming DHCP requests and inserts option 82 in the DHCP request packets and
overwrites the destination IP to that of DHCP relay server IP. An optional 
argument for user program is also provided to unload the xdp program. README 
file provides to instructions to build and load the xdp program.

Signed-off-by: Sachin Tiptur <sachin.tiptur.satyanarayana.gupta@hof-university.de>


dhcp-relay/dhcp-relay.h
```diff
@@ -0,0 +1,63 @@
...
+
+#define COOKIE_SIZE 4	/* Cookie size in DHCP header */
+#define DEST_PORT 0x0043 /*UDP destination port for dhcp */
+
+/* structure for sub-options in option 82*/
+struct sub_option {
+	__u8 option_id;
+	__u8 len;
+	__u16 val;
+	};
```
tohojo, line 35: Indentation error

tohojo (changes requested):
A bunch of comments, most of which are minor-ish style issues. In particular, please fix up the code style throughout. There are tools that can help you with that (like clang-format). You can grab the .clang-format file from the Linux kernel source directory for the correct style...

Also, please remove the binary files from the commit (.dhcp_kern_xdp.c.swp and dhcp_user_xdp). I'd suggest adding a .gitignore for them instead.

For the commit message, please make the title short and descriptive (like "Add DHCP relay program") and reflow the commit message itself to be 80 characters wide. Also please use your full name instead of your github username in the signed-off-by line.

When updating the PR, just update the existing commit (with `git commit --amend` instead of creating new ones on top of this one. If you prefer to keep the history you can of course also create a new local branch and work on the changes, but the PR should stay as one commit...

dhcp-relay/README
```diff
@@ -0,0 +1,16 @@
...
+
+Build instructions:
+cd bpf-examples/dhcp-relay
+sudo make
+
+Loading bpf program:
+sudo ./dhcp_user_xdp <netif> <dhcp relay IP>
+
+unloading program:
+sudo ./dhcp_user_xdp <netif> <dhcp relay IP> 1
```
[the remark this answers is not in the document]
sachints123: Like instead of 1, shall i use on/off strings?  and should I handle using optargs?

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,178 @@
...
+	int ether_type;
+ 	int h_proto;
+ 	int key = 0;
+	int len = 0;
+
+  	if(data+1 > data_end)
+	 	return XDP_ABORTED;
+
+  	nh.pos = data;
+  	ether_type = parse_ethhdr_vlan(&nh, data_end, &eth, &vlans);
```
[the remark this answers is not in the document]
sachints123: If there are no enough vlan tags then should we skip inserting option82?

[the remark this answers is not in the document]
tohojo: Yeah, either skip processing, or go straight to XDP_ABORTED to drop the packet...

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,191 @@
...
+	__u32 csum = bpf_csum_diff((__be32 *)oldip, size, (__be32 *)ip, size,
+				   ~oldcsum);
+	__u32 sum = (csum >> 16) + (csum & 0xffff);
+	sum += (sum >> 16);
+	return sum;
+}
+
+/* Offset to DHCP Options part of the packet */
+#define static_offset                                                          \
+	(sizeof(struct ethhdr) + DHCP_UDP_OVERHEAD +                           \
```
tohojo, line 56: Let's use `sizeof(struct iphdr) + sizeof(struct udphdr)` here instead of `DHCP_UDP_OVERHEAD`

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,191 @@
...
+	if (bpf_xdp_adjust_head(ctx, 0 - delta) < 0)
+		return XDP_ABORTED;
+
+	data_end = (void *)(long)ctx->data_end;
+	data = (void *)(long)ctx->data;
+
+	if (data + offset > data_end)
+		return XDP_ABORTED;
+
+	if (xdp_store_bytes(ctx, 0, buf, MAX_BYTES, 0))
```
tohojo, line 150: Instead of this double-copy we should just extend __bpf_memcpy() to handle the size we need. Just pushed a change to extend it by another 8 bytes, which I think is enough, right? So just rebase and do this in one go with the full size...

dhcp-relay/dhcp_user_xdp.c
```diff
@@ -0,0 +1,193 @@
...
+		printf(" --%-12s", options[i].name);
+		if (options[i].flag != NULL)
+			printf(" flag (internal value:%d)", *options[i].flag);
+		else
+			printf(" short-option: -%c", options[i].val);
+		printf("\n");
+	}
+	printf("Example:\n");
+	printf("To load program:\n dhcp_user_xdp -i eth0 -d 10.0.0.1\n");
+	printf("To unload program:\n dhcp_user_xdp -i eth0 -u\n");
```
tohojo, line 45: Let's use argv[0] here as well instead of hard-coding the program name...

dhcp-relay/dhcp_user_xdp.c
```diff
@@ -0,0 +1,193 @@
...
+	int key = 0;
+	char server[15] = "";
+	struct in_addr addr;
+	__u16 ifindex;
+
+	while ((opt = getopt_long(argc, argv, "hui:d:m:", options, NULL)) !=
+	       -1) {
+		switch (opt) {
+		case 'i':
+			strncpy(dev, optarg, IF_NAMESIZE);
```
tohojo, line 116: If `strncpy()` fails it'll leave the buffer without a terminating NULL byte, leading to subsequent overflows; so you need to do something like:
```C
strncpy(dev, optarg, IF_NAMESIZE-1)
dev[IF_NAMESIZE-1] = '\0';
```

dhcp-relay/dhcp_user_xdp.c
```diff
@@ -0,0 +1,193 @@
...
+				fprintf(stderr,
+					"Couldn't validate IP address:%s\n",
+					optarg);
+				return -EINVAL;
+			}
+			break;
+		case 'm':
+			if (strcmp(optarg, "skb") == 0) {
+				xdp_flags = XDP_FLAGS_SKB_MODE;
+			}
```
tohojo, line 134: Better handle invalid values as well here (and support 'native' as a no-op as well for good measure)

.gitignore
```diff
@@ -3,3 +3,6 @@ config.mk
 compile_commands.json
 *.ll
 *.o
+dhcp-relay/*.o
+dhcp-relay/*.ll
+dhcp-relay/dhcp_user_xdp
```
tohojo, line 8: Better to just create a separate .gitignore file in the dhcp-relay directory...

tohojo (changes requested):
Much better! A few more smallish issues to fix, then I think we're good to go.

Also, you seem to have accidentally included a change to the libbpf submodule in your commit; please get rid of that. And while adding the .clang-format file is fine, it should be in a separate commit, and added to the top-level directory so we can use it everywhere :)

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,186 @@
...
+		goto out;
+
+	h_proto = parse_iphdr(&nh, data_end, &ip);
+
+	/* only handle fixed-size IP header due to static copy */
+	if (h_proto != IPPROTO_UDP || ip->ihl > 5) {
+		goto out;
+	}
+	/*old ip hdr backup for re-calculating the checksum later*/
+	oldip = *ip;
```
tohojo, line 119: If you also store the byte offset here (`ip_offset = ip - data`), you can get back to the IP header after adjusting the packet size without re-parsing the packet (see comment below)

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,186 @@
...
+	len = parse_udphdr(&nh, data_end, &udp);
+	if (len < 0)
+		return XDP_ABORTED;
```
tohojo, line 122: This should just be `goto out`

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,186 @@
...
+
+	data_end = (void *)(long)ctx->data_end;
+	data = (void *)(long)ctx->data;
+
+	if (data + offset > data_end)
+		return XDP_ABORTED;
+
+	if (xdp_store_bytes(ctx, 0, buf, static_offset, 0))
+		goto out;
```
tohojo, line 149: here you've already extended the packet, so needs to return XDP_ABORTED

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,186 @@
...
+
+	if (offset > static_offset) {
+		offset = static_offset;
+		for (i = 0; i < VLAN_MAX_DEPTH; i++) {
+			if (vlans.id[i]) {
+				if (xdp_store_bytes(ctx, offset, buf + offset,
+						    4, 0))
+					goto out;
```
tohojo, line 157: And here as well (`XDP_ABORTED`)

dhcp-relay/dhcp_kern_xdp.c
```diff
@@ -0,0 +1,186 @@
...
+		return XDP_ABORTED;
+
+	nh.pos = data;
+	ether_type = parse_ethhdr_vlan(&nh, data_end, &eth, &vlans);
+	if (ether_type != bpf_htons(ETH_P_IP))
+		goto out;
+
+	h_proto = parse_iphdr(&nh, data_end, &ip);
+	if (h_proto < 0)
+		return XDP_ABORTED;
```
tohojo, line 173: Instead of re-parsing the packet, you can just use the offset (if you store it as suggested above):
```C
ip = data + ip_offset;
if (ip + 1 > data_end)
	return XDP_ABORTED;
```

The range check is needed to appease the verifier, but won't actually ever fail...

dhcp-relay/dhcp_user_xdp.c
```diff
@@ -0,0 +1,197 @@
...
+				fprintf(stderr,
+					"Couldn't validate IP address:%s\n",
+					optarg);
+				return -EINVAL;
+			}
+			break;
+		case 'm':
+			if (strcmp(optarg, "skb") == 0) {
+				xdp_flags = XDP_FLAGS_SKB_MODE;
+			} else {
```
tohojo, line 135: I think it's better to fail on an invalid value; so rather `else if (strcmp(optarg, "drv") != 0) { fprintf(stderr, "Invalid mode: %s\n", optarg); return -EINVAL; }`

dhcp-relay/dhcp_user_xdp.c
```diff
@@ -0,0 +1,197 @@
...
+		return -1;
+	}
+	if (prog_fd <= 0) {
+		printf("ERR: loading file: %s\n");
+		return -1;
+	}
+
+	/* read the map from prog object file and update the realy
+ * server IP to the map
+*/
```
tohojo, line 169: Indentation error; also s/realy/real/ ?

tohojo (changes requested):
Git history looks fine now; found a few more minor issues in the code, please fix. Also, there's a second signed-off-by tag with your username in the commit message, just get rid of that and keep the one with your name on it :)

tohojo (approved):
The .swp file snuck back in and the commit author name doesn't match the sign-off line making the bot unhappy. I'll fix up both of those while applying, the rest looks good. Nice work!

I'd suggest you amend your local git config to avoid those issues in the future. Something like:
```
git config --global user.name 'Sachin Tiptur'
git config --global core.excludesfile ~/.gitexclude
echo '*.swp' >> ~/.gitexclude
```

tohojo:
Merged as 0c8df86a5eeb8dc17088a281ce082651375d8f32

PR #23, opened 2021-06-22
Miscellaneous pping fixes

simosund:
Various different fixes for pping. Currently simply moved the ones that were previously part of the PR for improving the output, but had nothing to do with actually improving the output. Will likely add a bunch of other more or less unrelated improvements here as well.


simosund:
These fixes are based on the improve output PR, so once that is merged I can rebase this to only include the non-output stuff.

tohojo:
Alright, merged the output PR; please rebase this one...

simosund:
> Alright, merged the output PR; please rebase this one...

Ok, should be done now. And by merging it I guess you thought my rebased commit history was ok, but feel free to let me know if there were any problems with it or other ways I could improve it in the future.

I decided to start pushing commits that aren't entierly finished to my branch, partly so you can see what I'm working on if you're interested, and partly so that I have a remote backup also of unfinished work. I've marked those with WIP. This means I'll end up doing quite a few force-pushes (mainly to amend my latest WIP commit, but will probably need to do some additional cleanup before we merge anyways). If you think this is a bad idea, let me know and I'll stop.

simosund:
A couple of potential issues I've noticed that I wanted to write down while I still remeber them. If you have any thoughts on them, feel free to comment.
- I consider the src/dst adress from the context of the sent (timestamped) packet, whereas Kathie considers them from the perspective of the response (matched) packet. In practice this means that I write out the addresses in reverse order of Kathie's pping. I know that I originally noticed and fixed this a while ago, but forgot it again when adding and rewriting all the output formats.
  - For the ppviz format this should definitly be changed so that we actually match the expected format
  - For the "standard"/default output format, it should probably also be changed as it is quite similar to the default output of Kathie's pping
  - For the JSON format, it is a bit unclear to me if it should be changed (as it's the only format that explicitly labels what is src and what is dest). As far as I'm concered, it makes more sense to look at it from the perspective of the sent packet (that is the packet that will experience the reported RTT) rather than the response packet as Kathie does. However it might be good to change it in order to be "consistent" with the other formats (although the other formats don't explicitly label what is the src and dest part, I think most people would assume that the src address comes first, and that is how it's refered to in the [ppviz format documentation](http://www.pollere.net/ppviz.html))

- Are my choices for short/long option flags good (you can find them in long_options at the beginning of pping.c)? I've kinda just selected them add-hoc as I've implemented some new feature, and feel like they may not be optimal in the long run. If pping starts to be used by other people it's probably not a great idea to change flags around at that stage, so should probably try to plan it out a bit early. I should probably also write a better help-message, as right now it just lists the options without any explenation of what they do (can only be found as comments in the source code).

- The user-space cleanup uses the last_timestamp of a flow, which indicates that last time a timestamp was created for the flow (or would have been created in case the timestamp map was full at the moment), to determine if the flow is active or not. If one use a very long sampling interval, it is possible that the userspace could clean up an active flow simply because its rate limit doesn't allow it to create new timestamps. This in turn would allow it to actually bypass the rate-limit as the next packet from the flow would then be considered to come from a new flow after the userspace deleted the previous one. This does not seem like a very serious issue to me because using such a low sampling rate (userspace only cleans up flows after 300 sec) seems quite unlikely. The primary reason I would see for it would be to support a very large amount of flows (make it very unlikely that the timestamp map fills up by simply producing very few timestamps per flow), in which case it might be contra-productive to keep around all the flow-states for a really long time anyways.

simosund:
One small issue I've also noticed is that since the updated libbpf version, pping will now give out the error message "libbpf: Kernel error message: Exclusivity flag on, cannot modify" if there already is a clsact qdisc on the interface (libbpf seems to give that warning simply from attempting to create a tc hook, even if I handle the case with EEXIST). The program still runs as expected though.

simosund:
As quite a few commits with various unrelated changes and fixes have been added to this PR, is is probabably a bit of a nightmare to try and review. I will therefore try to split up the changes in this PR into multiple smaller PRs that I will start up in the comming days. Will keep this PR around as a draft for now simply as a point of reference for the other PRs.

simosund:
Everything in this PR has already been merged through other PRs, so closing this now.

PR #24, opened 2021-06-24
BPF example with ktrace and CO-RE

netoptimizer:
"Learning by doing":
 - Learning what CO-RE can do via writing a simple kprobe tracing program use some of these features.

I had to change the Makefile system to handle `SEC("kprobe")` and macro `BPF_KPROBE` as compiling with `-target bpf` result in `PT_REGS_PARM1` not being defined. As `include <bpf/bpf_tracing.h>` cannot determine the ARCH from compiler defines.
To solve this detect the `ARCH` our makefile system in [file lib/common.mk](../lib/common.mk) (via `uname -m`) and defines the `__TARGET_ARCH_$(ARCH)` in `BPF_CFLAGS`.

This PR also contains some `vmlinux.h` header variants for a number of ARCHs.
Choose to splitup `vmlinux_*.h` headers and only include what was needed for this example.
We will see how much maintenance this cause, and if this is the case we can consider adding the full `vmlinux.h` later.

### DIFF
--- a/headers/vmlinux/arch/arm64/vmlinux.h
+++ b/headers/vmlinux/arch/arm64/vmlinux.h
@@ -1,0 +1,38 @@
+#ifndef __VMLINUX_ARCH_ARM64_H__
+#define __VMLINUX_ARCH_ARM64_H__
+
+#ifdef __BPF_TRACING_H__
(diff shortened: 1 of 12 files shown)


netoptimizer:
Going to merge this as I need to reference some of the commits in emails.

PR #28, opened 2021-11-16
Add TCP Echo Server Example

dave-tucker:
This example illustrates how to use socket redirection

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,9 @@ compile_commands.json
 *.ll
 *.o
 *.skel.h
+
+# Rust
+**/debug/
+**/target/
+**/Cargo.lock
+**/*.rs.bk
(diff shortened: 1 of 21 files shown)


tohojo:
Thanks for the example. But, erm, what does it do? :)

I tried parsing the BPF example, and I'm not actually sure how it works; could you please add a bit more description to the README explaining this? Also, for us non-rustaceans, some hints as to which bit is supposed to be BPF code and which bit is userspace code would be helpful...

netoptimizer:
I'm having a hard time finding the BPF code in this example.
I assume it is in echo/echo-ebpf/src/main.rs, but I simply don't understand this code.
We need some more hand-holding the reader that want to learn from this bpf-example, before this can get merged.

dave-tucker:
Ack. I'm working on addressing your comments and will push something soon. Hopefully later today!

PR #30, opened 2021-11-24
AF_XDP example added some results

netoptimizer:
Describe **why** XDP RX-timestamps are essential for AF_XDP.

Also add some data-point results from different systems, measuring the latency between RX-softirq and AF_XDP processing.

Even-though, as explained, this latency is not important to the real use-case in question. The essential part is moving the time-stamping to XDP, as it makes is possible for the application to relax the latency requirements and process RX-packets in bulk.


netoptimizer:
Need this for a meeting tomorrow morning, thus going to merge this myself shortly.

PR #31, opened 2021-12-08
PPing loader improvements

simosund:
Several improvements and fixes to the loading process of the BPF programs, mainly focused on the loading of TC programs. Based on commits from [PR#23](https://github.com/xdp-project/bpf-examples/pull/23).


tohojo:
Looks mostly good, just a few minor comments:

- You can use `bpf_program__set_autoload()` to avoid both programs being loaded
  even if they're in the same file (I assume it's the verification time that's
  the issue here, right?)

- The limit applied in the last commit seems a bit arbitrary. 1 billion seconds
  (for cleanup) is just shy of 32 years, so why not just enforce it within a
  range of, say, a few days, and allow a value of 0 for disabling cleanup
  entirely? Right now the code seems to allow a value of 0 being passed in, but
  I don't think it handles it correctly?

simosund:
>  You can use `bpf_program__set_autoload()` to avoid both programs being loaded
>   even if they're in the same file (I assume it's the verification time that's
>   the issue here, right?)

Ah, good to know. Will check it out. And yes, in the current state (as of these commits) I think it was mainly to load time I was concerned about as the verifier took a noticeable amount of time verifying the programs. In general I guess it's also a bit unnecessary to load programs that aren't used, even if it doesn't cause much of an issue.

>  The limit applied in the last commit seems a bit arbitrary. 1 billion seconds
>  (for cleanup) is just shy of 32 years, so why not just enforce it within a
>  range of, say, a few days, and allow a value of 0 for disabling cleanup
>  entirely? Right now the code seems to allow a value of 0 being passed in, but
>  I don't think it handles it correctly?

If I remember it correctly, it was a bit of an arbitrarily picked value that was high enough that it should not put any limit on what the user could effectively do but low enough that it wouldn't cause an overflow. Using 0 to disable cleanup seems like a good idea, right now I think passing 0 for the cleanup period simply ends up running the cleanup in a busy loop.

simosund:
Hi @tohojo,
I think I have addressed you comments now.

I have updated the commit that introduced tc/ingress as an alternative to XDP to make use of the `bpf_program__set_autload()` function.

I have also updated the commit which adds bounds to the float arguments to have an upper limit of one week (instead of 1 billion ms or s) and disable the periodical cleanup in case the value 0 is passed.

So feel free to take another look and let me know if there's anything else that can be improved.

pping/pping.c
```diff
@@ -296,66 +297,131 @@ static int xdp_attach(struct bpf_object *obj, const char *sec, int ifindex,
...
+	err = bpf_tc_hook_create(&hook);
+	if (err == -EEXIST)
+		created_hook = false;
+	else if (err)
+		return err;
 
-	if (snprintf(prog_path, sizeof(prog_path), "%s/%s", pin_dir, section) < 0)
-		return -EINVAL;
+	prog_fd = bpf_program__fd(
+		bpf_object__find_program_by_title(obj, sec));
```
tohojo, line 357: The `bpf_object__find_program_by_title()` API is being deprecated entirely, so we should probably switch this over to `bpf_object__find_program_by_name()` instead. See https://lore.kernel.org/r/20211214035931.1148209-1-kuifeng@fb.com

tohojo (changes requested):
Looks good, but let's get rid of the deprecated API usage while we're at it...

pping/pping.c
simosund: Ok, wasn't aware of that, but then it makes sense to fix that. I have however gone on vacation now so might take a little while until I get around to it.

simosund: Hi @tohojo, just wanted to notify you that I've now changed all instances of `bpf_object__find_program_by_title()` to `bpf_object__find_program_by_name()`. As I already had several instances of `bpf_object__find_program_by_title()` before the commits in this PR I decided to simply add this change as a separate commit.

pping/pping_kern.c
```diff
@@ -405,7 +405,7 @@ static void pping_ingress(void *ctx, struct parsing_context *pctx)
 // Programs
 
 // Egress path using TC-BPF
-SEC(SEC_EGRESS_TC)
+SEC("classifier/egress")
```
tohojo, line 408: These should just be `SEC("tc")` as that's the New And Improved(tm) way of doing this; see: https://lore.kernel.org/bpf/270e27b1-e5be-5b1c-b343-51bd644d0747@iogearbox.net/ 

(and yes, you can have two programs with the same section name, that's the point of moving to function name-based lookup :) )

simosund: Hmm, as of right now though changing to SEC("tc") causes the program to fail loading with error: 
```
libbpf: prog 'pping_tc_egress': missing BPF prog type, check ELF section name 'tc'
```
i.e. libbpf is unable to automatically infer the correct program type from the section name. Not sure if that's something that has changed in a newer version of libbpf (i.e. if you updating the libbpf submodule would fix it). 

I could of course explicitly set the program type of each program in the loader, but think it's rather convenient that libbpf automatically infers it. So do you think relying on libbpf to infer the program type is a bad idea and that I should change it to SEC("tc") and then explicitly set the program type in the loader (in which case I guess I should also do that with the XDP program to be consistent), or should I leave it be for now?

tohojo:
Ah yes, you're right, we should update the libbpf version. Just pushed
that change, so try rebasing and see if that helps. Be warned that you
may get some other deprecation errors as well,
bpf_object__get_program_by_title() was not the only function that was deprecated...

simosund:
I've now rebased on the latest master (with the updated libbpf submodule) and think I've fixed all instances where some deprecated libbpf API was used (at least I no longer get any deprecation warnings when compiling).

tohojo (approved):
Alright, awesome; let's merge this, then, so we can make some progress on the other PRs! :)

PR #38, opened 2022-02-01
TC policy example of overriding netstack TXQ

netoptimizer:
Controlling TC qdisc TXQ selection via BPF

Our use-case is as a policy we don't want any traffic generated by the Linux networking stack, to use transmit queue **zero**.  This is connected with the [AF_XDP-interaction](../tree/master/AF_XDP-interaction/) example, what want to **reserve TXQ zero** for high priority traffic.

Cc. @jopbs and @jskrogh


netoptimizer:
I worried about local sockets sending and selecting another TXQ than our TC-BPF program selected.

I've tested it on the `main-ctrl` box (and wrote a more advanced bpftrace script) to verify that TC-BPF program **do** control the TXQ also for local generated traffic.

Look at the kernel code, if no socket queue have been explicitly set  (`sk_tx_queue_mapping`), then our TC-BPF selection is used, and this selection is stored/cached in the socket (`sk_tx_queue_set()`) for all future queue selections (in `netdev_pick_tx()`) for "full" sockets.

Bottomline: Yes, this solution works.

netoptimizer:
Going to merge this so @jopbs can test it as part of Yocto rake build

PR #39, opened 2022-03-06
Update btf example

chenhengqi:


BTF-playground/btf_module_read.c
```diff
@@ -66,17 +65,24 @@ int main(int argc, char **argv)
...
+		pr_err("ERROR(%d): btf__load_module_btf() module_name: %s\n",
 		       err, module_name);
 		goto out;
 	}
 
-        type_id = btf__find_by_name(module_btf, symbol_name);
+	type_id = btf__find_by_name(module_btf, symbol_name);
+	if (type_id < 0) {
+		pr_err("ERROR(%d): btf__find_by_name() symbol_name: %s\n",
+		       err, symbol_name);
```
tohojo, line 76: The error is not actually in `err` at this point, so this print is wrong; but it probably should just be propagated to err, so the program exit code is set correctly...

tohojo (changes requested):
Thank you for the fixes! Just one small nit, otherwise LGTM :)

BTF-playground/btf_module_read.c
chenhengqi: Updated.

netoptimizer:
Thanks for improving this example.
LGTM

tohojo (approved):

PR #41, opened 2022-03-10
PPing optimization: Tracking outstanding timestamps

simosund:
An attempt to improve the performance of ePPing by avoiding unnecessary lookups in the timestamp map. Instead of looking for matching timestamp for every (valid) packet, only look for matches in case we know there are unmatched timestamps for the flow.

I think it already works, but may still need some cleanup or other fixes before this is ready to be merged.

### DIFF
--- a/pping/TODO.md
+++ b/pping/TODO.md
@@ -66,6 +66,9 @@
     - Keeping entries around for a long time allows the map to grow
     unnecessarily large, which slows down the cleaning and may block
     new entries
+- [x] Keep track of outstanding timestamps, only match when necessary
+  - Can avoid doing lookups in timestamp hash map if we know that
+  there are no outstanding (unmatched) timestamps for the flow
 
 
 # Potential issues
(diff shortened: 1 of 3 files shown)


tohojo (approved):
LGTM, but going to hold off on merging until you have some performance data :)

simosund:
So, regarding the performance I've done a couple of tests on our normal testbed (testbed-lenovo) according to our normal setup of running traffic between two endhosts with the machine running ePPing forwarding the traffic between the endhosts. Offloads have been disabled on the forwarding machine as well is iptables and conntrack, and packet processing together with (the userspace) ePPing application has been pinned to a single core.

So the results when not per-flow rate-limit sampling (passing -r 0 to ePPing) are as below, show a very small improvement (except perhaps at 1 flow, but those results are generally not very consistent and the forwarding machine is not really the bottleneck there).
refcount_effect_r=0

When using per flow rate limit sampling (100ms delay between reports, -r 100) the improvement is slightly more noticiable, but still rather small.
refcount_effect_r=100

The results are from 3 runs lasting for 3 minutes each (for which only the last 2 minutes have been included). The points are the mean throughput (as reported by sar on the interface end3f1 where ePPing is running) from all 3 * 120 = 360 points, and the errorbars are standard-deviation (the 95% confidence interval for the mean is not really visible).

PR #44, opened 2022-04-20
pping: Use libxdp to load and attach XDP program

tohojo:
Since we now have libxdp as a submodule, we can switch pping over to using
libxdp for loading its XDP program. This requires switching the order of
attachment, because libxdp needs to be fed the BPF object in an unloaded
state (and will load it as part of attaching).

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -19,6 +19,7 @@ static const char *__doc__ =
 #include <sys/resource.h> // For setting rlmit
 #include <time.h>
 #include <pthread.h>
+#include <xdp/libxdp.h>
 
 #include "json_writer.h"
 #include "pping.h" //common structs for user-space and BPF parts
@@ -74,8 +75,8 @@ struct pping_config {
 	char *packet_map;
 	char *flow_map;
 	char *event_map;
-	int xdp_flags;
 	int ifindex;
+	struct xdp_program *xdp_prog;
 	int ingress_prog_id;
 	int egress_prog_id;
 	char ifname[IF_NAMESIZE];
@@ -258,7 +259,6 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 			break;
 		case 'f':
 			config->force = true;
-			config->xdp_flags &= ~XDP_FLAGS_UPDATE_IF_NOEXIST;
 			break;
 		case 'T':
 			config->bpf_config.track_tcp = true;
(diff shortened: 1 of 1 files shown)


tohojo:
@simosund PTAL :)

pping/pping.c
```diff
@@ -76,6 +77,7 @@ struct pping_config {
 	char *event_map;
 	int xdp_flags;
```
simosund, line 78: As far as I can tell, xdp_flags is no longer used for anything, right? So it should probably be removed.

But a couple of questions related to what the XDP_FLAGS was used for (and could potentially be used for):

1. I'm not that familiar with libxdp, but if I recall correctly it let's you attach multiple XDP-programs to an interface (by chaining them with tail calls?). In that case, I guess we don't need any sort of replacement for no longer being able to "force" overwrite any existing XDP program on the interface. 

2. Right now I never let user's specify if "generic" xdp driver should be used or not (have just used the default behavior which falls back on generic if native is not supported). But if we think it's something that might be useful to add in the future, is there any way to send XDP_FLAGS to libxdp or some other way to affect which driver it tries to attach the XDP program with?

tohojo: > As far as I can tell, xdp_flags is no longer used for anything, right? So it should probably be removed.
> 
> But a couple of questions related to what the XDP_FLAGS was used for (and could potentially be used for):
> 
>     1. I'm not that familiar with libxdp, but if I recall correctly it let's you attach multiple XDP-programs to an interface (by chaining them with tail calls?). In that case, I guess we don't need any sort of replacement for no longer being able to "force" overwrite any existing XDP program on the interface.

Yup, exactly, 'force' is no longer necessary (and actively discouraged).

>     2. Right now I never let user's specify if "generic" xdp driver should be used or not (have just used the default behavior which falls back on generic if native is not supported). But if we think it's something that might be useful to add in the future, is there any way to send XDP_FLAGS to libxdp or some other way to affect which driver it tries to attach the XDP program with?

You can pass a mode to xdp_program__attach(); I currently pass `XDP_MODE_UNSPEC` which lets the kernel pick; but you can pass an explicit mode here instead.

I'll get rid of the flags field for now, thanks for the review :)

tohojo:
There's no rush to merge this, just figured I'd do the conversion now that I finally got around to merging the libxdp support into the repository. So take your time and have a closer look if you want :)

tohojo:
Yeah, never did get that error message resolved upstream. Guess I'll take another crack at it. Will merge this in the meantime so it doesn't get stale once you start moving things around :)

PR #45, opened 2022-04-26
Cleanups for AF_XDP-interaction code

netoptimizer:
Add BTF unit test type program for decoding the BTF info in userspace.
And general cleanups.

I want to remove the "hashmap" code (*but didn't have time*), as it is not being used in this code example (and only adds confusion).
 - The "hashmap" code introduce runtime overhead, when looking up struct members.
 - The code example show how "hashmap" is not necessary, by having `struct xsk_btf_member` store offset and size.
 - Via `struct xsk_btf_member` we get faster code by doing lookups at "setup-time" and let runtime use offset+size directly


netoptimizer:
Going to merge this, then I can finish the work in another branch

PR #46, opened 2022-06-01
Test global func

simosund:
Don't intend this to ever be merged, but just to allow some discussion on how to fix the code so that `parse_packet()` can pass the verifier as a global function.

In general this code is similar to the pping code, but vastly simplified to easier focus on the issue at hand. See the commit messages for some more details.


simosund:
So as Toke has figured out what the problem is (kernel bug) and fixed it [upstream](https://lore.kernel.org/all/20220606075253.28422-1-toke@redhat.com/), I'm closing this.

PR #47, opened 2022-06-21
pping - Reduce verification complexity

simosund:
This PR tries to reduce the verification complexity (number of instructions it takes the verifier to verify the program) by breaking out parts of the program (specifically the packet parsing) into global functions that can be verified independently from the rest of the code. 

The first commit just moves some struct members from the generic `parsing_contex` struct to the `packet_info` struct, partly because it semantically fits better, and partly to better separate the packet parsing functions from functions that operate on the information from the parsed packet. This will make it easier to break out the packet parsing into a global function later, as other functions will no longer require access to the `parsing_context` anymore.

The second commit removes some packet pointers in `packet_info` and instead adds a couple of fields (`ip_len` and `ip_tos`) that the FIB lookup needs. So instead of later retrieving relevant packet pointers and parse the necessary information to do a FIB lookup, parse the required information directly as part of the packet parsing stage and then simply get it from `packet_info` when doing the FIB lookup.

The third commit actually breaks out the packet parsing into global functions (one for tc and one for XDP), and this commit will require [Toke's kernel fix](https://lore.kernel.org/netdev/165462463712.11355.13615369109121225276.git-patchwork-notify@kernel.org/T/) to work if using the XDP ingress hook.

While this PR does sucessfully lower the verification complexity (and makes it possible to run it when compiled with LLVM-14 on kernels > 5.15 without hitting the 1m limit), it's a bit unclear how much of it can really be attributed to using global functions and how much of it is simply from restructuring the code. If one inlines the global functions, it takes ~302k insn to verify the program if compiled with LLVM-13 and ~301k if compiled with LLVM-14. But if called as global functions (as is done here), it takes ~231k (plus 228k for the global function) with LLVM-13 and ~221k (plus ~218k for the global function) with LLVM-14. So while the global functions do help reduce the maximum insn somewhat, total verification time will likely be a bit higher due to having to verify both the global function and the program (which added together takes more instructions).

### DIFF
--- a/pping/pping_kern.c
+++ b/pping/pping_kern.c
@@ -88,20 +88,16 @@ struct parsing_context {
  * in the packet_ts map.
  */
 struct packet_info {
-	union {
-		struct iphdr *iph;
-		struct ipv6hdr *ip6h;
-	};
-	union {
-		struct icmphdr *icmph;
-		struct icmp6hdr *icmp6h;
-		struct tcphdr *tcph;
-	};
 	__u64 time;                  // Arrival time of packet
 	__u32 payload;               // Size of packet data (excluding headers)
 	struct packet_id pid;        // flow + identifier to timestamp (ex. TSval)
 	struct packet_id reply_pid;  // rev. flow + identifier to match against (ex. TSecr)
 	__u32 ingress_ifindex;       // Interface packet arrived on (if is_ingress, otherwise not valid)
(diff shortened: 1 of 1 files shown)


tohojo (approved):
Nice set of cleanups! Doesn't really matter if the improvement comes from the use of global functions, or just generally from better code, does it? Either way it's a win; nice work!

PR #52, opened 2022-08-23
Add samples from Linux

magnus-karlsson:
These two commits adds the two AF_XDP samples that were removed from Linux as libbpf got its xsk support deprecated. These are xdpsock and xsk_fwd. They are pretty much brought over untouched, though I have added a README file for both of them, to make using them simpler. xdpsock is now AF_XDP-example and xsk_fwd is AF_XDP-forwarding.

### DIFF
--- a/AF_XDP-example/xdpsock.c
+++ b/AF_XDP-example/xdpsock.c
@@ -1,0 +1,2055 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright(c) 2017 - 2022 Intel Corporation. */
+
+#include <errno.h>
+#include <getopt.h>
+#include <libgen.h>
+#include <linux/bpf.h>
+#include <linux/err.h>
+#include <linux/if_link.h>
+#include <linux/if_xdp.h>
+#include <linux/if_ether.h>
+#include <linux/ip.h>
+#include <linux/limits.h>
+#include <linux/udp.h>
+#include <arpa/inet.h>
+#include <locale.h>
+#include <net/ethernet.h>
+#include <netinet/ether.h>
+#include <net/if.h>
+#include <poll.h>
+#include <pthread.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/capability.h>
+#include <sys/mman.h>
+#include <sys/resource.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/un.h>
+#include <time.h>
+#include <unistd.h>
+#include <sched.h>
+
+#include <xdp/xsk.h>
+#include <xdp/libxdp.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "xdpsock.h"
+
+#ifndef SOL_XDP
+#define SOL_XDP 283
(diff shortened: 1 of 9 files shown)


netoptimizer:
I'm okay with the directory naming `AF_XDP-example/` and `AF_XDP-forwarding/` for the programs.

I'm not a fan of the `xdpsock` name, but we can keep it given this came from the kernel tree.

magnus-karlsson:
Toke, any comments?

tohojo (changes requested):
I'm OK with merging these mostly as-is, with couple of nits:

- Please convert the README files to README.org and at a minimum add headings to both of them. People will be consuming this from the github web interface, so nicely formatted README files makes things easier to read.
- Let's complete the conversion to libxdp. I.e., use the libxdp APIs for the manual load/unload of XDP programs that the utilities do, and `s/libbpf_flags/libxdp_flags/`

AF_XDP-forwarding/xsk_fwd.c
```diff
@@ -0,0 +1,1085 @@
...
+	quit = 1;
+}
+
+static void remove_xdp_program(void)
+{
+	int i;
+
+	for (i = 0 ; i < n_ports; i++)
+		bpf_xdp_detach(if_nametoindex(port_params[i].iface),
+			       port_params[i].xsk_cfg.xdp_flags, NULL);
```
tohojo, line 975: This won't actually work correctly; the XSK socket will load the XDP program with a dispatcher, but here we're just detaching the program fd from the interface, so all the pinned stuff will stay around in bpffs.

I see two options here:
- Wait for the refcnt PR to be merged into libxdp, and just get rid of this removal logic entirely
- Change this to do `xdp_multiprog__detach(xdp_multiprog__get_from_ifindex))` instead; see the `do_unload()` function in xdp-loader for the `opt->all` case.

WDYT?

magnus-karlsson: I will change it to xdp_multiprog__detach() as I do not know when the refcnt PR will be merged. We can always remove this logic afterwards.

AF_XDP-example/xdpsock.c
```diff
@@ -0,0 +1,2062 @@
...
+	struct xdp_multiprog *mp;
+	int err;
+
+	mp = xdp_multiprog__get_from_ifindex(opt_ifindex);
+	if (IS_ERR_OR_NULL(mp)) {
+		printf("No XDP program loaded on %s\n", opt_if);
+		return;
+	}
+
+	err = xdp_multiprog__detach(mp);
```
tohojo, line 582: I don't think we should be changing this to do the full interface multiprog detach; in this example we have a reference to the xdp_prog, so just doing a regular `xdp_program__detach()` is fine (and will remove the dispatcher if it's the only program loaded).

tohojo (changes requested):

AF_XDP-example/xdpsock.c
magnus-karlsson: Yes, you are correct. That is a bit too drastic. New version pushed.

tohojo (approved):

magnus-karlsson:
I can fix these before merging. Will submit a new version.

Thanks: Magnus

PR #53, opened 2022-08-31
BTF playground for extracting Kernel BTF object ID

netoptimizer:
Userspace code (libbpf based) experimenting with looking up the Kernel's BTF **object** *ID* for kernel modules.

This is needed as part of the XDP-hints effort as we are looking at having a "full" u64 BTF ID that comprise of the BTF object ID + BTF type ID (inside the object ID).

- This is done to be able to fully identify which kernel module created the BTF ID.

### DIFF
--- a/BTF-playground/Makefile
+++ b/BTF-playground/Makefile
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-2.0
 
 USER_TARGETS := btf_module_read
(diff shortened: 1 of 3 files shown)


netoptimizer:
Hi @maryamtahhan please notice this PR

netoptimizer:
Going to merge this, so I can easier point @anakryiko to look at my approach

PR #55, opened 2022-09-01
PPing: Add option to ignore SYN-packets

simosund:
Add an option for ePPing to ignore TCP SYN packets, so that the
initial handshake phase of the connection is ignored. More details in the commit message.

For now I've only made it an opt-in option, and by default ePPing will still track SYN-packets, and any SYN-flood attacks would cause it to fall apart. The trace used in the paper ["Continuous in-network round-trip time monitoring"](https://dl.acm.org/doi/10.1145/3544216.3544222) suggests that incomplete TCP connections are quite common, so it might be worth considering if ignoring the SYN packets should be the default option.


tohojo (changes requested):
Hmm, yeah, let's make the "safe" alternative the default and make tracking SYNs opt-in...

PR #56, opened 2022-09-02
Prepare for newer libbpf in bpf-examples

netoptimizer:
Newer kernels got a new BTF type "kind" (BTF_KIND_ENUM64) and needs a newer libbpf version, else it fails decoding BTF provided by the Linux kernel.

Running libbpf with verbose/debug print statements shows this error:
```
libbpf: Unsupported BTF_KIND:19
libbpf: loading kernel BTF '/boot/vmlinux-5.19.0-bpf-next-xdp-hints+': -22
libbpf: failed to find valid kernel BTF
```

Before we can upgrade the libbpf version, we need to fixup the bpf-examples to use the newer APIs. This PR is a step in that direction, meaning the API conversion isn't fully done.

### DIFF
--- a/BTF-playground/btf_module_ids.c
+++ b/BTF-playground/btf_module_ids.c
@@ -217,6 +217,9 @@ int find_btf_id_by_name(const char *btf_name, int *btf_size)
 		if (info.name_len == 0) /* Skip non/empty names */
 			continue;
 
+		if (info.kernel_btf == 0) /* Looking for kernel module BTF */
+			continue;
+
 		name = u64_to_ptr(info.name);
 		if (strncmp(name, btf_name, 127) == 0) {
 
--- a/BTF-playground/btf_module_ids.c
+++ b/BTF-playground/btf_module_ids.c
@@ -121,26 +121,6 @@ int __btf_obj_info_via_fd(int fd, struct bpf_btf_info *info)
 #undef SZ
 }
 
-struct btf *open_vmlinux_btf(void)
-{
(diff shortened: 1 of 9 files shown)


tohojo (approved):
Nice set of cleanups. I guess now that we have more things in the top-level Makefile, we should add a CI job to at least compile everything so we catch any regressions going forward...

Also the original issue you're fixing implies that libbpf is not forwards-compatible. I.e., if you upgrade your kernel and that adds new BTF types, old programs will break when trying to load the vmlinux BTF... So much for CO-RE providing API stability across kernel versions :(

netoptimizer:
> Nice set of cleanups. I guess now that we have more things in the top-level Makefile, we should add a CI job to at least compile everything so we catch any regressions going forward...
> 

Yes, that is a good idea! :+1: 

> Also the original issue you're fixing implies that libbpf is not forwards-compatible. I.e., if you upgrade your kernel and that adds new BTF types, old programs will break when trying to load the vmlinux BTF... So much for CO-RE providing API stability across kernel versions :(

Yes, unfortunately.
I want to mention that above error happens to me with libbpf v0.7.0 on a new kernel, but upgrading to libbpf v0.8.0 solves the issue (so, I'm not forced to use v1.0.0).

netoptimizer:
I'm merging this as I'm unlikely to have time before LPC.
We are closer to being able/read to upgrade to libbpf 0.8.0 but not fully done.

Issue #49, opened 2022-06-22
Unclear kernel requirements for ePPing


simosund:
With the recent [PR#47](https://github.com/xdp-project/bpf-examples/pull/47), ePPing requires a very recent [kernel patch](https://lore.kernel.org/netdev/20220606075253.28422-1-toke@redhat.com/) to run with XDP (which is the default), which I'm guessing very few Linux devices out there currently have. If attempting to run ePPing (in XDP mode) without this patch the verifier will reject the program (with a somewhat confusing error message), and it may be very unclear for a user what the issue is.

So should probably clarify that the issue is due to an unsupported kernel version, by for example adding an error message that includes suggestions to try with a newer kernel (or a kernel with the mentioned patch backported), or possibly to try and run it with the tc hook instead. 

Mostly leaving this as a TODO for myself, so I don't forget it when I get back from my vacation.

PR #57, opened 2022-09-05
PPing minor userspace fixes

simosund:
A few unrelated minor fixes to the userspace component of ePPing. Threw them together in the same PR to avoid lots of tiny PRs.

- The first commit fixes an issue where the clsact would not be cleaned up if using tc mode (--ingress-hook tc)
- The second commit adds a small hint suggesting trying a newer kernel or using tc mode if it fails loading the XDP program. This addresses #49.
- The third commit simply adds a couple of calls to `bfp_object__close()` on shutdown/error. I'm not sure if this is really needed when the program ends directly after anyways, but most recent examples seem to do it so figured I'd better do it as well in case it cleans up some resources that's not automatically freed by the kernel when the program ends.

As for the second commit, I'm not sure if it's the best way to address #49. Simone suggested to automatically try to load and attach the tc ingress program in case the XDP one fails. While this should be possible (although a bit more of a hassle) and in some cases quite convenient, I'm not sure if automatically changing user-specified arguments is a good idea (even if it's not done silently) as it might be easy to miss unless paying close attention to stderr, especially if ePPing is started via ex. a script.

A somewhat related discussion is if we should perhaps always try to load the XDP program in native mode (`XDP_MODE_NATIVE`), instead of in unspecified mode (`XDP_MODE_UNSPEC`) which will (silently) fall back on generic XDP if there's no driver support. From some old discussions with Jesper I got the impression that the performance of XDP generic is rather lackluster, and it probably being better to run with tc than with generic XDP. If that's the case I'm not sure if it makes any sense to allow running in XDP generic mode (and also being unclear to the user whether it's is running in native or generic).

Finally, should XDP mode really be the default? It's less likely to work (due to requiring a more recent kernel), does (still not?) work with ex. GRO and may offer worse performance than tc if running in generic mode. Even if it's running in native XDP mode do we really expect it to have much of a performance benefit here anyways? As ePPing only monitors the packets it's always letting all packets through to the kernel anyways (always return `XDP_PASS`), so we don't really benefit from being able to create a fastpath that bypasses the kernel like some other XDP use cases might. So perhaps it makes more sense to have tc as the default instead (which can be changed to XDP by passing -I/--ingress-hook xdp)?

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -62,6 +62,7 @@ struct map_cleanup_args {
 	struct bpf_link *tsclean_link;
 	struct bpf_link *flowclean_link;
 	__u64 cleanup_interval;
+	int err;
 	bool valid_thread;
 };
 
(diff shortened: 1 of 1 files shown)


simosund:
So, I've expanded this PR a bit. A brief overview of the commits as of now:
- The first commit fixes an issue where the clsact qdisc would not be cleaned up if using tc ingress (-I/--ingress-hook tc).
- The second commit `#define`s the BPF program names in the user space loader to avoid them being hardcoded in multiple places. The idea is that if the program names were to change, it might lead to (potentially sneaky) errors if the program names are not correctly updated in all the places. Ideally both the BPF progs and the user space side would use names from a single place (ex. defined in pping.h), but I'm not aware of a way to use a single definition that works both as a (function) identifier and as a string.
- The third commit changes ePPing to use the tc ingress hook instead of XDP as default (as TC should be more likely to work correctly, and for our use case we shouldn't expect XDP to provide a considerable performance advantage)
- The fourth commit adds the option to chose which mode to load the XDP program in (set to native by default). I'm not sure if this option is really needed (can't see much of a use case with running ePPing with XDP in any other mode than native). I've also left "hardware" mode out as I suspect that will not work well with the rest of ePPing (still need to share maps tc egress etc.), although I know too little about hardware offloaded XDP to say if this is really the case. Finally, as this commit does two things (adds the option and changes default from unspecified to native) it should arguably be two separate commits, but figured that this PR already contains plenty of small commits and these were highly related. 
- The fifth commit simply adds an additional error message/hint in case the XDP load fails. After the change to use tc ingress as default, I'm not sure to what degree this is needed anymore. I suspect it could also be improved by checking the actual error code to distinguish between different errors instead of the very simple assumption it does right now based on XDP mode.
- The sixth commit adds calls to `bpf_object__close` on error/shutdown. Not sure how necessary this is as I would assume the kernel would free up most resources when the program terminates anyways, but as other examples do so I assume it's good practice.
- The seventh commit aborts the program if the periodical map cleanup fails. While a single failure to run the cleanup might not be too problematic, in case the cleanup fails constantly the maps will eventually fill up with stale entries. This would leave ePPing in a state where it still adds overhead due to processing every packet, but being unable to will be unable to provide any measurements as it will be unable to add new flows/timestamps. Therefore, it probably makes more sense to just abort the program as soon as the cleanup fails rather than letting it keep running and hope it recovers.

simosund:
Yes, had not rebased it after you #55 and #62, but fixed that now.

tohojo (approved):

PR #59, opened 2022-09-29
Add aggregation option to ePPing

simosund:
This PR adds an initial simple version of aggregating RTTs (instead of dumping every single RTT sample as normally done). It aggregates all RTTs (no option to filter or group the aggregation yet) by adding all samples to a histogram and keeping track of minimum and maximum RTT.

The histogram is of fixed size (`pping.h:AGG_RTT_NR_BINS`) and uses equally wide bins (`pping.h:AGG_RTT_BIN_WIDTH`). The current setting is 1000 bins that are 0.1 ms wide, allowing ePPing to capture RTTs from 0-100ms with 0.1ms resolution (the last bin is half-open, so RTTs > 100 ms will also go there). There is currently no option to configure the histogram from the command line, but it would be easy to at least make the bin-width configurable (nr of bins is trickier as it affects the size of structs used in the map).

- The first commit is just a minor refactor which makes the BPF code easier to follow once the aggregation is added in the next commit.
- The second commit adds an initial version of aggregation which is never reset (so each report contains all RTTs since the start of ePPing).
- The third commit resets the aggregation stats after each report, so each report only contains the RTTs since the last report. To avoid concurrency issues between the userspace and BPF progs two aggregation maps are used so that the BPF progs can use one map while the userspace is reading and clearing the other.

Another point worth mentioning is that the periodical aggregation runs in a separate thread in userspace. Initially I planned to simply replace the main loop in userspace which processes the events with the periodical aggregation instead. However, in addition to RTT and flow events the BPF progs also sends debug/warning events through the same perf-buffer, so the main loop for processing events is still needed to handle these.

A potentially nice side effect of the aggregation running in a separate thread is that it should be possible to both report individual RTTs and aggregated stats at the same time, however I've currently made them mutually exclusive for two reasons:
- The userspace does not write the reports "atomically", parts of the individual and aggregated RTT reports may be split and interleaved.
- I have not bothered to figure out how the aggregated reports should work together with the different formats (default, JSON, ppviz).
So for now using aggregated RTTs explicitly disables reporting of individual RTTs.

Could you a quick look at this relatively soon-ish @tohojo to see if it seems reasonable and does contains any major flaws/errors? It doesn't have to be merged any time soon, but it would be nice if it's not complete rubbish as I start testing the performance of it to have something for the upcoming paper.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -23,6 +23,8 @@ static const char *__doc__ =
 #include <sys/signalfd.h>
 #include <sys/timerfd.h>
 #include <sys/epoll.h>
+#include <linux/unistd.h>
+#include <linux/membarrier.h>
 
 #include "json_writer.h"
 #include "pping.h" //common structs for user-space and BPF parts
@@ -105,8 +107,9 @@ struct aggregation_config {
 };
 
 struct aggregation_maps {
-	int map_v4_fd;
-	int map_v6_fd;
+	int map_active_fd;
+	int map_v4_fd[2];
+	int map_v6_fd[2];
 };
 
 // Store configuration values in struct to easily pass around
@@ -1068,9 +1071,9 @@ static void print_histogram(FILE *stream,
 {
 	int i;
 
-	fprintf(stream, "[%llu", rtt_stats->bins[0]);
+	fprintf(stream, "[%u", rtt_stats->bins[0]);
 	for (i = 1; i < n_bins; i++)
-		fprintf(stream, ",%llu", rtt_stats->bins[i]);
+		fprintf(stream, ",%u", rtt_stats->bins[i]);
 	fprintf(stream, "]");
 }
 
@@ -1121,6 +1124,41 @@ merge_percpu_aggreated_rtts(struct aggregated_rtt_stats *percpu_stats,
 	}
 }
 
+// Stolen from BPF selftests
+int kern_sync_rcu(void)
+{
+	return syscall(__NR_membarrier, MEMBARRIER_CMD_SHARED, 0, 0);
+}
+
+/* Changes which map the BPF progs use to aggregate the RTTs in.
+ * On success returns the map idx that the BPF progs used BEFORE the switch
+ * (and thus the map filled with data up until the switch, but no longer
+ * beeing activly used by the BPF progs).
+ * On failure returns a negative error code */
+static int switch_agg_map(int map_active_fd)
+{
+	__u32 prev_map, next_map, key = 0;
+	int err;
+
+	// Get current map being used by BPF progs
+	err = bpf_map_lookup_elem(map_active_fd, &key, &prev_map);
+	if (err)
+		return err;
+
+	// Swap map being used by BPF progs to agg RTTs in
+	next_map = prev_map == 1 ? 0 : 1;
+	err = bpf_map_update_elem(map_active_fd, &key, &next_map, BPF_EXIST);
+	if (err)
+		return err;
+
+	// Wait for current BPF programs to finish
+	// This should garantuee that after this call no BPF progs will attempt
+	// to update the now inactive maps
+	kern_sync_rcu();
+
+	return prev_map;
+}
+
 static void report_aggregated_rtt_mapentry(
 	struct ipprefix_key *prefix, struct aggregated_rtt_stats *percpu_stats,
 	int n_cpus, int af, __u8 prefix_len, __u64 t_monotonic,
@@ -1131,10 +1169,14 @@ static void report_aggregated_rtt_mapentry(
 	merge_percpu_aggreated_rtts(percpu_stats, &merged_stats, n_cpus,
 				    agg_conf->n_bins);
 
-	// Only print prefixes which have RTT samples
-	if (!aggregated_rtt_stats_empty(&merged_stats))
+	// Only print and clear prefixes which have RTT samples
+	if (!aggregated_rtt_stats_empty(&merged_stats)) {
 		print_aggregated_rtts(stdout, t_monotonic, prefix, af,
 				      prefix_len, &merged_stats, agg_conf);
+
+		// Clear out the reported stats
+		memset(percpu_stats, 0, sizeof(*percpu_stats) * n_cpus);
+	}
 }
 
 static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
@@ -1149,6 +1191,8 @@ static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
 	__u32 count = AGG_BATCH_SIZE;
 	bool remaining_entries = true;
 	int err = 0, i;
+
+	DECLARE_LIBBPF_OPTS(bpf_map_batch_opts, batch_opts, .flags = BPF_EXIST);
 
 	values = calloc(n_cpus, sizeof(*values) * AGG_BATCH_SIZE);
 	keys = calloc(AGG_BATCH_SIZE, keysize);
@@ -1174,6 +1218,12 @@ static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
 						       t_monotonic, agg_conf);
 		}
 
+		// Update cleared stats
+		err = bpf_map_update_batch(map_fd, keys, values, &count,
+					   &batch_opts);
+		if (err)
+			goto exit;
+
 		total += count;
 		count = AGG_BATCH_SIZE; // Ensure we always try to fetch full batch
 	}
@@ -1188,14 +1238,18 @@ static int report_aggregated_rtts(struct aggregation_maps *maps,
 				  struct aggregation_config *agg_conf)
 {
 	__u64 t = get_time_ns(CLOCK_MONOTONIC);
-	int err;
+	int err, map_idx;
 
-	err = report_aggregated_rtt_map(maps->map_v4_fd, AF_INET,
+	map_idx = switch_agg_map(maps->map_active_fd);
+	if (map_idx < 0)
+		return map_idx;
+
+	err = report_aggregated_rtt_map(maps->map_v4_fd[map_idx], AF_INET,
 					agg_conf->ipv4_prefix_len, t, agg_conf);
 	if (err)
 		return err;
 
-	err = report_aggregated_rtt_map(maps->map_v6_fd, AF_INET6,
+	err = report_aggregated_rtt_map(maps->map_v6_fd[map_idx], AF_INET6,
 					agg_conf->ipv6_prefix_len, t, agg_conf);
 	return err;
 }
@@ -1466,12 +1520,25 @@ static int handle_pipefd(int pipe_rfd)
 int fetch_aggregation_map_fds(struct bpf_object *obj,
 			      struct aggregation_maps *maps)
 {
-	maps->map_v4_fd = bpf_object__find_map_fd_by_name(obj, "map_v4_agg");
-	maps->map_v6_fd = bpf_object__find_map_fd_by_name(obj, "map_v6_agg");
+	maps->map_active_fd =
+		bpf_object__find_map_fd_by_name(obj, "map_active_agg_instance");
+	maps->map_v4_fd[0] =
+		bpf_object__find_map_fd_by_name(obj, "map_v4_agg1");
+	maps->map_v4_fd[1] =
+		bpf_object__find_map_fd_by_name(obj, "map_v4_agg2");
+	maps->map_v6_fd[0] =
+		bpf_object__find_map_fd_by_name(obj, "map_v6_agg1");
+	maps->map_v6_fd[1] =
+		bpf_object__find_map_fd_by_name(obj, "map_v6_agg2");
 
-	if (maps->map_v4_fd < 0 || maps->map_v6_fd < 0) {
-		fprintf(stderr, "Unable to find aggregation maps (%d/%d).\n",
-			maps->map_v4_fd, maps->map_v6_fd);
+	if (maps->map_active_fd < 0 || maps->map_v4_fd[0] < 0 ||
+	    maps->map_v4_fd[1] < 0 || maps->map_v6_fd[0] < 0 ||
+	    maps->map_v6_fd[1] < 0) {
+		fprintf(stderr,
+			"Unable to find aggregation maps (%d/%d/%d/%d/%d).\n",
+			maps->map_active_fd, maps->map_v4_fd[0],
+			maps->map_v4_fd[1], maps->map_v6_fd[0],
+			maps->map_v6_fd[1]);
 		return -ENOENT;
 	}
 
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -1262,12 +1262,20 @@ static void report_aggregated_rtt_mapentry(
 	struct aggregation_config *agg_conf, bool *del_entry)
 {
 	struct aggregated_rtt_stats merged_stats;
+	struct ipprefix_key backup_key = { 0 };
 	int i;
+
+	// Report the backup keys as 0.0.0.0/0 or ::/0
+	if ((af == AF_INET && prefix->v4 == IPV4_BACKUP_KEY) ||
+	    (af == AF_INET6 && prefix->v6 == IPV6_BACKUP_KEY)) {
+		prefix = &backup_key;
+		prefix_len = 0;
+	}
 
 	merge_percpu_aggreated_rtts(percpu_stats, &merged_stats, n_cpus,
 				    agg_conf->n_bins);
 
-	if (prefix_len > 0 && // There's no point in deleting /0 entries (can only be one of them)
+	if (prefix_len > 0 && // Pointless deleting /0 entry, and ensures backup keys are never deleted
 	    agg_conf->timeout_interval > 0 &&
 	    merged_stats.last_updated < t_monotonic &&
 	    t_monotonic - merged_stats.last_updated >
@@ -1667,6 +1675,46 @@ int fetch_aggregation_map_fds(struct bpf_object *obj,
 	return 0;
 }
 
+static int init_agg_backup_entries(struct aggregation_maps *maps, bool ipv4,
+				   bool ipv6)
+{
+	struct aggregated_rtt_stats *empty_stats;
+	struct ipprefix_key key;
+	int instance, err = 0;
+
+	empty_stats = calloc(libbpf_num_possible_cpus(), sizeof(*empty_stats));
+	if (!empty_stats)
+		return -errno;
+
+	if (ipv4) {
+		key.v4 = IPV4_BACKUP_KEY;
+
+		for (instance = 0; instance < 2; instance++) {
+			err = bpf_map_update_elem(maps->map_v4_fd[instance],
+						  &key, empty_stats,
+						  BPF_NOEXIST);
+			if (err)
+				goto exit;
+		}
+	}
+
+	if (ipv6) {
+		key.v6 = IPV6_BACKUP_KEY;
+
+		for (instance = 0; instance < 2; instance++) {
+			err = bpf_map_update_elem(maps->map_v6_fd[instance],
+						  &key, empty_stats,
+						  BPF_NOEXIST);
+			if (err)
+				goto exit;
+		}
+	}
+
+exit:
+	free(empty_stats);
+	return err;
+}
+
 static int setup_timer(__u64 init_delay_ns, __u64 interval_ns)
 {
 	struct itimerspec timercfg = {
@@ -1700,6 +1748,16 @@ static int init_aggregation_timer(struct bpf_object *obj,
 	err = fetch_aggregation_map_fds(obj, &config->agg_maps);
 	if (err) {
 		fprintf(stderr, "Failed fetching aggregation maps: %s\n",
+			get_libbpf_strerror(err));
+		return err;
+	}
+
+	err = init_agg_backup_entries(&config->agg_maps,
+				      config->agg_conf.ipv4_prefix_len > 0,
+				      config->agg_conf.ipv6_prefix_len > 0);
+	if (err) {
+		fprintf(stderr,
+			"Failed initalized backup entries in aggregation maps: %s\n",
 			get_libbpf_strerror(err));
 		return err;
 	}
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -1095,6 +1095,13 @@ static bool aggregated_rtt_stats_nortts(struct aggregated_rtt_stats *stats)
 	return stats->max == 0;
 }
 
+static __u64 aggregated_rtt_stats_maxbins(struct aggregated_rtt_stats *stats,
+					  __u64 bin_width, __u64 n_bins)
+{
+	return stats->max / bin_width < n_bins ? stats->max / bin_width + 1 :
+						 n_bins;
+}
+
 static void
 merge_percpu_aggreated_rtts(struct aggregated_rtt_stats *percpu_stats,
 			    struct aggregated_rtt_stats *merged_stats,
@@ -1141,7 +1148,8 @@ static void print_aggrtts_standard(FILE *stream, __u64 t, const char *prefixstr,
 				   struct aggregated_rtt_stats *rtt_stats,
 				   struct aggregation_config *agg_conf)
 {
-	__u64 nb = agg_conf->n_bins, bw = agg_conf->bin_width;
+	__u64 bw = agg_conf->bin_width;
+	__u64 nb = aggregated_rtt_stats_maxbins(rtt_stats, bw, agg_conf->n_bins);
 
 	print_ns_datetime(stream, t);
 	fprintf(stream,
@@ -1170,7 +1178,8 @@ static void print_aggrtts_json(json_writer_t *ctx, __u64 t,
 			       struct aggregated_rtt_stats *rtt_stats,
 			       struct aggregation_config *agg_conf)
 {
-	__u64 nb = agg_conf->n_bins, bw = agg_conf->bin_width;
+	__u64 bw = agg_conf->bin_width;
+	__u64 nb = aggregated_rtt_stats_maxbins(rtt_stats, bw, agg_conf->n_bins);
 	int i;
 
 	jsonw_start_object(ctx);
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -69,6 +69,7 @@ static const char *__doc__ =
 #define PPING_ABORT 5555
 
 #define ARG_AGG_REVERSE 256
+#define AGG_ARG_TIMEOUT 257
 
 enum pping_output_format {
 	PPING_OUTPUT_STANDARD,
@@ -104,6 +105,7 @@ struct map_cleanup_args {
 
 struct aggregation_config {
 	__u64 aggregation_interval;
+	__u64 timeout_interval;
 	__u64 n_bins;
 	__u64 bin_width;
 	__u8 ipv4_prefix_len;
@@ -166,6 +168,7 @@ static const struct option long_options[] = {
         { "aggregate-subnets-v4", required_argument, NULL, '4' }, // Set the subnet size for IPv4 when aggregating (default 24)
 	{ "aggregate-subnets-v6", required_argument, NULL, '6' }, // Set the subnet size for IPv6 when aggregating (default 48)
 	{ "aggregate-reverse",    no_argument,       NULL, ARG_AGG_REVERSE }, // Aggregate RTTs by dst IP of reply packet (instead of src like default)
+	{ "aggregate-timeout",    required_argument, NULL, AGG_ARG_TIMEOUT }, // Interval for timing out subnet entries in seconds (default 30s)
 	{ 0, 0, NULL, 0 }
 };
 
@@ -414,6 +417,15 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 			break;
 		case ARG_AGG_REVERSE:
 			config->bpf_config.agg_by_dst = true;
+			break;
+		case AGG_ARG_TIMEOUT:
+			err = parse_bounded_long(&user_int, optarg, 0,
+						 7 * S_PER_DAY,
+						 "aggregate-timeout");
+			if (err)
+				return -EINVAL;
+			config->agg_conf.timeout_interval =
+				user_int * NS_PER_SECOND;
 			break;
 		case 'h':
 			printf("HELP:\n");
@@ -1152,6 +1164,10 @@ merge_percpu_aggreated_rtts(struct aggregated_rtt_stats *percpu_stats,
 	memset(merged_stats, 0, sizeof(*merged_stats));
 
 	for (i = 0; i < n_cpus; i++) {
+		if (percpu_stats[i].last_updated > merged_stats->last_updated)
+			merged_stats->last_updated =
+				percpu_stats[i].last_updated;
+
 		if (aggregated_rtt_stats_empty(&percpu_stats[i]))
 			continue;
 
@@ -1164,6 +1180,13 @@ merge_percpu_aggreated_rtts(struct aggregated_rtt_stats *percpu_stats,
 		for (bin = 0; bin < n_bins; bin++)
 			merged_stats->bins[bin] += percpu_stats[i].bins[bin];
 	}
+}
+
+static void clear_aggregated_rtts(struct aggregated_rtt_stats *stats)
+{
+	__u64 last_updated = stats->last_updated;
+	memset(stats, 0, sizeof(*stats));
+	stats->last_updated = last_updated;
 }
 
 // Stolen from BPF selftests
@@ -1204,12 +1227,22 @@ static int switch_agg_map(int map_active_fd)
 static void report_aggregated_rtt_mapentry(
 	struct ipprefix_key *prefix, struct aggregated_rtt_stats *percpu_stats,
 	int n_cpus, int af, __u8 prefix_len, __u64 t_monotonic,
-	struct aggregation_config *agg_conf)
+	struct aggregation_config *agg_conf, bool *del_entry)
 {
 	struct aggregated_rtt_stats merged_stats;
+	int i;
 
 	merge_percpu_aggreated_rtts(percpu_stats, &merged_stats, n_cpus,
 				    agg_conf->n_bins);
+
+	if (prefix_len > 0 && // There's no point in deleting /0 entries (can only be one of them)
+	    agg_conf->timeout_interval > 0 &&
+	    merged_stats.last_updated < t_monotonic &&
+	    t_monotonic - merged_stats.last_updated >
+		    agg_conf->timeout_interval)
+		*del_entry = true;
+	else
+		*del_entry = false;
 
 	// Only print and clear prefixes which have RTT samples
 	if (!aggregated_rtt_stats_empty(&merged_stats)) {
@@ -1217,7 +1250,10 @@ static void report_aggregated_rtt_mapentry(
 				      &merged_stats, agg_conf);
 
 		// Clear out the reported stats
-		memset(percpu_stats, 0, sizeof(*percpu_stats) * n_cpus);
+		if (!*del_entry)
+			for (i = 0; i < n_cpus; i++) {
+				clear_aggregated_rtts(&percpu_stats[i]);
+			}
 	}
 }
 
@@ -1226,19 +1262,21 @@ static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
 				     struct aggregation_config *agg_conf)
 {
 	struct aggregated_rtt_stats *values = NULL;
-	void *keys = NULL;
+	void *keys = NULL, *del_keys = NULL;
 	int n_cpus = libbpf_num_possible_cpus();
 	size_t keysize = af == AF_INET ? sizeof(__u32) : sizeof(__u64);
 	__u64 batch, total = 0;
-	__u32 count = AGG_BATCH_SIZE;
+	__u32 count = AGG_BATCH_SIZE, del_idx = 0;
 	bool remaining_entries = true;
 	int err = 0, i;
+	bool del_key;
 
 	DECLARE_LIBBPF_OPTS(bpf_map_batch_opts, batch_opts, .flags = BPF_EXIST);
 
 	values = calloc(n_cpus, sizeof(*values) * AGG_BATCH_SIZE);
 	keys = calloc(AGG_BATCH_SIZE, keysize);
-	if (!values || !keys) {
+	del_keys = calloc(MAP_AGGREGATION_SIZE, keysize);
+	if (!values || !keys || !del_keys) {
 		err = -ENOMEM;
 		goto exit;
 	}
@@ -1257,7 +1295,11 @@ static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
 			report_aggregated_rtt_mapentry(keys + i * keysize,
 						       values + i * n_cpus,
 						       n_cpus, af, prefix_len,
-						       t_monotonic, agg_conf);
+						       t_monotonic, agg_conf,
+						       &del_key);
+			if (del_key)
+				memcpy(del_keys + del_idx++ * keysize,
+				       keys + i * keysize, keysize);
 		}
 
 		// Update cleared stats
@@ -1270,9 +1312,15 @@ static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
 		count = AGG_BATCH_SIZE; // Ensure we always try to fetch full batch
 	}
 
+	// Delete keys outside of fetch/update loop to avoid messing with key order
+	if (del_idx > 0)
+		err = bpf_map_delete_batch(map_fd, del_keys, &del_idx,
+					   &batch_opts);
+
 exit:
 	free(values);
 	free(keys);
+	free(del_keys);
 	return err;
 }
 
@@ -1801,6 +1849,7 @@ int main(int argc, char *argv[])
 		.clean_args = { .cleanup_interval = 1 * NS_PER_SECOND,
 				.valid_thread = false },
 		.agg_conf = { .aggregation_interval = 1 * NS_PER_SECOND,
+			      .timeout_interval = 30 * NS_PER_SECOND,
 			      .ipv4_prefix_len = 24,
 			      .ipv6_prefix_len = 48,
 			      .n_bins = RTT_AGG_NR_BINS,
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -21,10 +21,15 @@ static const char *__doc__ =
 #include <pthread.h>
 #include <xdp/libxdp.h>
 #include <sys/signalfd.h>
+#include <sys/timerfd.h>
 #include <sys/epoll.h>
 
 #include "json_writer.h"
 #include "pping.h" //common structs for user-space and BPF parts
+
+// Maximum string length for IP prefix (including /xx[x] and '\0')
+#define INET_PREFIXSTRLEN (INET_ADDRSTRLEN + 3)
+#define INET6_PREFIXSTRLEN (INET6_ADDRSTRLEN + 4)
 
 #define PERF_BUFFER_PAGES 64 // Related to the perf-buffer size?
 
@@ -46,9 +51,12 @@ static const char *__doc__ =
 #define PPING_EPEVENT_TYPE_PERFBUF (1ULL << 63)
 #define PPING_EPEVENT_TYPE_SIGNAL (1ULL << 62)
 #define PPING_EPEVENT_TYPE_PIPE (1ULL << 61)
+#define PPING_EPEVENT_TYPE_AGGTIMER (1ULL << 60)
 #define PPING_EPEVENT_MASK                                                     \
 	(~(PPING_EPEVENT_TYPE_PERFBUF | PPING_EPEVENT_TYPE_SIGNAL |            \
-	   PPING_EPEVENT_TYPE_PIPE))
+	   PPING_EPEVENT_TYPE_PIPE | PPING_EPEVENT_TYPE_AGGTIMER))
+
+#define AGG_BATCH_SIZE 64 // Batch size for fetching aggregation maps (bpf_map_lookup_batch)
 
 /* Value that can be returned by functions to indicate the program should abort
  * Should ideally not collide with any error codes (including libbpf ones), but
@@ -88,12 +96,27 @@ struct map_cleanup_args {
 	bool valid_thread;
 };
 
+struct aggregation_config {
+	__u64 aggregation_interval;
+	__u64 n_bins;
+	__u64 bin_width;
+	__u8 ipv4_prefix_len;
+	__u8 ipv6_prefix_len;
+};
+
+struct aggregation_maps {
+	int map_v4_fd;
+	int map_v6_fd;
+};
+
 // Store configuration values in struct to easily pass around
 struct pping_config {
 	struct bpf_config bpf_config;
 	struct bpf_tc_opts tc_ingress_opts;
 	struct bpf_tc_opts tc_egress_opts;
 	struct map_cleanup_args clean_args;
+	struct aggregation_config agg_conf;
+	struct aggregation_maps agg_maps;
 	char *object_path;
 	char *ingress_prog;
 	char *egress_prog;
@@ -117,20 +140,23 @@ static json_writer_t *json_ctx = NULL;
 static void (*print_event_func)(const union pping_event *) = NULL;
 
 static const struct option long_options[] = {
-	{ "help",             no_argument,       NULL, 'h' },
-	{ "interface",        required_argument, NULL, 'i' }, // Name of interface to run on
-	{ "rate-limit",       required_argument, NULL, 'r' }, // Sampling rate-limit in ms
-	{ "rtt-rate",         required_argument, NULL, 'R' }, // Sampling rate in terms of flow-RTT (ex 1 sample per RTT-interval)
-	{ "rtt-type",         required_argument, NULL, 't' }, // What type of RTT the RTT-rate should be applied to ("min" or "smoothed"), only relevant if rtt-rate is provided
-	{ "force",            no_argument,       NULL, 'f' }, // Overwrite any existing XDP program on interface, remove qdisc on cleanup
-	{ "cleanup-interval", required_argument, NULL, 'c' }, // Map cleaning interval in s, 0 to disable
-	{ "format",           required_argument, NULL, 'F' }, // Which format to output in (standard/json/ppviz)
-	{ "ingress-hook",     required_argument, NULL, 'I' }, // Use tc or XDP as ingress hook
-	{ "xdp-mode",         required_argument, NULL, 'x' }, // Which xdp-mode to use (unspecified, native or generic)
-	{ "tcp",              no_argument,       NULL, 'T' }, // Calculate and report RTTs for TCP traffic (with TCP timestamps)
-	{ "icmp",             no_argument,       NULL, 'C' }, // Calculate and report RTTs for ICMP echo-reply traffic
-	{ "include-local",    no_argument,       NULL, 'l' }, // Also report "internal" RTTs
-	{ "include-SYN",      no_argument,       NULL, 's' }, // Include SYN-packets in tracking (may fill up flow state with half-open connections)
+	{ "help",                 no_argument,       NULL, 'h' },
+	{ "interface",            required_argument, NULL, 'i' }, // Name of interface to run on
+	{ "rate-limit",           required_argument, NULL, 'r' }, // Sampling rate-limit in ms
+	{ "rtt-rate",             required_argument, NULL, 'R' }, // Sampling rate in terms of flow-RTT (ex 1 sample per RTT-interval)
+	{ "rtt-type",             required_argument, NULL, 't' }, // What type of RTT the RTT-rate should be applied to ("min" or "smoothed"), only relevant if rtt-rate is provided
+	{ "force",                no_argument,       NULL, 'f' }, // Overwrite any existing XDP program on interface, remove qdisc on cleanup
+	{ "cleanup-interval",     required_argument, NULL, 'c' }, // Map cleaning interval in s, 0 to disable
+	{ "format",               required_argument, NULL, 'F' }, // Which format to output in (standard/json/ppviz)
+	{ "ingress-hook",         required_argument, NULL, 'I' }, // Use tc or XDP as ingress hook
+	{ "xdp-mode",             required_argument, NULL, 'x' }, // Which xdp-mode to use (unspecified, native or generic)
+	{ "tcp",                  no_argument,       NULL, 'T' }, // Calculate and report RTTs for TCP traffic (with TCP timestamps)
+	{ "icmp",                 no_argument,       NULL, 'C' }, // Calculate and report RTTs for ICMP echo-reply traffic
+	{ "include-local",        no_argument,       NULL, 'l' }, // Also report "internal" RTTs
+	{ "include-SYN",          no_argument,       NULL, 's' }, // Include SYN-packets in tracking (may fill up flow state with half-open connections)
+	{ "aggregate",            required_argument, NULL, 'a' }, // Aggregate RTTs every X seconds instead of reporting them individually
+        { "aggregate-subnets-v4", required_argument, NULL, '4' }, // Set the subnet size for IPv4 when aggregating (default 24)
+	{ "aggregate-subnets-v6", required_argument, NULL, '6' }, // Set the subnet size for IPv6 when aggregating (default 48)
 	{ 0, 0, NULL, 0 }
 };
 
@@ -197,19 +223,49 @@ static int parse_bounded_double(double *res, const char *str, double low,
 	return 0;
 }
 
+static int parse_bounded_long(long long *res, const char *str, long long low,
+			      long long high, const char *name)
+{
+	char *endptr;
+	errno = 0;
+
+	*res = strtoll(str, &endptr, 10);
+	if (endptr == str || strlen(str) != endptr - str) {
+		fprintf(stderr, "%s %s is not a valid integer\n", name, str);
+		return -EINVAL;
+	}
+
+	if (errno == ERANGE) {
+		fprintf(stderr, "%s %s overflowed\n", name, str);
+		return -ERANGE;
+	}
+
+	if (*res < low || *res > high) {
+		fprintf(stderr, "%s must be in range [%lld, %lld]\n", name, low,
+			high);
+		return -ERANGE;
+	}
+
+	return 0;
+}
+
 static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 {
 	int err, opt;
 	double user_float;
+	long long user_int;
 
 	config->ifindex = 0;
-	config->bpf_config.localfilt = true;
 	config->force = false;
+
+	config->bpf_config.localfilt = true;
 	config->bpf_config.track_tcp = false;
 	config->bpf_config.track_icmp = false;
 	config->bpf_config.skip_syn = true;
+	config->bpf_config.push_individual_events = true;
+	config->bpf_config.agg_rtts = false;
 
-	while ((opt = getopt_long(argc, argv, "hflTCsi:r:R:t:c:F:I:x:",
+	while ((opt = getopt_long(argc, argv, "hflTCsi:r:R:t:c:F:I:x:a:4:6:",
 				  long_options, NULL)) != -1) {
 		switch (opt) {
 		case 'i':
@@ -319,6 +375,33 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 				return -EINVAL;
 			}
 			break;
+		case 'a':
+			/* Aggregated output currently disables individual RTT */
+			config->bpf_config.push_individual_events = false;
+			config->bpf_config.agg_rtts = true;
+
+			err = parse_bounded_long(&user_int, optarg, 1,
+						 7 * S_PER_DAY, "aggregate");
+			if (err)
+				return -EINVAL;
+
+			config->agg_conf.aggregation_interval =
+				user_int * NS_PER_SECOND;
+			break;
+		case '4':
+			err = parse_bounded_long(&user_int, optarg, 0, 32,
+						 "aggregate-subnets-v4");
+			if (err)
+				return -EINVAL;
+			config->agg_conf.ipv4_prefix_len = user_int;
+			break;
+		case '6':
+			err = parse_bounded_long(&user_int, optarg, 0, 64,
+						 "aggregate-subnets-v6");
+			if (err)
+				return -EINVAL;
+			config->agg_conf.ipv6_prefix_len = user_int;
+			break;
 		case 'h':
 			printf("HELP:\n");
 			print_usage(argv);
@@ -334,6 +417,12 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 			"An interface (-i or --interface) must be provided\n");
 		return -EINVAL;
 	}
+
+	config->bpf_config.ipv4_prefix_mask =
+		htonl(0xffffffffUL << (32 - config->agg_conf.ipv4_prefix_len));
+	config->bpf_config.ipv6_prefix_mask =
+		htobe64(0xffffffffffffffffUL
+			<< (64 - config->agg_conf.ipv6_prefix_len));
 
 	return 0;
 }
@@ -649,6 +738,35 @@ static __u64 convert_monotonic_to_realtime(__u64 monotonic_time)
 	return monotonic_time + offset;
 }
 
+// Stolen from xdp-tool/lib/util/util.c
+int try_snprintf(char *buf, size_t buf_len, const char *format, ...)
+{
+	va_list args;
+	int len;
+
+	va_start(args, format);
+	len = vsnprintf(buf, buf_len, format, args);
+	va_end(args);
+
+	if (len < 0)
+		return -EINVAL;
+	else if ((size_t)len >= buf_len)
+		return -ENAMETOOLONG;
+
+	return 0;
+}
+
+/*
+ * Is the passed ip an IPv4 address mapped into the IPv6 space as specified by
+ * RFC 4291 sec 2.5.5.2?
+ */
+static bool is_ipv4_in_ipv6(const struct in6_addr *ip)
+{
+	__u16 ipv4_prefix[] = { 0x0, 0x0, 0x0, 0x0, 0x0, 0xFFFF };
+
+	return memcmp(ipv4_prefix, ip, sizeof(ipv4_prefix)) == 0;
+}
+
 /*
  * Wrapper around inet_ntop designed to handle the "bug" that mapped IPv4
  * addresses are formated as IPv6 addresses for AF_INET6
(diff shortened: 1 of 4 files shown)


pping/pping.c
```diff
@@ -942,7 +994,13 @@ static int get_aggregated_rtts(int map_fd,
 			rtt_stats->bins[j] += per_cpu_rtt_stats[i].bins[j];
 	}
 
-exit:
+exit_after_switch:
+	// Clear RTT stats
+	memset(per_cpu_rtt_stats, 0, sizeof(*per_cpu_rtt_stats) * nr_cpus);
+	err2 = bpf_map_update_elem(map_fd, &key, per_cpu_rtt_stats, BPF_EXIST);
+	err = err ? err : err2;
```
tohojo, line 1001: If err is non-zero at this point that means that the lookup failed; does it really make sense to do the update in this case (and this dance of `err` and `err2`)? That doesn't seem like it's something we can actually reasonably expect to recover from, so maybe it's better to just error out and exit completely if this happens?

pping/pping.c
```diff
@@ -1149,6 +1208,15 @@ int setup_periodical_aggregate_rtts(struct bpf_object *obj,
...
 
+	fd = bpf_object__find_map_fd_by_name(obj, config->agg_map_idx);
+	if (fd < 0) {
+		fprintf(stderr, "Unable to find aggregation idx map %s: %s\n",
+			config->agg_map_idx, get_libbpf_strerror(fd));
+		return fd;
+	}
+	config->agg_args.map_idx_fd = fd;
+
+
```
tohojo, line 1219: Duplicated empty line

pping/pping.c
```diff
@@ -833,6 +886,95 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
...
+		err = get_aggregated_rtts(argp->map_fd, &rtt_stats);
+		if (err) {
+			libbpf_strerror(err, buf, sizeof(buf));
+			fprintf(stderr, "Failed aggregating per-CPU RTTs: %s\n",
+				buf);
+		} else {
+			print_aggregated_rtts(stdout, &rtt_stats);
+		}
+
+		nanosleep(&interval, NULL);
```
tohojo, line 1031: Will this get interrupted by `pthread_cancel()`? Because otherwise this can lead to fairly long exit times if the interval is long...

pping/pping.c
```diff
@@ -163,15 +176,17 @@ static int parse_bounded_double(double *res, const char *str, double low,
...
 	config->ifindex = 0;
 	config->bpf_config.localfilt = true;
 	config->force = false;
 	config->bpf_config.track_tcp = false;
 	config->bpf_config.track_icmp = false;
+	config->bpf_config.push_individual_events = true;
+	config->bpf_config.agg_rtts = false;
 
-	while ((opt = getopt_long(argc, argv, "hflTCi:r:R:t:c:F:I:",
+	while ((opt = getopt_long(argc, argv, "hflTCi:r:R:t:c:F:I:a::",
```
tohojo, line 191: Two colons?

pping/pping.c
```diff
@@ -833,6 +886,95 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
...
 
+static void print_histogram(FILE *stream,
+			    struct aggregated_rtt_stats *rtt_stats)
+{
+	int i;
+
+	fprintf(stream, "[%llu", rtt_stats->bins[0]);
+	for (i = 1; i < RTT_AGG_NR_BINS; i++)
+		fprintf(stream, ",%llu", rtt_stats->bins[i]);
+	fprintf(stream, "]");
```
tohojo, line 899: So this format will be mostly useful for aggregating later, right? Maybe we could do some kind of fancy ascii-art lines for each bin instead so this works better for an interactive display?

tohojo (changes requested):
I think the approach looks reasonable. Only thing (other than these minor-ish nits) that struck me was that it was maybe a bit overkill to have a separate per-CPU map to store the current index: a single global variable could be enough - I *think* the syscall enter-exit when looking up the values should serve as a memory barrier so that you can be reasonably sure that every CPU has seen the update if you just flip a single global var...

And yeah, as you note the output format probably needs a bit of work (which I also note below) :)

If you want to make the size of the histogram configurable, you could just make each bin a separate map index and just resize the map (instead of having a single map value with a huge array); you'd have to have two maps to swap between, but that can just be open-coded (you don't even need to use a map-in-map).

pping/pping.c
```diff
@@ -911,12 +913,57 @@ static void print_aggregated_rtts(FILE *stream,
...
+static int switch_agg_map(int map_idx_fd)
+{
+	int nr_cpus = libbpf_num_possible_cpus();
+	__u32 active_idx, next_idx, key = 0;
+	int i, err;
+
+	/* While map value is __u32, map lookup/update will 8-byte align each
+	 * CPU entry. Pretending each value is __u64 is easier than having to do
+	 * a bunch of pointer casting to align __u32s correctly*/
+	__u64 *percpu_map_idx;
```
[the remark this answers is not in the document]
simosund: Yhea that probably makes more sense... Will change.

pping/pping.c
simosund: My thought was if it for some reason was an intermittent failure than we could just lose one report and then go on as normal. But yhea, have no idea why the map update would fail and chances are if it fails it will continue to fail, so maybe completely stopping makes more sense.

pping/pping.c
simosund: Nanosleep is on the list of functions that are required to be cancellation points, so yes it should be interrupted (and at least on my system it is, i.e. I can set period to be an hour but it will cancel pretty much instantly on Ctrl-C)

pping/pping.c
simosund: Two colons is for optional arguments (single for required), although not a huge fan of how getopt_long() handles optional arguments.

pping/pping.c
simosund: Yhea, right now it's just something hacked together that I figured wouldn't be too annoying to parse. I guess a visually more useful representation could be better for the standard format, and leaving dumping of raw bin-counts to ex. the JSON format.

simosund:
> I think the approach looks reasonable. Only thing (other than these minor-ish nits) that struck me was that it was maybe a bit overkill to have a separate per-CPU map to store the current index: a single global variable could be enough - I _think_ the syscall enter-exit when looking up the values should serve as a memory barrier so that you can be reasonably sure that every CPU has seen the update if you just flip a single global var...

Ok. Any how is it with contention when multiple CPUs try to access the same global var, does that risk hurting performance (although I guess I already have a couple of other global vars). Also not entirely sure how to update a single global var from userspace, but guess I can look that up.
 
> And yeah, as you note the output format probably needs a bit of work (which I also note below) :)

Yhea, output format is not really something I've thought too much about right now. Just wanted to get something together that works well enough that I can test it, and preferably not be too hard to parse so I can hack together some script to parse and process the data.

> If you want to make the size of the histogram configurable, you could just make each bin a separate map index and just resize the map (instead of having a single map value with a huge array); you'd have to have two maps to swap between, but that can just be open-coded (you don't even need to use a map-in-map).

Hmm, good idea, I will consider that, but might wait a bit with that for now and focus on getting some initial results to see how it performs.

pping/pping.c
tohojo: Ah, right; never used those I guess :)

Why not just make it mandatory to supply an interval, then?

simosund: I thought it would be nice to be able to just specify the that you wanted the aggregated format and then have a default interval. But considering the somewhat odd requirements on the format for those optional arguments it may overall be less confusing to just have it be required instead.

pping/pping_kern.c
```diff
@@ -149,9 +149,16 @@ struct {
...
-	__uint(max_entries, 1);
+	__uint(max_entries, 2);
 } agg_rtt_stat SEC(".maps");
 
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, __u32);
+	__type(value, __u64);
+	__uint(max_entries, 1);
+} agg_rtt_stat_idx SEC(".maps");
```
netoptimizer, line 160: Why is this "agg_rtt_stat_idx" map using a PERCPU variant? (`BPF_MAP_TYPE_PERCPU_ARRAY`)
I think a normal `BPF_MAP_TYPE_ARRAY` would be sufficient.

Plus, this could be defined as a global variable (that get turned into a array map behind the scenes).

simosund: Well, by using a PERCPU map it allows me to unnecessarily complicate my user space code so I can inflate the lines of code I've written! On a more serious note, you are completely right that a `BPF_MAP_TYPE_ARRAY` should be sufficient (as the BPF progs only read the map so should not have any concurrency issues), so will switch to that.

While using a global variable for `agg_rtt_stat_idx` (like Toke also suggested) seems like a good idea, I've simply been unable to to figure out how to only update a single global variable from user space in an atomic manner. All (non-constant) global variables seem to go into a single value in the `(pping_ke.)bss` map as far as I can tell. So reading that one value in the bss map from user space would get me all global variables. While I assume I could use some sort of BTF information to find the right offset for the `agg_rtt_stat_idx` variable and only update that, I assume that once I update the bss map from user space it would go ahead and write over all the global variables with whatever memory blob I have in user space. As the BPF programs write to some of the global variables I assume I could get some lost updates then if I first read all the global variables from user space, one of the BPF progs updates one of the global variables, and I then update the global variables from user space where it left all global variables except the `agg_rtt_stat_idx` the same as when it first read the map. I guess there might be some much better way to update global variables from user space, so if you have some pointers on how to do it I would happily try that.

tohojo: To only update a single element, you need to `mmap()` the contents of the map; then you can just write to the right offset and that global memory will be updated.

The `bpftool skeleton` code does this for you, providing handy member accesses. The simplest way to do it without the skeleton code is probably do define all the global variables inside a containing struct so you can just cast the mmap'ed area to that and update individual members...

Note, however, that if you're writing to global variables in the BPF program, any updates you do from userspace *will* cause contention if they share the same cache line (within 64 bytes of each other), even if it's not the same variable...

simosund: So unless I switch to using skeletons (which maybe is something I should do but seems a bit out of scope for this PR), I will have to place all global variables in a struct which user space has to know about, make sure to pad out the map index member from the other existing global variables to avoid them being on the same cache line and then somehow memory-map the contents of the bss map (I assumed that the kernel would generally not want to give user space programs direct access to map memory). 

To me this seems complicated than just having the map index in its own map as it is now, or? Is there any performance benefit of having it as a global variable (which turns into a map behind the scenes) than directly putting it in a map, or are there any other advantages to having it as a global variable vs as a map?

tohojo: Yeah, that's more or less accurate. I don't think there will be a measurable performance difference between just having a separate single-element array map, and using a global variable. The verifier will optimise the global variable access, but it also optimises array map lookups, so it'll probably be roughly equivalent :)

simosund:
Have just rebased it and done some minor changes based on some of your previous comments, but not changed to LPM map yet so no point in re-reviewing it yet. There are however a couple of things I want to check so I can maybe avoid a couple of mistakes. 

The first is just regarding the comments on using a global variable instead of a separate map to tell the BPF programs which is the active map to currently aggregate the RTTs in. Based on some discussion above my conclusion was that in the end using a global variable would just be unnecessarily complicated (due to it being somewhat complicated to only update a single variable from userspace) for no considerable performance gain, so keeping it as a separate map (but not per-CPU anymore) would be ok.

The second one was regarding how to swap between the active map (which BPF programs are currently aggregating RTTs in) and the stale map that the user space will read, print out and reset. While simply using two different indices in an array map was quite convenient for me first implementation, that won't really work when I switch to an LPM map. @tohojo suggested that I could just use two maps and have it "open-coded" (rather than using a map-in-map), but don't really understand what "open-coded" means. Did it mean manually creating two separate maps and basically do an if-else choice between the two maps directly in the code, similar to how it is in my updated code in commit [#bab4ad9b5b9 pping: Reset aggregated RTTs after each report](https://github.com/xdp-project/bpf-examples/commit/bab4ad9b5b941630466b4d7e87bcabe56dff094e).

simosund:
My plan was first to simply add the change to an LPM map as an additional commit. However, as far as I can tell there is no per-CPU version of the LPM map, so will have to redo a fair bit of the logic anyways, so figured it's probably better to just start of with an LPM map from the first commit in this PR.

I was kinda glad that I for once could use per-CPU maps so I didn't have to worry about any concurrency issues, but with an LPM map I guess I will have to rely on `__sync_fetch_and_add()` to avoid any lost updates when incrementing the bin-counts. Hopefully there will not be too much contention to the same bin of the same IP-prefix at the same time.  

To keep track of min and max accurately I guess I would have to use a `bpf_spin_lock`, locking the aggregation state for the entire IP-prefix during that part, which feels less than ideal. Alternatively I guess I could just ignore concurrency control for the min/max checks to avoid some locking, but at the risk of the min/max not being entirely accurate (which has been my general approach for everything else so far, but then it's been per-flow and not per IP-prefix).

simosund:
So, tried to add a `bpf_spin_lock` to the LPM map to guard against concurrent updates of the aggregated state (as there's no per-CPU version of the LPM map). But I had no luck (see the "WIP" commit). It seems like it might not be possible to use spinlocks in LPM maps? In the man-pages it says that spinlocks can only be used in `BPF_MAP_TYPE_ARRAY` and `BPF_MAP_TYPE_HASH` maps. Is this still the case, and do you know if it might change soon?

Otherwise I guess I'll have to just rely on `__synch_fetch_and_add()` for the bin counts, and no guards at all for the min/max.

tohojo:

> So, tried to add a `bpf_spin_lock` to the LPM map to guard against
> concurrent updates of the aggregated state (as there's no per-CPU
> version of the LPM map). But I had no luck (see the "WIP" commit). It
> seems like it might not be possible to use spinlocks in LPM maps? In
> the man-pages it says that spinlocks can only be used in
> `BPF_MAP_TYPE_ARRAY` and `BPF_MAP_TYPE_HASH` maps. Is this still the
> case, and do you know if it might change soon?

Hmm, good question. The number of map types supporting this have already
been expanded, and I don't think there's anything that fundamentally
prevents supporting spin locks in LPM maps as well. I can put it on my
list of things to implement when I have some spare cycles.

However, even if that does land in a future kernel version, relying on
it for pping is probably not a good idea; so I guess you'll just have to
use the atomic ops instead for now at least...

tohojo:

> What do you think? Can keeping histograms for the IP-prefixes actually
> used in a separate hashmap be a way to salvage this type of
> aggregation, or is histogram per IP-prefix just a poor strategy to
> begin with?

I think this sounds like an excellent idea, yeah :)

You could keep a "last used" timestamp in each hash map bucket along
with the histogram, and have the logic that walks it remove entries
entirely after some timeout, so busy buckets don't have to be re-created
after every poll...

simosund:
> I think this sounds like an excellent idea, yeah :)

What I dislike about the idea is that to some degree I'm trading high memory usage and periodically high user space cost against higher BPF/per packet processing, which generally seems like the opposite of what I want to be doing. In the case where we only have a few IP-prefixes (or even just a single global one), having the histograms in a separate hashmap will just be additional overhead without much gain. 

But the current implementation (keeping the histograms in the LPM map) results in the userspace having to pull several GBs of (mostly empty) data, a process that can take several seconds, is only remotely feasible if you can basically dedicate a separate core for the user space process and don't need to report more frequently than every 10 seconds or so. So some change is needed I guess, and unless anyone can come up with a better idea dumping histograms for the actively used IP-prefixes in a separate hashmap is the the best alternative I can come up with right now.

One potential advantage of using hashmaps for the histograms is that I have some better options for dealing with concurrency. I could for example switch out the atomic add for histogram bin count increment to a spinlock that also guards the min/max updating. Having a spinlock basically guarding the entire aggregation data structure could cause a lot of contention though (ex. all cores handling traffic to the same IP-prefix), which is only a problem for the current atomic add in case all cores are trying to increment the same histogram bin. I'm more leaning towards using per-CPU maps instead,  although having a separate copy of the histogram per core could lead to the same issue I'm having now. In case we have 10k active prefixes and running on a 128-core CPU, then I suddenly have 1.28m histograms to fetch, leaving me back at square one... 

Guess an alternative to having a "full" per-CPU map (with a separate copy per core) would be to have a smaller number of duplicates to chose between, ex. keeping 8 copies and then doing core-id % 8 (or based on timestamp or really any semi-random aspect to distribute the accesses). That wouldn't fully protect against concurrency, just make it less likely, but could perhaps be some compromise between concurrency protection and the potentially high memory usage for a full per-CPU map. But maybe I'm getting a bit too far ahead of myself, as this is something we can worry about later. Just starting to become a bit less keen on spending a bunch of time implementing things that don't scale one way or another... It's a great learning process, but a slow way of getting any results.

> You could keep a "last used" timestamp in each hash map bucket along with the histogram, and have the logic that walks it remove entries entirely after some timeout, so busy buckets don't have to be re-created after every poll...

That's a good idea! Without any real data to back it up, my feeling is that creating a new hashmap entry is a lot slower than simply looking up an existing entry. Unfortunately this logic will in some sense be a bit out of date, as the user space logic will operate on the "inactive" map, which the BPF progs will only start using again as the "active" and "inactive" maps are swapped the next report period. But unless the traffic distribution is very volatile, i.e. there's a very small overlap between the set of IP-prefixes that traffic actually hit during consecutive report periods, I guess this could still help a bit.

But before I start implementing this whole idea of keeping the histograms for active IP-prefixes in a separate hashmap, I will also do some testing to see how the LPM map lookup speed is affected by having a 1m+ entries. As if the performance from the LPM-lookup itself becomes very poor with a large number of IP-prefixes, then this new implementation will not really help (as it will still need to keep all the IP-prefixes in the LPM map and do LPM-map lookups to get the correct prefix). So will probably not start working on it until after my next supervisor meeting (2023-03-06), where I can maybe get some input from Anna and Simone as well if they think this is a good solution, or if we should maybe do the aggregation in a completely different way.

tohojo:

>> I think this sounds like an excellent idea, yeah :)
>
> What I dislike about the idea is that to some degree I'm trading high
> memory usage and periodically high user space cost against higher
> BPF/per packet processing, which generally seems like the opposite of
> what I want to be doing. In the case where we only have a few
> IP-prefixes (or even just a single global one), having the histograms
> in a separate hashmap will just be additional overhead without much
> gain.

Yes, if only we had some way to dynamically change the code we're
running based on some parameter... ;)

> But maybe I'm getting a bit too far ahead of myself, as this is
> something we can worry about later. Just starting to become a bit less
> keen on spending a bunch of time implementing things that don't scale
> one way or another... It's a great learning process, but a slow way of
> getting any results.

Yes, welcome to the world of performance engineering. Lots of tradeoffs,
slow progress, and all but impossible to do well without some actual
data to base your optimisation decisions on. This is basically the
origin of that "premature optimisation is evil" quote you've probably
heard :)

Also, don't view all this as wasted effort: you're learning and
improving things, and describing these problems and how you solve them
can absolutely be part of your next paper!


>> You could keep a "last used" timestamp in each hash map bucket along
>> with the histogram, and have the logic that walks it remove entries
>> entirely after some timeout, so busy buckets don't have to be
>> re-created after every poll...
>
> That's a good idea! Without any real data to back it up, my feeling is
> that creating a new hashmap entry is a lot slower than simply looking
> up an existing entry. Unfortunately this logic will in some sense be a
> bit out of date, as the user space logic will operate on the
> "inactive" map, which the BPF progs will only start using again as the
> "active" and "inactive" maps are swapped the next report period. But
> unless the traffic distribution is very volatile, i.e. there's a very
> small overlap between the set of IP-prefixes that traffic actually hit
> during consecutive report periods, I guess this could still help a
> bit.

Yeah, it'll operate on the "wrong" map, but you probably want the
timeout to cover a couple of poll periods anyway to avoid churn, so
that's not too big of an issue. It may be worth benchmarking different
options here as well, though; hashmaps do have a built-in memory cache
to speed up allocation, and there's the option of using preallocated
maps as well...

simosund:
> Yes, if only we had some way to dynamically change the code we're running based on some parameter... ;)

So now you expect me to support multiple different implementations without making a mess of the code? You're the one who's gonna have to review it, so I guess you'll see how that goes =). But on the topic of reviewing, in case it wasn't clear from having this conversation, you can probably delay reviewing this PR a bit, as it will likely still have to undergo quite a few changes one way or another.

>  It may be worth benchmarking different options here as well, though; hashmaps do have a built-in memory cache to speed up allocation, and there's the option of using preallocated maps as well...

With preallocated maps you simply mean that the memory for all buckets are preallocated? I.e. not passing the `BPF_F_NO_PREALLOC` flag. But even then I guess there will be some more overhead with actually creating a new entry in the BPF prog as then it has to copy the initial value from the stack to the map (instead of just giving me a pointer to an existing map value), and I suspect there's also some more locking going on to handle concurrency.

Furthermore, at least the way I do "create a new entry" for the flow state now actually means that creating an entry consists of two lookup calls and one update call. First a lookup to check if the key already exists, then an update call to actually create the entry, and finally another lookup call to actually get a pointer to the map value in the map (rather than on the stack). I guess that last lookup could be avoided by simply filling up the initial value on the local stack with all relevant data before doing the update call, but that complicates the code slightly (need to delay update call until the code to fill in the entry has finished) and may risk doing unnecessary work in case the update call fails (due to the map already being full).

So definitely think there's a point in avoiding some unnecessary churn by keeping around the entries for frequently accessed IP-prefixes (just zeroing their aggregation stats). But as you say, we probably need to run some benchmarks to really know.

simosund:
Hi @tohojo,
I've done an initial rebase to change my aggregation implementation to dynamically create IP-prefixes (stored in a hashmap) of a pre-configured size based on the traffic instead of doing an LPM lookup on a pre-configured set of IP-prefixes. It's not finished yet (among other things it doesn't expire any IP-prefixes right now), so it's not ready for a full review. I have however run into a bit of a problem that I could use some help with.

Basically, I can't manage to just copy the user-configured size of IP-prefix into my key for the aggregation, see my review above for more details. I've also tried to copy the full IP-address and then just zero out the last bytes with memset (instead of initializing the IP to 0 and then copy the first bytes), but that gives similar issues but with memset instead of memcpy. Doing it with a fixed constant, ex. 15 bytes (corresponding to /24 of an IPv4 address (mapped in IPv6 space)) works fine, but the problem seems to occur when I try to do it with a (at compile time) unknown size. 

Any ideas for how to solve this?

simosund:
I'll also mention that I've kept a reference to my LPM-based implementation (i.e. this branch before this rebase) at [pping-agg-prototype-LPM](https://github.com/simosund/bpf-examples/tree/pping-agg-prototype-LPM), so that it's not completely lost.

tohojo:
Instead of limiting the copied bytes, I think I would implement this as masking instead; since you support arbitrary bit lengths, you're not going to hit full byte boundaries anyway. So something like:

```c
/* config var */
static const volatile in6_addr v6_aggregate_mask = { 0xff, 0xff, 0xff, 0xf0, 0x0, 0x0, 0x0, 0x0 }; /* pxlen 60 */
/* ... */

int lookup_aggregate_bin(const struct in6_addr *addr) {
  in6_addr key = *addr;
  __u64 *mask = (__u64 *)&v6_aggregate_mask, *key_ptr = (__u64 *)&key;

  *key_ptr &= *mask;
  *(key_ptr+1) &= *(mask+1);

  bpf_map_lookup(map, &key /*etc*/);
}
```

The casting to `__u64` pointers is a bit ugly, but it's efficient! :)

simosund:
> Instead of limiting the copied bytes, I think I would implement this as masking instead; since you support arbitrary bit lengths, you're not going to hit full byte boundaries anyway. So something like:

Hmm, good idea. Was planning on doing masking for the last byte to get the last few bits (as the current memcmp-like approach only supports whole bytes), but as I couldn't even get the whole-byte part working I left that for later. But your approach utalizing that we already know the key size (128 bits) and just using a pre-calculated mask seems like it should be more efficient (and be easier for the verifier to verify). So will try it out (even if I'm still slightly confused why the `my_memcpy()` approach doesn't compile due to somehow being an invalid call to the built-in memcpy).

> The casting to `__u64` pointers is a bit ugly, but it's efficient! :)

Not that I'm gonna pretend to have a whole lot of experience with it, but seems like there's often a bit of a tradeoff between efficient and "pretty" code. But I suppose performance actually matters when the code will potentially run millions of times every second =). So doing two 64-bit instead of up to 16 8-bit operations in a loop seems like a good idea.

simosund:
Small status update. Have pushed a new version using separate aggregation maps for IPv4 and IPv6 (with just 32 bits for IPv4 and the first 64 bits for IPv6). I've kept a reference to the previous version [here](https://github.com/simosund/bpf-examples/tree/pping-agg-prototype-take2-HashmapAllIPv6) , which used a single map with a full IPv6-address (`struct in6_addr`) as key, where IPv4 addresses were simply mapped into IPv6 space.

This is still not ready for review though, as it in addition to a bit of cleanup (ex. the update of the README is outdated and refers to the LPM implementation) I also need to add:
- User space deleting stale IP-prefixes
- A /0 prefix for traffic that could not be given their own IP-prefix due to prefix map being full
- Add packet and byte counts of for RX/TX of each prefix
- Add a bunch of counters for missed traffic (non-TCP traffic, TCP traffic without TSval etc.)

simosund:
Hi @tohojo, I think this is ready to be reviewed now.

The main features of this PR are:
- Add the in-kernel aggregation of RTT-stats into linear histograms on a per IP-prefix basis. Dynamically created the entries in the BPF programs based on the traffic. Prefix-size is configurable, and supports values in the range [0, 32] for IPv4 and [0, 64] for IPv6 (limited to 64 to limit the keysize). Use per-CPU hashmaps to avoid lost updates or other "fun" concurrency issues, as unlike individual flows it's not unlikely that multiple packets from the same IP-prefix may be processed concurrently.
- Add periodic user space reporting of the aggregated stats. To avoid the user space process attempting to read and reset values while they're concurrently being accessed by the BPF programs, use two sets of maps (creatively suffixed with `1` and `2`). When user space is about to report the stats, switch which set of maps the BPF programs actively use, and then fetch and reset the now inactive set of maps. Also delete IP-prefixes which haven't observed any traffic for a while (default 30s). 
- Output the aggregated statistics in either `standard` or `json` format. The `standard` format just prints some statistics calculated from the histograms, while the `json` format additionally dumps the histograms for later post-processing.
- Add support for writing the output directly to file instead of to stdout. Additionally, add options to allow ePPing to "rotate" the files in a similar manner as the `tcpdump` `-G`/`-C` and `-W` options. This works both for both the aggregated output and the previous individual RTT output.

I haven't added counters for the non-tracked traffic yet, but figured that's not really inherently related to adding an "aggregation option", so can do that in a separate PR (this one is already quite big as it is). Being able to write the output to (rotating) files is not really directly tied to the aggregation either, so can move that to a separate PR if you prefer (although that PR would still be based on top of this, as it assumes the aggregated output mode exists).

pping/pping.c
```diff
@@ -77,6 +77,7 @@ struct aggregation_config {
 	__u64 bin_width;
 	__u8 ipv4_prefix_len;
 	__u8 ipv6_prefix_len;
+	enum PPING_OUTPUT_FORMAT format;
```
tohojo: Side note: Why is the name of the enum upper-cased? That looks really weird, we should fix that (the *values* can be upper-case, but the enum name should be lower-case like all other type names)

pping/pping.c
```diff
@@ -362,6 +364,15 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 		case 'e':
 			config->bpf_config.agg_by_dst = true;
 			break;
+		case 'o':
+			err = parse_bounded_double(&user_val, optarg, 0,
+						   7 * S_PER_DAY,
+						   "aggregate-timeout");
```
tohojo, line 392: Same comment wrt just making this an integer number of seconds...

pping/pping.c
```diff
@@ -1147,19 +1179,31 @@ static int report_aggregated_rtt_map(int map_fd, int af, __u8 prefix_len,
...
 				goto exit;
 		}
 
+		if (agg_conf->timeout_interval > 0 &&
+		    merged_stats.last_updated < t_monotonic &&
+		    t_monotonic - merged_stats.last_updated >
+			    agg_conf->timeout_interval) {
+			del_prev = true;
+			del_key = next_key;
+		}
```
tohojo, line 1345: Why the indirection of the `del_*` variables, can't you just call `bpf_map_delete_elem()` here directly? You could even amend the check above so you avoid doing update-then-delete for expired entries?

pping/pping_kern.c
```diff
@@ -180,6 +181,34 @@ struct {
...
+	__uint(max_entries, 1);
+} map_v6_agg_backup1 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, __u32);
+	__type(value, struct aggregated_rtt_stats);
+	__uint(max_entries, 1);
+} map_v6_agg_backup2 SEC(".maps");
+
```
tohojo, line 215: Seems a bit much to create four extra maps just for this? Why not just use a special key in the existing maps for this purpose? All-ones, say. Userspace can just pre-create that entry before loading the BPF program (to ensure it's always available)...

(Side note related to this: you mentioned in one of the commit descriptions that the aggregation maps may baloon up to 32G of memory if there are a lot of prefixes and CPUs in the system; probably a good idea to protect against this, say by setting reasonable cgroup limits in the systemd unit file when running in production - assuming the kernel is new enough to have moved to cgroup-based memory accounting for BPF, that is...)

pping/pping.c
```diff
@@ -1299,6 +1303,15 @@ static void *periodic_rtt_aggregation(void *args)
...
 
+		t = get_time_ns(CLOCK_MONOTONIC) - t;
+#ifdef DEBUG
+		fprintf(stderr, "Time for reporting all entries: %.6g ms\n",
+			(double)t / NS_PER_MS);
+#endif
+		set_timespec(&interval,
+			     t < argp->agg_conf->aggregation_interval ?
+				     argp->agg_conf->aggregation_interval - t :
+				     0);
```
tohojo, line 1441: Instead of manually calculating and compensating for this drift, how about blocking on a timerfd instead of using nanosleep? See `timerfd_create()`...

pping/pping.c
```diff
@@ -150,6 +151,7 @@ static const struct option long_options[] = {
 	{ "aggregate-subnets-v6", required_argument, NULL, '6' }, // Set the subnet size for IPv6 when aggregating (default 48)
 	{ "aggregate-reverse",    no_argument,       NULL, 'e' }, // Aggregate RTTs by dst IP of reply packet (instead of src like default)
 	{ "aggregate-timeout",    required_argument, NULL, 'o' }, // Interval for timing out subnet entries in seconds (default 30s)
+	{ "truncate-histograms",  no_argument,       NULL, 'u' }, // Truncate trailing zeros from JSON histograms
```
tohojo, line 167: Does this really need to be an option? You're already running out of alphabet, and every option just makes things more complicated for the user. It this optimisation is worth doing I think we should just define the output format such that histograms are always truncated...

pping/pping.c
```diff
@@ -386,6 +389,17 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
...
 			break;
+		case 'w':
+			if (strlen(optarg) >
+			    sizeof(config->out_ctx.output_file) - 1) {
+				fprintf(stderr, "File name too long\n");
+				return -EINVAL;
+			}
+
+			strncpy(config->out_ctx.output_file, optarg,
+				sizeof(config->out_ctx.output_file));
```
tohojo, line 409: A slight optimisation here (which obviously doesn't matter in practice, but figured I'd mention it anyway): you're already doing an `strlen()` above, so you could just save the value of that and make the copy a more efficient `memcpy()` instead of looping through the string again here...

pping/pping.c
```diff
@@ -400,6 +407,15 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
...
 			break;
+		case 'G':
+			err = parse_bounded_double(&user_val, optarg, 1e-9,
+						   7 * S_PER_DAY,
+						   "rotate-seconds");
+			if (err)
+				return err;
+			config->out_ctx.rotate_period =
+				user_val * NS_PER_SECOND;
+			break;
```
tohojo, line 420: Same comment with intervals and doubles. Also should probably check somewhere that an output file has actually been set along with this; and that the output file name is not too long now that we're appending stuff to it...

pping/pping.c
```diff
@@ -1664,13 +1689,25 @@ int setup_periodical_aggregate_rtts(struct bpf_object *obj,
 	return 0;
 }
 
+static void gen_filename(struct output_context *out_ctx, char *buf, size_t size)
+{
+	if (out_ctx->rotate_period == 0) {
+		strncpy(buf, out_ctx->output_file, size);
+	} else {
+		snprintf(buf, size, "%s%u", out_ctx->output_file,
+			 out_ctx->rotate_counter);
```
tohojo, line 1707: Should this have a period separator before the number?

pping/pping.c
```diff
@@ -1664,13 +1689,25 @@ int setup_periodical_aggregate_rtts(struct bpf_object *obj,
...
+	}
+}
+
 static int open_output(struct output_context *out_ctx, bool print_aggconf,
 		       struct aggregation_config *agg_conf)
 {
+	char filename[sizeof(out_ctx->output_file) + 32];
```
tohojo, line 1714: Since you're using `MAX_PATH_LENGTH` for the size of `out_ctx->output_file` I don't think this will work - you can't actually create filenames that are longer, so you need to take the option into account when checking the input length, cf the comment above...

pping/pping.c
```diff
@@ -166,6 +167,7 @@ static const struct option long_options[] = {
 	{ "truncate-histograms",  no_argument,       NULL, 'u' }, // Truncate trailing zeros from JSON histograms
 	{ "write",                required_argument, NULL, 'w' }, // Write output to file (instead of stdout)
 	{ "rotate-seconds",       required_argument, NULL, 'G' }, // Create a new output file every X seconds, with a number suffixed to the file name (similar to tcmpdump -G, but with file naming of tcpdump -C)
+	{ "wrap-rotation",        required_argument, NULL, 'W' }, // In conjunction with -G, set max files that can be created (will then start overwriting previous files, similar to tcpdump -W)
```
tohojo, line 170: I dunno, is this option really needed? Seems a bit premature to add this until we've demonstrated a real use case for it. Also, if you *are* going to overwrite existing files, I'm not sure that doing so by truncating them in place is a good idea? In fact, the unconditional "open with truncation" mode seems like a footgun in any case; I suspect most people will expect pping to just append to the file; with the current behaviour, restarting epping with identical arguments will lead to data loss...

Also, there's no way to make epping resume with the same sequence number it left off at on the previous run. The more I think about it, the less I like the rotation idea as implemented. I think it's better to just keep the "write to file" option and just implement reopening of the output file on some appropriate signal. That way, rotation can be handled by logrotate instead of pping having to reimplement all this logic...

tohojo (changes requested):
Okay, so did end up with 20 comments, but that's still less that one per commit, so not so bad! And some of them are asking you to remove code, so that should be quick, right? ;)

simosund (commented):
So apparently GitHub no longer actually publish my replies to your comments that I make directly in the conversation view. Didn't realize I now have to go the the files-changed tab to submit a review for those comments to show up for anyone else... So anyways, here are (hopefully) the replies to your comments that I wrote at the start of the week...

pping/pping.c
tohojo: It's fine to just include a fixup commit at the beginning of the PR :)

pping/pping.c
tohojo: Oh, right, of course, silly me. Hmm, but are you sure that deleting a key can't mess up the ordering of later elements as well? May be better to do a separate delete pass later as you say. And yeah, if it's not too annoying to do moving to the batch methods might be worthwhile; think it might even make things easier?

pping/pping_kern.c
tohojo: Yeah, the requirement for libxdp to load the program is one of the more annoying warts of the API, but unfortunately there is no way to avoid it currently, and also no way to run code in-between load and attach. In practice it's probably fine to just input the key after attaching, I very much doubt the map will fill up before you get around to it. However, the kernel code will have to be prepared to deal with the key not existing, of course.

The memory usage comment was not specific to the backup maps, I just thought about it while writing that other comment which is why it is included here. My best guess for the behaviour of `libbpf_num_possible_cpus()` is that the VM probably supports hotplugging that many CPUs?

pping/pping.c
tohojo: Right, an optimisation that sometimes helps a lot, and at worst does nothing, and which is also easy to implement, does sound worthwhile to me, so let's just do it unconditionally :)

pping/pping.c
tohojo: I think it's pretty standard to just use `memcpy()` so go ahead and do that :)

tohojo:
> So apparently GitHub no longer actually publish my replies to your comments that I make directly in the conversation view

Ah, so Github acually managed to make their already terrible review interface even worse? Lovely.

Left a few additional comments (in the original conversation on my review that you already replied to). As a general point, while I do appreciate your thorough replies, you don't *have* to reply to everything, anything where you're just going to go with my suggestion you can just fix and not bother to reply :)

pping/pping_kern.c
simosund: > I very much doubt the map will fill up before you get around to it. However, the kernel code will have to be prepared to deal with the key not existing, of course.

Yhea, depending on what meaning you put into the "be prepared" to dealing with the key not existing, this may become somewhat complicated. The easiest (from an implementation perspective) way to interpret that is just that the BPF code can't expect the map lookup to be successful, but to even pass the verifier I pretty much have to include a check if the map lookup failed or not anyways. So the more interesting question is what action to take when that map lookup fails. The easiest is probably to just ignore aggregating that traffic (which is what I do currently, even if it should never really happen due the separate backup entry I currently have). 

From the BPF point of view, if we can't successfully lookup the backup key, I basically see 3 options:
1. Simply don't aggregate that traffic, essentially missing it.
2. Have a backup entry for the backup entry (for example in a separate map). This seems somewhat pointless, as at that point there's not really any point in not using the separate backup entry to begin with, and you basically just end up with the current solution (a separate backup) + some extra work for also having a backup key in the original map.
3. Ensure there's always space for a backup entry in the BPF programs, for example by having a counter (protected by a spinlock) of the number of entries created and stop creating entries when there's only one slot left. Once you reach this point, you could "miss" new traffic as described above until the user space process is able to create the backup entry. However, if you already have all of this logic to prevent the spot for the backup entry being taken, you might as well just create the backup entry in the BPF program space once you detect that can't create any more "unique" entries. Ideally, this logic should be inactivated after the first reporting period as then the backup entry should be guaranteed to exist. 

From the user space perspective, the creating of the backup entry may fail if the map is already full. I also see roughly 3 ways to handle that:
- a) Simply abort the entire ePPing program. In practice the map is unlikely to fill up that fast, so this shouldn't happen often. And if it does happen, you will hopefully get luckier next time you try to load ePPing...
- b) Just go on as normal, just that the backup keys may not exist for all instances of the aggregation maps, so some traffic may be lost every other period...
- c) Go on for now, but then try to insert the keys once the starts are reset after the first aggregation period. To make space for the backup entry, it may have to delete one of the existing entires (which it could example select as the entry with the least amount of traffic during the first period). If it still fails at inserting the backup entry...I guess we abort now? Or continue going on hoping it will work next reporting period...

Unless you have any objections, I will probably just go for easy 1a combination here, of on the BPF side missing traffic if the map lookup fails, and from the user space side abort the program if we are unable to create the backup entries. As you say, this edge case of the map filling up before user space can create the backup entry should hopefully not be much of a problem in practice. If you think I should go for one of the more complicated solutions that actually tries to handle this edge case, I would probably argue for just using the current solution with separate maps for the backup entries which avoids this edge case to begin with. While perhaps a bit ugly, it is less complicated than the other approaches I suggest here, and should also not really have any run time overhead (unlike for example keeping a synced counter of the number of entries created).

tohojo:
> Unless you have any objections, I will probably just go for easy 1a
> combination here, of on the BPF side missing traffic if the map lookup
> fails, and from the user space side abort the program if we are unable
> to create the backup entries.

Sure, by all means go with the easiest option!

simosund:
Hi again @tohojo,
Now I think I've resolved all of the things you pointed out in the last review, so please have another look at it.

To summarize, the main changes are:
- All new options that were previously parsed as float values are now parsed as int values
- Have made the differently sized IPv4 and IPv6 keys for the aggregation maps type-safter. Instead of void pointers it now mainly relies on a union instead (which can either have a 32-bit IPv4 key or a 64-bit IPv6 key). While not perfect, it's hopefully at least a significant improvement from before.
- The user space process now uses a separate loop for fetching/resetting the aggregated entries, and deleting expired entries. Also switched to using batched operations (with `AGG_BATCH_SIZE` set to 64 as some rather arbitrary compromise between reducing syscall overhead and have to allocate large chunks of memory).
- The backup aggregation entries are now just special keys in the existing aggregation maps rather than entierly separate maps. To avoid these special keys colliding with real traffic, I've used prefixes blocks reserved for documentation (specifically `192.0.2.255/32` for IPv4 and `2001:db80:ffff:ffff::/64` for IPv6). In case user space fails to add these special keys it will error out.
- Blocking on a timer instead of `nanosleep()` in the aggregation loop to avoid the aggregation period drifting.
- The user space process will no longer overwrite any existing file, If you try to write output (using the `-w`/`--write`) option to an existing file, the program will error out.
- Removed the "rotation" options. Have instead added a signal handler for `SIGHUP` which causes the output file to be reopened, making it possible to use ex. logrotate. Combined with the change of not overwriting files this means that the program will error out in case it receives a `SIGHUP` signal and its current file has NOT been renamed (as it cannot reopen the same file).
- Added a check that tries to restrain the length of the filename provided to `-w`/`--write`, so that there's room to append some stuff (see the final commit) when rotating the file. I would prefer to drop this commit.  First off, I'm not convinced that this check will actually work properly (apparently you can't really trust `PATH_MAX` and `FILENAME_MAX`, so tried to get those values using `pathconf()` on the containing folder, but not sure how reliable that is). Second off, as we no longer have any built-in option to rotate the files, I don't think it's our concern anymore. If the user wants to rotate the files with some external utility, they will simply have to make sure they leave some space for appending stuff in the filename themselves. We can't really know what they would even like to append, so there's no way for us to really know how much space to reserve anyways. While I would prefer to drop this commit, I have left it here for now so you can have a say as well. 

A small implementation related note is that  I'm not really sure how to best import your `try_snprintf()` function, as it does not seem to be a publically exported function from libxdp. So I simply went and copy-pasted that specific function. If you have any suggestions for how to properly import it I'll gladly change that.

simosund:
Did a force push after I requested the review, but it only touched `lhist.h` commit #[f6b03a7](https://github.com/xdp-project/bpf-examples/pull/59/commits/f6b03a72e496de92ffb76ac31afa0e724aa96d5f) ("Improve aggregated output format"). Basically I just fixed an error with how the 0th and 100th percentile (aka min and max values) were calculated from the histogram, or any percentile when there was just a single value in the histogram.

pping/pping.c
```diff
@@ -869,6 +1290,398 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
...
+	int n_cpus = libbpf_num_possible_cpus();
+	size_t keysize = af == AF_INET ? sizeof(__u32) : sizeof(__u64);
+	__u32 count = AGG_BATCH_SIZE, total = 0, del_idx = 0;
+	__u64 batch;
+	int err = 0, i;
+	bool del_key;
+
+	DECLARE_LIBBPF_OPTS(bpf_map_batch_opts, batch_opts, .flags = BPF_EXIST);
+
+	values = malloc(sizeof(*values) * n_cpus * AGG_BATCH_SIZE);
```
tohojo, line 1540: Elsewhere you're using `calloc()` for the per-cpu allocation; I'd suggest you do the same here. I know it has to be a *very* high number of CPUs for this to overflow, but passing anything that's multiplied by a runtime value to `malloc()` always looks scary :)

pping/pping.c
```diff
@@ -155,40 +206,140 @@ static int parse_bounded_double(double *res, const char *str, double low,
...
+	if (*res < low || *res > high) {
+		fprintf(stderr, "%s must be in range [%lld, %lld]\n", name, low,
+			high);
+		return -ERANGE;
+	}
+
 	return 0;
 }
 
+static int verify_filename_length(const char *path, int reserve_space)
```
tohojo, line 256: Let's drop this. Not worth the complexity: as long as we make sure to not overflow any internal buffers, just make it up to the user to make sure they don't overflow their system path length...

pping/pping.c
```diff
@@ -155,40 +206,140 @@ static int parse_bounded_double(double *res, const char *str, double low,
...
+	config->bpf_config.localfilt = true;
 	config->bpf_config.track_tcp = false;
 	config->bpf_config.track_icmp = false;
 	config->bpf_config.skip_syn = true;
+	config->bpf_config.push_individual_events = true;
+	config->bpf_config.agg_rtts = false;
+	config->bpf_config.agg_by_dst = false;
 
-	while ((opt = getopt_long(argc, argv, "hflTCsi:r:R:t:c:F:I:x:",
+	while ((opt = getopt_long(argc, argv, "hflTCsei:r:R:t:c:F:I:x:a:4:6:o:w:",
```
tohojo, line 332: Not every long option *has* to have a short option as well. To save a bit of alphabet I'd suggest thinking about which options are unlikely to be set in the common case, and just drop the short options for those. `--aggregate-reverse` and `--aggregate-timeout` come to mind as candidates for this...

tohojo (changes requested):
A few nits, but other than that I think the handling of file reopening is the main thing that needs to be fixed before we can merge this. I.e., when reopening we need to open the new file before closing the old one, so we can keep running with the old file if the reopen fails, instead of crashing.

I think the way to go about this is to switch the signal handling to using a signalfd. When combined with a suitably set `sigprocmask`, this avoids the weird asynchronous signal callbacks, and you can just have the main event loop combine the timerfd with the signalfd using one of the standard mechanisms (select/poll/epoll/etc; no strong preference, but epoll seems to be the modern interface). And then, when you're not aggregating, just drop the timerfd, but keep the loop the same just reacting to the signalfd.

As for how to integrate this with the perf buffer polling, I can see two options:

1. Stop using `perf_buffer__poll()`, and do your own polling of the underlying file descriptors. Libbpf actually exposes an API to do this: you get the fds for the per-cpu buffers using `perf_buffer__buffer_fd()`, then add those to the same fd multiplexing you're using for the rest above, and when one of those fds comes up as ready, you call `perf_buffer__consume_buffer()` with the corresponding index. This way you can integrate everything into the same event loop.
2. You keep the `perf_buffer__poll()` loop in a separate thread (but move that out of the main thread to its own thing), and use regular cross-thread communication of some kind to signal that thread to reopen the file (or reopen in the main thread, but keep a lock on the write fd to synchronise).

My guess would be that 1. would be the simplest, just because it avoids the cross-thread synchronisation, so you can run everything in the main thread. You could even integrate the periodic map cleanup into the same loop and avoid threading completely, although since the cleanup is potentially heavy, and pretty self-contained maybe keeping it in its own thread is better...

tohojo:
As for the variable-size histograms, the canonical way to represent such a "struct with an unknown-size member at the end" is by including a *flexible array* member at the end: https://en.wikipedia.org/wiki/Flexible_array_member

Not sure how BTF represents this, but there are quite a few of those in the kernel, so I guess they are representable? With this, you should be able to avoid the void pointer soup :)

pping/pping.c
simosund: Hmm, I was using `calloc()`mainly to get it zero initalized, but I guess making it less likely to overflow can be a valid reason as well. That said, we have three values multiplied here, so one of them will always be multiplied with a runtime value, and I personally find it more natural to multiply the bigger ones here (`sizeof(*values)` and `n_cpus`).

pping/pping.c
simosund: Good, we agree on that.

simosund:
> My guess would be that 1. would be the simplest, just because it avoids the cross-thread synchronisation, so you can run everything in the main thread. You could even integrate the periodic map cleanup into the same loop and avoid threading completely, although since the cleanup is potentially heavy, and pretty self-contained maybe keeping it in its own thread is better...

I like the idea of combining the perf-buffer and aggregation loops, so I don't need separate threads for them (espeically as only one them is going to be particularly active at a time anyways). So will try to implement your suggestion 1 early next week (quite busy with trying to hit a paper deadline this week).

> As for the variable-size histograms, the canonical way to represent such a "struct with an unknown-size member at the end" is by including a flexible array member at the end: https://en.wikipedia.org/wiki/Flexible_array_member

Huh, (obviously) wasn't aware of that being a thing, so thanks for the pointer. That should make the code a lot clearer.

pping/pping.c
tohojo: Heh, yeah, the main reason `calloc()` exists is to allocate arrays without overflowing. The zero-init is just an incidental (but useful) side effect. Also, `n_cpus` is the only of the three values that's not a compile-time constant, so that's definitely the one that should be moved to be the first element of `calloc()` instead :)

tohojo:

>> My guess would be that 1. would be the simplest, just because it avoids the cross-thread synchronisation, so you can run everything in the main thread. You could even integrate the periodic map cleanup into the same loop and avoid threading completely, although since the cleanup is potentially heavy, and pretty self-contained maybe keeping it in its own thread is better...
>
> I like the idea of combining the perf-buffer and aggregation loops, so
> I don't need separate threads for them (espeically as only one them is
> going to be particularly active at a time anyways). So will try to
> implement your suggestion 1 early next week (quite busy with trying to
> hit a paper deadline this week).

Sounds good!

simosund:
Hi @tohojo, the PR is now ready for you to have another look at it. 

I've gone with merging the perf-buffer, aggregation timer and signal handling (via signalfd) into a single event loop. To try and add it in a way that makes sense in this PR I've added an additional commit early in the PR ([Use signalfd instead of signalhandler](https://github.com/xdp-project/bpf-examples/pull/59/commits/e9d2c0ee37b457aa8189cf37ecedb3ad41dd3625)) which adds the synchronous signal handling and merges it with the perf-buffer polling. The first commit which adds the actual aggregation implementation ([Add option to aggregate RTTs](https://github.com/xdp-project/bpf-examples/pull/59/commits/41ee3a9f931d5496d300ee3f608ce2f989cd620c)) then simply adds the timerfd used for the periodic reporting into the now existing epoll loop. As the aggregation uses timerfd right from the start now I also dropped the commit which changes it from a sleep-loop to using timerfd.

pping/pping.c
```diff
@@ -1663,6 +1662,28 @@ static int close_output(struct output_context *out_ctx)
...
+	int err;
+
+	if (!out_ctx->write_to_file)
+		return 0;
+
+	memcpy(&prev_out, out_ctx, sizeof(*out_ctx));
+
+	err = open_output(out_ctx, print_aggconf, agg_conf);
+	if (err)
+		return err;
```
tohojo, line 1682: This should log an error if it fails...

tohojo (changes requested):
Only looked at the new event handling code, and that looks mostly fine with just a few comments here and there :)

pping/pping.c
```diff
@@ -1046,12 +1053,170 @@ static int setup_periodical_map_cleaning(struct bpf_object *obj,
...
+		fd = perf_buffer__buffer_fd(pb, cpu);
+		if (fd < 0)
+			return fd;
+
+		ev.data.u64 = cpu;
+		ev.data.u64 |= EPOLL_EVENT_TYPE_PERFBUF;
+
+		err = epoll_ctl(epfd, EPOLL_CTL_ADD, fd, &ev);
+		if (err)
+			return -errno;
```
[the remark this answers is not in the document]
simosund: Not sure I quite follow, what error-checking wrapper are you talking about here?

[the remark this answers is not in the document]
tohojo: Oh, ffs, the useless github review interface ate that comment :(

I was suggesting replacing the `add_signalfd_to_epoll_instance()` function with a generic function, like:

```c
static int epoll_add_event_type(int epfd, __u64 event_type, __u64 value)
{
	struct epoll_event ev = {
		.events = EPOLLIN,
		.data = { .u64 = event_type | value },
	};

	if (value & ~PPING_EVENT_MASK)
		return -EINVAL;

	return epoll_ctl(epfd, EPOLL_CTL_ADD, sigfd, &ev) ? -errno : 0;
}
```

pping/pping.c
```diff
@@ -36,6 +37,12 @@ static const char *__doc__ =
 #define PROG_INGRESS_XDP "pping_xdp_ingress"
 #define PROG_EGRESS_TC "pping_tc_egress"
 
+#define EPOLL_TIMEOUT_MS 100
```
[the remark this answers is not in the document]
simosund: Well, there is one reason to have a reasonably short timeout here still and that's if the periodical map cleanup fails and needs to abort the program, the main event loop won't stop until the next...well loop. But I guess the map cleaning thread could send a `SIGINT`/`SIGTERM` signal to the main thread instead of changing the global `keep_running` variable to abort it in a more synchronous manner. Then we could get rid of the `keep_running` variable entirely, and just have the signal-handling function return some special "error" which causes the main event loop to break without printing an error.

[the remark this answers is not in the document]
tohojo: Right, you could either do a `kill(0, SIGINT)` from the cleanup thread, or you could create a pipe (with `pipe(2)`), but the reader end into the epoll loop, and have the cleanup thread write to the writer end when it wants to exit. Not sure which is simpler; I suspect the `kill()`, but that will have side effects if you ever introduce child processes...

pping/pping.c
```diff
@@ -1115,6 +1410,109 @@ static int init_perfbuffer(struct bpf_object *obj, struct pping_config *config,
...
 }
 
+int fetch_aggregation_map_fds(struct bpf_object *obj,
+			      struct aggregation_maps *maps)
+{
+	char map_name[64];
+	int fd, ipv;
+	int *fd_dst;
+
+	for (ipv = 4; ipv <= 6; ipv += 2) {
```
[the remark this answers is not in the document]
simosund: Yhea, partly a remnant of older code when I had separate maps for the backup entries as well, so had a total of 8 aggregation maps (and did not want to repeat the same map-fetching logic 8 times, although it instead resulted in a rather gnarly 3-level loop). When there's just two maps this loop is definitely just complicating things. 

In the next commit though it does become 4 maps (2 instances each of the v4 and v6 maps). So to deal with that, what do you think of something like:
```c
int fetch_aggregation_map_fds(struct bpf_object *obj,
			      struct aggregation_maps *maps)
{
	char *map_names[] = { "map_v4_agg1", "map_v4_agg2", "map_v6_agg1",
			      "map_v6_agg2" };
	int *fd_dsts[] = { &maps->map_v4_fd[0], &maps->map_v4_fd[1],
			   &maps->map_v6_fd[0], &maps->map_v6_fd[1] };
	int i, fd;

	for (i = 0; i < 4; i++) {
		fd = bpf_object__find_map_fd_by_name(obj, map_names[i]);
		if (fd < 0) {
			fprintf(stderr,
				"Unable to find aggregation map %s: %s\n",
				map_names[i], get_libbpf_strerror(fd));
			return fd;
		}

		*fd_dsts[i] = fd;
	}

	return 0;
}
```
Or would you prefer just having one loop each for the v4/v6? Or just straight up just have 4 copy-pasted calls to fetch the maps?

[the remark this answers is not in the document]
tohojo: Hmm, so this is an error that only occurs when there's something wrong with the BPF program compilation, and there's not really much a user can do with the error message anyway. So how about just collapsing it to:

```c
int fetch_aggregation_map_fds(struct bpf_object *obj,
			      struct aggregation_maps *maps)
{
	maps->map_v4_fd[0] = bpf_object__find_map_fd_by_name(obj, "map_v4_agg1");
	maps->map_v4_fd[1] = bpf_object__find_map_fd_by_name(obj, "map_v4_agg2");
	maps->map_v6_fd[0] = bpf_object__find_map_fd_by_name(obj, "map_v6_agg1");
	maps->map_v6_fd[1] = bpf_object__find_map_fd_by_name(obj, "map_v6_agg2");

	if (maps->map_v4_fd[0] < 0 || maps->map_v4_fd[1] < 0 || maps->map_v6_fd[0] < 0 || maps->map_v6_fd[1] < 0) {
			fprintf(stderr,
				"Unable to find aggregation maps (%d/%d/%d/%d).\n",
				maps->map_v4_fd[0] , maps->map_v4_fd[1] , maps->map_v6_fd[0] , maps->map_v6_fd[1]);
			return -ENOENT;
		}
	}

	return 0;
}
```

pping/pping.c
```diff
@@ -1663,6 +1662,28 @@ static int close_output(struct output_context *out_ctx)
 	return 0;
 }
 
+/* Will try to reopen the output.
+ * If it is able to open the new output it will replace the current output
+ * and close the old output. Otherwise, it will simply leave the current output
+ * as is. */
```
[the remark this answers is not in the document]
simosund: So now that I updated this I realized that this approach actually requires two copies. I must first copy `out_ctx` to `new_out` in order for `open_output(&new_out)` to work (so it knows what file to open etc), and then copy it `new_out` back to `out_ctx` after it's been closed. 

Was already not a huge fan of the one copy I had there to begin with, so starting to wonder if it's perhaps better refactor it in some way so I won't have to do these copies back and fourth. To tired right now to think of a good way to do it, but would either have to rework the `open_output()` and `close_output()` functions a bit to make them more flexible or factor out components of them that I can reuse in `reopen_output()` without calling them directly.

pping/pping.c
simosund: Added it. But this means it will write an error in case a SIGHUP is received before the file has been renamed (as it will then fail to exclusively open the file as it already exists), although it will continue running.

tohojo: Yeah, that's the point - if it doesn't complain, users are not likely to notice this and think that everything went well :)

pping/pping.c
```diff
@@ -1991,8 +2018,10 @@ static int poll_events(int epfd, struct pping_config *config,
 			if (err)
 				goto err_abort;
 		} else if (events[i].data.u64 & EPOLL_EVENT_TYPE_SIGNAL) {
-			err = handle_signalfd(events[i].data.u64 &
-					      EPOLL_DATA_MASK);
+			err = handle_signalfd(
+				events[i].data.u64 & EPOLL_DATA_MASK,
+				&config->out_ctx, config->bpf_config.agg_rtts,
+				&config->agg_conf);
```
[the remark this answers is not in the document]
simosund: Well, in this particular instance it's mainly because when I wrote the `handle_signalfd()` function I didn't think about that those 3 arguments could all be found in the config object. But on a more general level I typically only let "outer" functions that the main-function calls directly handle the config object, while "inner" functions only get the parts they need. Otherwise I could change almost every function to just take the config object, as it contains just about everything.

Now I've been pretty bad at using the `const` specifier in the parameters as most functions shouldn't actually change anything in the structs they're passed. But assuming I had been a bit better at doing that splitting up the parameters would also allow me to better protect/indicate which parts the function can actually modify (in this instance it may need to modify the `out_ctx`, but not `agg_conf`).

pping/pping.c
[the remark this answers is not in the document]
tohojo: I'd suggest the following:
- Move the filename and write_to_file boolean out of the output_ctx struct (they're configuration options, not context values)
- Have the open_output function dynamically allocate the context; something like this:

```c
static struct output_context *open_output(const char *filename, enum pping_output_format format,
                                          struct aggregation_config *agg_conf)
{
        struct output_context *out_ctx;

        out_ctx = calloc(1, sizeof(*out_ctx));
        if (!out_ctx)
                return NULL;

        out_ctx->format = format;

	if (filename) {
		out_ctx->stream = fopen(filename, "ax");
		if (!out_ctx->stream)
			goto err;
	} else {
		out_ctx->stream = stdout;
	}

	if (out_ctx->format == PPING_OUTPUT_JSON) {
		out_ctx->jctx = jsonw_new(out_ctx->stream);
                if (!out_ctx->jctx)
                        goto err;
		jsonw_start_array(out_ctx->jctx);
	}

	if (agg_conf)
		print_aggmetadata(out_ctx, agg_conf);

	return out_ctx;
err:
        free(out_ctx);
        return NULL;
}
```

simosund:
Hi again @tohojo,
I think I've addressed all of the things you pointed out in your previous review. When it comes to how to tell the main loop to shut down from the map cleaning thread I went with the pipe as you suggested. While I think the pipe part works fine, I don't think the code for actually aborting the main loop in `epoll_pull_events()` turned out great, so I suspect you might have some comments on that...

I will also note that I've run into another issue, which is that sometimes it fails to create the backup entries `init_agg_backup_entries()` with "Cannot allocate memory" (specifically it's the bpf_map_update_elem() calls that fail). This seems a bit weird to me because the machine has plenty of free memory at the time, but I guess there might be some other restrictions comming into play when the BPF maps have to dynamically allocate memory? It's an intermittent fault, so not something I've been able to trigger reliably.

pping/pping.c
```diff
@@ -1094,75 +2159,110 @@ int main(int argc, char *argv[])
...
 	// Cleanup
-	if (config.output_format == PPING_OUTPUT_JSON && json_ctx) {
-		jsonw_end_array(json_ctx);
-		jsonw_destroy(&json_ctx);
-	}
+cleanup_epfd:
+	close(epfd);
+
+cleanup_aggfd:
+	if (config.bpf_config.agg_rtts)
```
tohojo, line 2259: Nit: Instead of checking the config var here, I'd check the fd; that has fever local dependencies (i.e., if you ever decide to change the name of the config var. you'll need to update this check as well, whereas if you just check the fd value you don't).

pping/pping.c
```diff
@@ -1094,75 +2159,110 @@ int main(int argc, char *argv[])
...
 
 	// Main loop
-	while (keep_running) {
-		if ((err = perf_buffer__poll(pb, PERF_POLL_TIMEOUT_MS)) < 0) {
-			if (keep_running) // Only print polling error if it wasn't caused by program termination
-				fprintf(stderr,
-					"Error polling perf buffer: %s\n",
-					get_libbpf_strerror(-err));
+	while (!abort) {
+		err = epoll_poll_events(epfd, &config, pb, -1, &abort);
```
tohojo, line 2246: You don't really need the abort argument as you're already checking the return code; just define a special return code for 'abort without error' and use that?

pping/pping.c
```diff
@@ -1039,12 +1642,469 @@ static int setup_periodical_map_cleaning(struct bpf_object *obj,
...
+			fprintf(stderr, "Warning: unexpected epoll data: %lu\n",
+				events[i].data.u64);
+		}
+	}
+
+	return 0;
+
+err:
+	fprintf(stderr, "Failed handling events: %s\n",
+		get_libbpf_strerror(err));
```
tohojo, line 2096: You already print local error messages where the errors occur, so this additional error message doesn't really add that much value, I think? In which case you could get rid of it, initialise 'err' variable to 0 at the start of the function, turn the `return 0` above into `return err` and turn the `goto err` statements into simple `break`s. Along with getting rid of the `abort` variables, that should make the above somewhat shorter.

You could also restructure the if statements into a `switch`, like:
```c
for (i = 0; i < nfds; i++) {
  switch (events[i].data.u64 & ~PPING_EEVENT_MASK) {
  case PPING_EEVENT_TYPE_PERFBUF:
    err = perf_buffer__consume_buffer(...);
    break;
....
  default:
    fprintf(stderr, "Warning: unexpected epoll data: %lu\n",
				events[i].data.u64);
    err = -EINVAL;
    break;
  }
  if (err)
    break;
}
```

pping/pping.c
```diff
@@ -1039,12 +1642,469 @@ static int setup_periodical_map_cleaning(struct bpf_object *obj,
...
+		fprintf(stderr,
+			"Failed creating timer for periodic aggregation: %s\n",
+			get_libbpf_strerror(fd));
+		return fd;
+	}
+
+	return fd;
+}
+
+static int handle_aggregation_timer_expiration(
```
tohojo, line 1935: Nit: long function name, maybe drop `_expiration`?

pping/pping.c
```diff
@@ -1039,12 +1642,469 @@ static int setup_periodical_map_cleaning(struct bpf_object *obj,
...
+
+	return out_ctx;
+err:
+	free(out_ctx);
+	return NULL;
+}
+
+static int close_output(struct output_context *out_ctx)
+{
+	if (out_ctx->format == PPING_OUTPUT_JSON && out_ctx->jctx) {
```
tohojo, line 1683: Nit: Here you also have the redundant double check; just check `out_ctx->jctx`

pping/pping.c
```diff
@@ -1039,12 +1642,469 @@ static int setup_periodical_map_cleaning(struct bpf_object *obj,
...
+		jsonw_end_array(out_ctx->jctx);
+		jsonw_destroy(&out_ctx->jctx);
+	}
+
+	if (out_ctx->stream && out_ctx->stream != stdout) {
+		if (fclose(out_ctx->stream) != 0)
+			return -errno;
```
tohojo, line 1690: This return skips the `free()` below.

tohojo (changes requested):
Mostly a few nits left; getting close! :)

tohojo:
> I will also note that I've run into another issue, which is that sometimes it fails to create the backup entries `init_agg_backup_entries()` with "Cannot allocate memory" (specifically it's the bpf_map_update_elem() calls that fail). This seems a bit weird to me because the machine has plenty of free memory at the time, but I guess there might be some other restrictions comming into play when the BPF maps have to dynamically allocate memory? It's an intermittent fault, so not something I've been able to trigger reliably.

Oh, and I don't really have a good explanation for this. Could be a memory limit issue (BPF uses memory cgroup limits on newer kernels), maybe? Hard to tell without a reliable reproducer :(

simosund:
So a small update. Have not gotten around to addressing the latest comments today as I had hoped to. While testing out getting rid of the `BPF_F_NO_PREALLOC` flag I found another issue (hurray...). It seems like the batched lookups of the aggregation maps (in `report_aggregated_rtt_map()`) are not working properly. 

More specifically, they don't always loop over all entries. So I'm trying to figure out what's wrong there. If I switch to using non-batched lookups (`bpf_map_get_next_key()` combined with `bpf_map_lookup_elem()`) it seems to be working as expected, but with batched lookups (`bpf_map_lookup_batch()`) it's skipping over some entries in a somewhat inconsistent manner (generally quite consistent between all reports within one run of the pping application, but different between every run). So will need to figure out exactly what's wrong with how I'm doing the batched lookups and fix that. 

Based on some limited testing with concurrent traffic for around 1000 IP-prefixes, the problem seems to get worse with a smaller batch size. With the default 64 batch size, it will often be able to report all of the entries from one map, but miss several hundred from the second instance of the map. When lowering the batch size to just 2 it will only report a handful (around 4 - 200 out of 1k in my tests) of entries from each instance of the map, and when lowering it down to 1 it will report between 20-80 prefixes before crashing with `ENOSPC`.

simosund:
Hi @tohojo,
I've now fixed the things you pointed out in your last review. I've also added an additional commit at the end with removes the `BPF_F_NO_PREALLOC` flag from the aggregation maps, due to the occasional memory allocation issues I've been having when not preallocating all entries. In that commit I also reduced the histogram resolution to compensate for the increased memory usage.

Furthermore, I've slightly rewritten the loop for fetching the aggregated data in batches a little bit, so that it should no longer semi-randomly skip a number of entries. The issue was that I had assumed that it would fetch the full batch size up until the final round (i.e. when there's less than batch size left). This assumption turned out to be wrong, some rounds it may fetch less than the asked for batch size even if there's plenty of entries left, so have now rewritten the logic to simply continue the loop until `ENOENT` is returned. It still randomly fails with `ENOSPC` if the batch size is too small, but from a batch size of 4 and upwards it seems to work for 1k entries at least (and I'm using a batch size of 64, so should hopefully be fine).

simosund:
In my recent force-push I just fixed an issue with the PPING_ABORT "error" being propegated, causing pping to exit with a non-zero exit code. So I simply changed the main loop from:
```
	while (true) {
		err = epoll_poll_events(epfd, &config, pb, -1);
		if (err) {
			if (err != PPING_ABORT)
				fprintf(stderr, "Error polling events: %s\n",
					get_libbpf_strerror(err));
			break;
		}
	}
```
to:
```
	while (true) {
		err = epoll_poll_events(epfd, &config, pb, -1);
		if (err) {
			if (err == PPING_ABORT)
				err = 0;
			else
				fprintf(stderr, "Error polling events: %s\n",
					get_libbpf_strerror(err));
			break;
		}
	}
```

simosund:
Hi again @tohojo,
I understand if you have a lot of other things to deal with, but if you have time to take a final look at this before you go on vacation that would be splendid. Have a nice summer!

pping/pping.c
```diff
@@ -36,7 +45,32 @@ static const char *__doc__ =
...
+	(~(PPING_EPEVENT_TYPE_PERFBUF | PPING_EPEVENT_TYPE_SIGNAL |            \
+	   PPING_EPEVENT_TYPE_PIPE | PPING_EPEVENT_TYPE_AGGTIMER))
+
+#define AGG_BATCH_SIZE 64 // Batch size for fetching aggregation maps (bpf_map_lookup_batch)
+
+/* Value that can be returned by functions to indicate the program should abort
+ * Should ideally not collide with any error codes (including libbpf ones), but
+ * can also be seperated by returning as positive (as error codes are generally
+ * returned as negative values). */
+#define PPING_ABORT 5555
```
tohojo, line 68: Missed opportunity here of making this be 6666 instead 😅🤘😈

pping/pping.c
```diff
@@ -1039,12 +1655,465 @@ static int setup_periodical_map_cleaning(struct bpf_object *obj,
...
+					&config->agg_conf :
+					NULL);
+			break;
+		case PPING_EPEVENT_TYPE_PIPE:
+			err = handle_pipefd(events[i].data.u64 &
+					    PPING_EPEVENT_MASK);
+			break;
+		default:
+			fprintf(stderr, "Warning: unexpected epoll data: %lu\n",
+				events[i].data.u64);
```
tohojo, line 2100: So this doesn't set err, which means the loop will continue. I'm not sure whether this was deliberate, actually; continuing in this case would not arguably be the wrong thing to do...

tohojo (approved):
So I basically ran out of things to complain about - nice work! :)

Just two minor nits below, none of which are critical (and only one that's actually totally serious). No reason to hold up the PR because of this, so going to merge it now, and feel free to follow up if you think any of the minor issues are worth fixing...

pping/pping.c
simosund: Yes, in this case that was deliberate. My thinking was basically that if we run into some unknown epoll event we can just skip it (with a warning). Granted I don't really see any reason to why we would run into an "unknown" event.

tohojo: Right, that's fine, and yeah, shouldn't happen so let's just leave it like this :)

PR #65, opened 2022-11-08
Fix ePPing issue with LLVM-15

### LINKED ISSUE #63: pping: issues with LLVM-15
First of all, it would be of immense help if the steps to execute an application is mentioned in detail (at least the step by step commands). This saves a lot of time for people who are starting out with eBPF. 

The steps I am following are mentioned below one by one.

```git clone https://github.com/xdp-project/bpf-examples.git```
```./configure```
```make``` gives me error from both the root directory ```bpf-examples``` or from ```pping```

To overcome the error, I have to go ```/lib/xdp-tools``` and then run ```./configure``` followed by ```make```
Next, I again run ```make``` from the root directory ```bpf-examples```. Although errors due to ```libxdp.c``` are not generated, yet error is generated due to ```net64-bpf``` application
```
nat64-bpf
    CLANG    nat64_kern.o
    LLC      nat64_kern.o
    GEN      nat64_kern.skel.h
libbpf: elf: skipping unrecognized data section(9) .rodata.str1.1
    CC       nat64
nat64.c:20:10: fatal error: libmnl/libmnl.h: No such file or directory
   20 | #include <libmnl/libmnl.h>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
```

Then all the other applications are commented out except ```pping``` in the ```Makefile``` of the root directory. ```make``` is run again from the root and compilation is successful.

Next, running ```pping``` for TC gives the following error:
```
sudo ./pping -i lo -F json -I tc
Starting ePPing in json mode tracking TCP on lo
libbpf: prog 'tsmap_cleanup': BPF program load failed: Permission denied
libbpf: prog 'tsmap_cleanup': -- BEGIN PROG LOAD LOG --
R1 type=ctx expected=fp
; int tsmap_cleanup(struct bpf_iter__bpf_map_elem *ctx)
0: (bf) r8 = r1
; __u64 *timestamp = ctx->value;
1: (79) r9 = *(u64 *)(r8 +24)
; struct packet_id *pid = ctx->key;
2: (79) r6 = *(u64 *)(r8 +16)
; __u64 now = bpf_ktime_get_ns();
3: (85) call bpf_ktime_get_ns#5
4: (bf) r7 = r0
; debug_update_mapclean_stats(ctx, &events, !ctx->key || !ctx->value,
5: (79) r1 = *(u64 *)(r8 +16)
; debug_update_mapclean_stats(ctx, &events, !ctx->key || !ctx->value,
6: (55) if r1 != 0x0 goto pc+3
 R0_w=invP(id=3) R1_w=invP0 R6_w=rdonly_buf_or_null(id=2,off=0,imm=0) R7_w=invP(id=3) R8_w=ctx(id=0,off=0,imm=0) R9_w=buf_or_null(id=1,off=0,imm=0) R10=fp0
; ctx->meta->seq_num, now,
7: (79) r1 = *(u64 *)(r8 +0)
func 'bpf_iter_bpf_map_elem' arg0 has btf_id 30838 type STRUCT 'bpf_iter_meta'
; ctx->meta->seq_num, now,
8: (79) r1 = *(u64 *)(r1 +16)
9: (05) goto pc+4
; if (map_stats->start_time) { // Non-empty map
14: (18) r2 = 0xffffb359c01a2010
.
.  // Other statements not included
.
164: (bf) r1 = r8
165: (07) r1 += 72
R1 pointer arithmetic on map_value_or_null prohibited, null-check it first
processed 254 insns (limit 1000000) max_states_per_insn 0 total_states 19 peak_states 19 mark_read 10
-- END PROG LOAD LOG --
libbpf: failed to load program 'tsmap_cleanup'
libbpf: failed to load object 'pping_kern.o'
Failed loading bpf programs in pping_kern.o: Permission denied
Failed loading and attaching BPF programs in pping_kern.o
```

Platform: Ubuntu 20.04.5 LTS; 
Kernel: 5.15.0-52-generic

Any help is really appreciated.

simosund:
ePPing failed to load when compiled with LLVM-15, as the verifier would reject the generated bytecode for the `tsmap_cleanup` BPF program. This commit PR contains a simple commit to solve this issue, and thus resolves #63. See the commit message for more details.

### DIFF
--- a/pping/pping_kern.c
+++ b/pping/pping_kern.c
@@ -228,9 +228,6 @@ static void make_dualflow_key(struct network_tuple *key,
 static struct flow_state *fstate_from_dfkey(struct dual_flow_state *df_state,
 					    bool is_dfkey)
 {
-	if (!df_state)
-		return NULL;
-
 	return is_dfkey ? &df_state->dir1 : &df_state->dir2;
 }
 
@@ -1127,7 +1124,7 @@ SEC("iter/bpf_map_elem")
 int tsmap_cleanup(struct bpf_iter__bpf_map_elem *ctx)
 {
 	struct packet_id local_pid;
-	struct flow_state *f_state;
+	struct flow_state *f_state = NULL;
 	struct dual_flow_state *df_state;
 	struct network_tuple df_key;
 	struct packet_id *pid = ctx->key;
@@ -1146,7 +1143,8 @@ int tsmap_cleanup(struct bpf_iter__bpf_map_elem *ctx)
 
 	make_dualflow_key(&df_key, &pid->flow);
 	df_state = bpf_map_lookup_elem(&flow_state, &df_key);
-	f_state = get_flowstate_from_dualflow(df_state, &pid->flow);
+	if (df_state)
+		f_state = get_flowstate_from_dualflow(df_state, &pid->flow);
 	rtt = f_state ? f_state->srtt : 0;
 
 	if ((rtt && now - *timestamp > rtt * TIMESTAMP_RTT_LIFETIME) ||


tohojo:
What happens if you just get rid of the old check (instead of having two of them)?

simosund:
Will have to check that. I suspect that I might have to add additional NULL-checks in other places as well then (anywhere `fstate_from_dfkey()` might be called). But will have to take a closer look at that tomorrow.

simosund:
Turns out there was no problems getting rid of the old NULL-check inside the `fstate_from_dfkey()` function, as in all the other places it ended up being called there were already NULL-checks on `df_state` anyways (to allow for early returns).

PR #72, opened 2022-11-22
Implement AF_XDP_filter test case

tjcw:
Split into 4 commits
Bring in baseline source and headers from xdp-tutorial
Implement the code for flow filtration
Provide run scripts
Remove obsolete (commented-out) code

Still to be done 
Fix unexpected ENOTSOCK which looks like a kernel or libbpf issue
Fix why first packet of a flow (injected by tun/tap) is rejected by kernel as a martian packet


tjcw:
The unexpected ENOTSOCK was due to a bug in libxdp. This had already been fixed by commit b572cd6b5a6462280db389b94586c2b5450887a9 , so rebuilding with this commit or newer results in the eBPF code loading as desired.

The first packet of a flow being dropped is due to the "reverse path filtering" feature of the linux kernel; an attempt to avoid processing of 'spoof' packets which come in on an interface that is not configured for them. This feature can be turned off with this in the run script
```
for device in /proc/sys/net/ipv4/conf/*
do
  echo 0 >${device}/rp_filter
done
```
and this done, the test case works as desired.

The test case is now fully functional, and I will clear the 'draft' flag.

simosund:
Hi, I think @tohojo has gone on vacation over the Christmas holidays, so I think you may simply have to wait until some time in January when he gets back. I will not attempt to review this myself, but I see that the [DCO](https://github.com/apps/dco) test failed, so I can at least point out one little thing I think @tohojo would request you fix. So please rewrite your commit messages so that all of them contain the `Signed-off-by` tag (seem to mainly be missing in your first few commit messages).

Another small thing I think @tohojo might ask for is that you keep the "title" of the commit message below 72 characters, and line wrap any text in the body of the commit message at 72 characters as well. This may seem petty, but having properly formatted commit can help a lot when you have to review lots of code.

tjcw:
@simosund How do I rewrite and sign off a commit that is not at the tip of development ? I know how to use `git commit --amend` to revise the newest commit, but not how to go back and change history.
I wrote (and have permission to contribute) all the revisions in this PR, so giving a DCO is not a problem.

tjcw:
@simosund I have figured out how to revise commits further back than HEAD; now my DCOs are correct. I will look for lines longer than 72 characters.

tjcw:
@simosund I have reformatted the commit comments to fit in 72 characters per line

simosund:
Great, now the commits look a lot nicer and pass the DCO test! Unfortunately I don't really think I'm qualified to review the code itself (and even if I was, I don't have time for that right now), so I'm afraid you will simply have to wait for @tohojo to get back. But at least you've fixed up a couple of smaller issues with the commit messages, so you've saved yourself from a couple of changes @tohojo would otherwise ask you to fix (based on my own experience of having my code reviewed by @tohojo =)).

tjcw:
@tohojo Is everyone back from the Christmas/New Year break now ? If so please can I have a review of this PR.

tohojo:
We're not merging this. This repository is meant to show off best-practice use cases, and this is a horrible hack...

tjcw:
@tohojo Can you explain more about what is a horrible hack and how to correct it ? I think passing the first packet of a flow to userspace for a verdict on allow/deny is a reasonable thing to want to do; my team lead @huang195 and I hope eventually to build a Kubernetes CNI based on this approach.

huang195:
@tjcw First of all, at the very top of this PR, can you please give a brief context what this new example is suppose to show and also how it is different from the other AF_XDP examples that are already in the repo? Second, you should mark this as WIP. AFAIK, the tun/tap approach that you're currently pursuing is an intermediate solution while we are getting it to work with veth pairs, which is not working yet the last time I checked. Once it is working, we should remove all the tun/tap stuff from the code. 

Hopefully with all these things better explained, @tohojo can tell us if it is worth keeping this in the repo to help new eBPF developers to use AF_XDP.

tjcw:
@huang195 I have added a description at the top of the PR. I don't know how to add a work-in-progress label. This level is a working staging point towards a full implementation; my latest code is in branch https://github.com/tjcw/bpf-examples/tree/tjcw-explore-sameeth , I would like this PR to be integrated (after code revisions if necessary) to be something which I could rebase the tjcw-explore-sameeth branch on.

tohojo:
Chris Ward ***@***.***> writes:

> @tohojo Can you explain more about what is a horrible hack and how to
> correct it ?

Well, the fact that you had to mess with the rp_filter setting should be
a hint: just arbitrarily moving packets to a different interface is a
hack and it breaks assumptions of both the kernel and upper-layer
protocols (and your defending it with "well, TCP is just going to
retransmit" certainly didn't help build confidence that this is a
reasonable thing to do).

> I think passing the first packet of a flow to userspace for a verdict
> on allow/deny is a reasonable thing to want to do; my team lead
> @huang195 and I hope eventually to build a Kubernetes CNI based on
> this approach.

I've asked you repeatedly to actually qualify *why* that is a reasonable
thing to do, and you have failed to do so. I don't, in fact, think it's
a reasonable thing to do. Having a map-based flow lookup to skip a lot
of logic is fine, but that logic can be implemented just fine in XDP
itself, without incurring any of the problems that you're trying to
paper over with the hacky reinjection logic.

Ultimately, it boils down to a lack of synchronisation primitives:
there's no way to trigger the switch from redirecting to userspace to
passing them to the kernel that's not racy (i.e., packets going to
userspace can be overtaken by subsequent ones passed to the kernel right
after the switch happens). It may be doable after we add the ability to
buffer packets in the kernel XDP path, but, well, that's still a ways
of. And I still don't think that makes it a good system design...

Anyway, you're of course welcome to continue experimenting with this
despite my reservations above. But this repository is definitely not the
right home for such experiments, so please don't bother opening any more
pull requests for this stuff here...

One thing that could be useful as an example is how to use AF_XDP with a
custom XDP program. But that would be more suited as an addition to one
of the existing AF_XDP example programs. For instance, a minimal example
would be an operating mode that only passes traffic on a particular port
to userspace, by way of a custom XDP program.

Issue #78, opened 2022-12-16
AF_XDP-forwarding test case fails


tjcw:
I am working through the bpf-examples, and for me the AF_XDP-forwarding test case fails .
I have a fresh clone of xdp-project/bpf-examples and have just run the `configure` script and run `make` in the AF_XDP-forwarding directory. When I try to run the test case I get
```
tjcw@tjcw-host:~/workspace3/bpf-examples/AF_XDP-forwarding$ sudo ./xsk_fwd -i veth1 -q 0 -i veth2 -q 0 -c 0
[sudo] password for tjcw: 
Buffer pool created successfully.
Port 0 initialization failed.
tjcw@tjcw-host:~/workspace3/bpf-examples/AF_XDP-forwarding$
```
Am I doing something wrong, or is the test case broken ?

tohojo:
Doesn't work for me on veths either; no idea if it's supposed to. @magnus-karlsson would know, hopefully?

magnus-karlsson:
Seems to be broken. I have never tried it on veth interfaces. Will take a look at it.

tjcw:
When I run it under gdb, I see a bind syscall coming back with errno 95 ( EOPNOTSUPP )

tjcw:
Yes, that allows the test case to run. I have created a PR. (I haven't actually sent any packets over the veths; all I have checked is that the test case starts and initialises)

PR #79, opened 2022-12-16
Arrange for bind for eth to fall back to 'copy' if 'zerocopy' is not supported

tjcw:
Issue #78 exposes a problem in that eths which do not support 'zerocopy' cause socket initialization to fail. The fix for this is to drop XDP_ZEROCOPY from the bind flags; this causes libxdp to first try to set up zerocopy mode and it this fails to fall back to copying the packet data.

Signed-off-by: Chris Ward <tjcw@uk.ibm.com>

### DIFF
--- a/AF_XDP-forwarding/xsk_fwd.c
+++ b/AF_XDP-forwarding/xsk_fwd.c
@@ -728,7 +728,7 @@ static const struct port_params port_params_default = {
 		.tx_size = XSK_RING_PROD__DEFAULT_NUM_DESCS,
 		.libxdp_flags = 0,
 		.xdp_flags = XDP_FLAGS_DRV_MODE,
-		.bind_flags = XDP_USE_NEED_WAKEUP | XDP_ZEROCOPY,
+		.bind_flags = XDP_USE_NEED_WAKEUP,
 	},
 
 	.bp = NULL,


tohojo:
Please write a proper commit message; also, not sure if we should just unconditionally disable the flag, or if it should be configurable. @magnus-karlsson WDYT?

magnus-karlsson:
> Please write a proper commit message; also, not sure if we should just unconditionally disable the flag, or if it should be configurable. @magnus-karlsson WDYT?

It is fine to just remove it. In this case, the bind will first try zero-copy and if that is not available, fall back to copy mode.

tohojo:
> It is fine to just remove it. In this case, the bind will first try zero-copy and if that is not available, fall back to copy mode.

Alright, cool, thanks!

tohojo (changes requested):
The commit message still needs work:

- Drop the reference to the Github issue# form the title
- The title should describe what change is made, not the issue being fixed
- The textual description should then describe what issue is being fixed, and why that fix is OK (see Magnus' comment about falli-back).

tjcw:
@tohojo @magnus-karlsson I have made the requested changes to the PR. Can this be committed now ?

tjcw:
@magnus-karlsson in requesting a re-review from tohojo, I seem to have dropped you from the reviewer list and I cannot figure how to put you back on the list. Please review the PR anyway.

tohojo:
...you literally didn't fix any of the issues I pointed out above re: the commit message, and now you're requesting a new review? Are you deliberately trying to waste our time here? :/

tohojo:
Yeah, that's better, thanks!

PR #80, opened 2023-02-03
AF_XDP-example: fix XDP prog attach for xdp-skb mode

tirthendu-intel:
The attach mode is by default set to XDP_MODE_NATIVE and needs to be overwritten to XDP_MODE_SKB when '-S' option is used. Instead of overwriting the attach mode was ORed and so was always running in NATIVE mode. This patch fixes that.

Signed-off-by: Tirthendu Sarkar <tirthendu.sarkar@intel.com>


tohojo:
Thanks for the fix!

PR #82, opened 2023-03-19
preserve-dscp: update for strict section names

rpigott:
The old "classifier/xxx" format is no longer supported.


tohojo:
Thank you for the fix!

PR #83, opened 2023-03-22
configure: update libbpf check for modern api

rpigott:
The system libbpf check would unconditionally fail with a modern libbpf because the example program used outdated and removed library functions. Update the test program so that system libbpf check can pass again.


tohojo:
Thank you for the fix!

PR #91, opened 2023-07-06
Add systemd unit files for ePPing setup

simosund:
I do not have a lot of experience with systemd unit files, so this may have many flaws. Please let me know of any issues you spot. Based on some initial testing it does at least seem to work together with LibreQoS. I will test out the memory limit soon and add that as well.

The path to the bpf-examples folder will need to be configured based on where you install it, and you will also need to insert the correct interface to monitor.

Based on top of the aggregation PR because it is intended to use the aggregation mode.


simosund:
Changed my mind about basing it on top of the aggregation PR, as that just "pollutes" this PR with a bunch of unrelated commits. This will simply assume the existence of an aggregation mode.

pping/systemd-files/pping.service
```diff
@@ -0,0 +1,24 @@
...
+Wants=rotate-pping.service
+
+# Uncomment below if running with LibreQoS
+# After=lqosd.service
+
+[Service]
+# Rotate any old output first to ensure ePPing can start writing to its intended file
+ExecStartPre=/opt/bpf-examples/pping/scripts/rotate-pping-output.sh
+
+WorkingDirectory=/opt/src/bpf-examples/pping
```
tohojo, line 13: You left a `src` in here that probably shouldn't be there?

pping/scripts/rotate-pping-output.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+mkdir -p /var/log/pping
+
+newname="/var/log/pping/pping.out.json.$(date -Iseconds)"
+mv /var/log/pping/pping.out.json "$newname" 2> /dev/null
+
+if systemctl is-active --quiet pping.service; then
+    systemctl kill --signal=SIGHUP pping.service 2> /dev/null
```
tohojo, line 9: Hmm, why not just have an `ExecReload=kill -HUP $MAINPID` in the main service file and then issue a `systemctl reload` here?

pping/scripts/rotate-pping-output.sh
```diff
@@ -0,0 +1,16 @@
...
+    # Give ePPing some time to finish writing to the file
+    sleep 1
```
tohojo, line 11: This seems racy; how about checking with `fuser -s` before gzipping below instead (in a loop, with a max number of iterations)?

pping/scripts/rotate-pping-output.sh
```diff
@@ -0,0 +1,16 @@
...
+fi
+
+gzip "$newname" 2> /dev/null
```
tohojo, line 14: Maybe only do this if we actually moved a file? :)

pping/systemd-files/rotate-pping.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Rotate ePPing output every minute
+PartOf=pping.service rotate-pping.service
+
+[Timer]
+Unit=rotate-pping.service
+OnUnitActiveSec=60
```
tohojo, line 7: That is going to be a lot of log files :)

It may be that this is a reasonable interval, but then I think you should make the rotation script move them into a subfolder per day, just to make browsing the log directory more manageable...

pping/systemd-files/pping.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=ePPing - Passive monitoring of network RTTs
+After=network.service
+Wants=rotate-pping.service
```
tohojo, line 4: Hmm, so you're running the rotation script as a pre-command below, so I think this should just be starting the timer? Otherwise you'll get an immediate activation of the rotation script after starting up, which seems a bit odd?

pping/systemd-files/rotate-pping.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Rotates the output file from ePPing (pping.service)
+After=pping.service
+Wants=rotate-pping.timer
```
tohojo, line 4: I don't think this is necessary. Usually with timer you just start the timer and that will handle all interactions with the service.

pping/systemd-files/rotate-pping.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Rotate ePPing output every minute
+PartOf=pping.service rotate-pping.service
```
tohojo, line 3: Making this PartOf the main service is a neat idea, but I don't think you need the second bit (for the same reasons as above) :)

tohojo (changes requested):

pping/systemd-files/pping.service
simosund: That shouldn't be there, no. I'm not really sure if /opt/ is a good "default" path, but that's what LibreQoS uses in its installation instructions (when installing from source), so went with that. There's unfortunately quite a few paths the user will have to modify based on where the install bpf-examples.

pping/scripts/rotate-pping-output.sh
simosund: Because I didn't know you could do that? =)
But thanks for the tip, seems a bit cleaner to let the pping.service itself deal with the details of how it actually reloads.

pping/systemd-files/rotate-pping.timer
simosund: Yhea...I'm not really sure what the most sensible interval is. A while back we talked about it creating a separate file for every report, so compared to that it's at least a bit less. With the reporting period of 10 seconds that I'm using it means that each file will only contain reports from 6 times, but we could potentially have 16384 prefixes that is reported at each of those times, which would still be quite a bit of data then. But yhea, not cluttering up the "root" folder with thousands of files each day is probably a good idea.

pping/systemd-files/pping.service
simosund: Yhea, was (and to some degree still am) a bit unsure how the timers work exactly, and the whole relation thing between different units. But have changed it around a bit now so that the pping.service only starts the timer, and only the timer starts the rotate-pping.service. To get the timer to ever trigger I also needed to add an OnActiveSec (in addition to the OnActiveUnitSec), as otherwise it would never trigger the service.

simosund:
Not sure what's the best way to really set the memory limit. Using the MemoryHigh limit just seems to make the startup (when it allocates all the memory) incredibly slow, which I'm not sure is particularly useful. So just went with a MemoryMax limit instead, although that does not really allow ePPing to shut down cleanly. 

Usually that's not really a big problem as it allocates almost all the memory right at the startup, and will therefore most likely fail before it can attach any BPF programs. But in case it would run out of memory during runtime for some reason, e.g. due to a memory leak, the OOM-kill that is invoked will likely still leave the BPF programs around (and thereby also the maps that eat up most of the memory). Created a small script that should help with cleaning up and remaining tc BPF programs (skipped XDP for now as XDP won't be used for the LibreQoS deployment).

pping/systemd-files/pping.service
```diff
@@ -0,0 +1,34 @@
...
+Restart=on-failure
+RestartSec=60
+
+# Set to whatever maximum memory footprint you can tolerate
+# Note, hard-kills the ePPing process without allowing it to clean up. Consider
+# running /scripts/cleanup-tc-progs after.
+MemoryMax=4G
+
+[Install]
+Also=rotate-pping.timer
```
tohojo, line 30: When you have the PartOf statement in the timer unit I don't think you need this? It's enough to just have this unit enabled, and when it is started, the timer will start as well due to the PartOf - if I understand things correctly at least.

pping/systemd-files/pping.service
```diff
@@ -0,0 +1,34 @@
...
+
+# Uncomment to automatically start ePPing on reboot
+# WantedBy=default.target
```
tohojo, line 33: This you can leave uncommented by default. The [Install] section is only used when someone explicitly does a `systemctl enable` of a unit...

pping/systemd-files/rotate-pping.timer
```diff
@@ -0,0 +1,14 @@
...
+[Timer]
+Unit=rotate-pping.service
+
+# Rotate ePPing output every X seconds
+OnActiveSec=60
+OnUnitActiveSec=60
+
+[Install]
+WantedBy=pping.service
+Also=rotate-pping.service
```
tohojo, line 14: Since the timer is expected to be started by the PartOf statement, just leave out the [Install] section entirely. That will make it impossible to `systemctl enable` the timer, which is exactly what we want (since it's not supposed to be used by itself).

pping/systemd-files/pping.service
tohojo: Well, the default would be whatever is in the `make install` target ;)

For now, just leave it in /opt as these are mostly supposed to be example units, we can update things if/when we get around to making pping properly installable and packaging it up :)

pping/systemd-files/pping.service
```diff
@@ -0,0 +1,34 @@
...
+
+[Service]
+# Ensure folder for saving output to exists
+ExecStartPre=/usr/bin/mkdir -p /var/log/pping
+
+# Rotate any old output first to ensure ePPing can start writing to its intended file
+ExecStartPre=/opt/bpf-examples/pping/scripts/rotate-pping-output.sh /var/log/pping/pping.out.json
+
+WorkingDirectory=/opt/bpf-examples/pping
+ExecStart=/opt/bpf-examples/pping/pping -i <interface> -l -r 10 -a 10 -F json -w /var/log/pping/pping.out.json
```
tohojo, line 16: So one way to deal with the <interface> thing would be by making this a parameterised unit (so you'd do `systemctl start pping@eth0` for instance. You probably need to make the logging be separate per instance as well then, as in this case multiple instances can run simultaneously...

tohojo (changes requested):
A few more comments. I think using MemoryMax is fine; we're mostly just trying to make sure pping doesn't take the whole system with it if it gobbles up too much memory :)

pping/systemd-files/pping.service
simosund: Yhea, just me being confused about how you start up systemd services. Most examples tell you do to enable followed by start, so got the impression that you needed to "enable" the services before they could even be started. But now I've realised that "enable" is pretty much just "install", so if you don't need something to start up automatically at boot or something you can just start them up right away. So I've now removed the install sections from everything except this default target on the pping service.

pping/systemd-files/pping.service
simosund: Yhea, thought it looked a bit complicated at first, but turned out it was pretty straight forward. Just had to do some minor tweaks to get it working (mostly adding a bunch of %i everywhere and get the rotation script to reload the right instance). Now each interface gets its own subfolder in /var/log/pping.

simosund:
Thanks for the quick reviews. After some initial confusion I think I've started comming to grips with these systemd unit files now. Seems to work well to just systemctl start/stop pping@interface.service now. 

If I've understood it correctly you're also on vacation now. But in case it's a rainy day or you just happen to be bored, it would be awesome if you could take another look at this.

PR #100, opened 2023-09-11
Add additional counters to ePPing

simosund:
Add some additional packet counters to ePPing, to provide a more complete picture of the encountered traffic.

Overall, the plan is to:
- Have 3 sets of packet/byte counters on a per-prefix level: 
  - TCP with timestamp
  - TCP without timestamp
  - Other (non-TCP)
- Have global per-protocol packet counters
  - Will still likely have to go with some selection of common-ish protocols, as just going going with the value from the IP-header protocol/next-header field gives you some 145 actual protocols (and another 111 invalid/reserved ones). 
  - It's probably also good to have a counter for non-IP as well as non-Ethernet packets (these will not really be possible to add to the per-prefix stats, as they won't have a valid IP-address)
- Change current debug counters on map utilization to be included in normal periodical output
  - Potential concurrency challenge, map cleanup done in separate thread
- Change current "map-full" warnings to count number of missed entries (rather than give a sample entry once per second), and add them to the periodical map stats.
  - Currently don't provided "map stats" on the per-prefix aggregation maps. Could potentially add some in the code that periodically pulls and reports the aggregated stats.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -121,6 +121,12 @@ struct aggregation_maps {
 	int map_active_fd;
 	int map_v4_fd[2];
 	int map_v6_fd[2];
+	int map_globcnt_fd;
+};
+
+struct aggregation_context {
+	struct aggregation_maps maps;
+	struct global_counters prev_counters;
 };
 
 // Store configuration values in struct to easily pass around
@@ -130,7 +136,7 @@ struct pping_config {
 	struct bpf_tc_opts tc_egress_opts;
 	struct map_cleanup_args clean_args;
 	struct aggregation_config agg_conf;
-	struct aggregation_maps agg_maps;
+	struct aggregation_context agg_ctx;
 	struct output_context *out_ctx;
 	char *object_path;
 	char *ingress_prog;
@@ -1136,6 +1142,233 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
 	fprintf(stderr, "Lost %llu events on CPU %d\n", lost_cnt, cpu);
 }
 
+static void print_counter_standard(FILE *stream, const char *name, __u64 val,
+				   bool *first)
+{
+	if (val == 0)
+		return;
+
+	if (*first)
+		*first = false;
+	else
+		fprintf(stream, ", ");
+
+	fprintf(stream, "%s=%llu", name, val);
+}
+
+static void print_pktbytes_tuple_standard(FILE *stream, const char *tuple_name,
+					  __u64 pkts, __u64 bytes,
+					  bool *first_tuple)
+{
+	bool first = true;
+
+	if (pkts == 0)
+		return;
+
+	if (*first_tuple)
+		*first_tuple = false;
+	else
+		fprintf(stream, ", ");
+
+	fprintf(stream, "%s=(", tuple_name);
+	print_counter_standard(stream, "pkts", pkts, &first);
+	print_counter_standard(stream, "bytes", bytes, &first);
+	fprintf(stream, ")");
+}
+
+static void
+print_globalcounters_standard(FILE *stream, __u64 t_monotonic,
+			      const struct global_counters *counters)
+{
+	char protostr[16];
+	bool first = true;
+	int proto;
+
+	print_ns_datetime(stream, t_monotonic);
+	fprintf(stream, ": ");
+
+	print_pktbytes_tuple_standard(stream, "non-IP", counters->nonip_pkts,
+				      counters->nonip_bytes, &first);
+	print_pktbytes_tuple_standard(stream, "TCP", counters->tcp_pkts,
+				      counters->tcp_bytes, &first);
+	print_pktbytes_tuple_standard(stream, "UDP", counters->udp_pkts,
+				      counters->udp_bytes, &first);
+	print_pktbytes_tuple_standard(stream, "ICMP", counters->icmp_pkts,
+				      counters->icmp_bytes, &first);
+	print_pktbytes_tuple_standard(stream, "ICMPv6", counters->icmp6_pkts,
+				      counters->icmp6_bytes, &first);
+
+	for (proto = 0; proto < N_IPPROTOS; proto++) {
+		if (counters->other_ipprotos[proto] > 0) { // Check to avoid uncecessary ipproto_to_str
+			ipproto_to_str(protostr, sizeof(protostr), proto);
+			print_pktbytes_tuple_standard(
+				stream, protostr,
+				counters->other_ipprotos[proto], 0, &first);
+		}
+	}
+
+	if (first) // Global counters are empty
+		fprintf(stream, "pkts=0, bytes=0");
+
+	fprintf(stream, "\n");
+}
+
+static void print_counter_json(json_writer_t *jctx, const char *name, __u64 val)
+{
+	if (val > 0)
+		jsonw_u64_field(jctx, name, val);
+}
+
+static void print_pktbytes_tuple_json(json_writer_t *jctx,
+				      const char *tuple_name, __u64 pkts,
+				      __u64 bytes)
+{
+	// Don't include empty pkt/byte tuples
+	if (pkts == 0)
+		return;
+
+	jsonw_name(jctx, tuple_name);
+	jsonw_start_object(jctx);
+	print_counter_json(jctx, "packets", pkts);
+	print_counter_json(jctx, "bytes", bytes);
+	jsonw_end_object(jctx);
+}
+
+static void print_globalcounters_json(json_writer_t *jctx, __u64 t_monotonic,
+				      const struct global_counters *counters)
+{
+	char protostr[16];
+	int proto;
+
+	jsonw_start_object(jctx);
+	jsonw_u64_field(jctx, "timestamp",
+			convert_monotonic_to_realtime(t_monotonic));
+
+	jsonw_name(jctx, "protocol_counters");
+	jsonw_start_object(jctx);
+
+	print_pktbytes_tuple_json(jctx, "non_IP", counters->nonip_pkts,
+				  counters->nonip_bytes);
+	print_pktbytes_tuple_json(jctx, "TCP", counters->tcp_pkts,
+				  counters->tcp_bytes);
+	print_pktbytes_tuple_json(jctx, "UDP", counters->udp_pkts,
+				  counters->udp_bytes);
+	print_pktbytes_tuple_json(jctx, "ICMP", counters->icmp_pkts,
+				  counters->icmp_bytes);
+	print_pktbytes_tuple_json(jctx, "ICMPv6", counters->icmp6_pkts,
+				  counters->icmp6_bytes);
+
+	for (proto = 0; proto < N_IPPROTOS; proto++) {
+		if (counters->other_ipprotos[proto] > 0) {
+			ipproto_to_str(protostr, sizeof(protostr), proto);
+			print_pktbytes_tuple_json(
+				jctx, protostr, counters->other_ipprotos[proto],
+				0);
+		}
+	}
+
+	jsonw_end_object(jctx);
+
+	jsonw_end_object(jctx);
+}
+
+static void print_globalcounters(struct output_context *out_ctx,
+				 __u64 t_monotonic,
+				 const struct global_counters *counters)
+{
+	if (out_ctx->format == PPING_OUTPUT_STANDARD)
+		print_globalcounters_standard(out_ctx->stream, t_monotonic,
+					      counters);
+	else if (out_ctx->jctx)
+		print_globalcounters_json(out_ctx->jctx, t_monotonic, counters);
+}
+
+static void update_globalcounters(struct global_counters *to,
+				  const struct global_counters *from)
+{
+	int proto;
+
+	to->nonip_pkts += from->nonip_pkts;
+	to->nonip_bytes += from->nonip_bytes;
+	to->tcp_pkts += from->tcp_pkts;
+	to->tcp_bytes += from->tcp_bytes;
+	to->udp_pkts += from->udp_pkts;
+	to->udp_bytes += from->udp_bytes;
+	to->icmp_pkts += from->icmp_pkts;
+	to->icmp_bytes += from->icmp_bytes;
+	to->icmp6_pkts += from->icmp6_pkts;
+	to->icmp6_bytes += from->icmp6_bytes;
+
+	for (proto = 0; proto < N_IPPROTOS; proto++) {
+		to->other_ipprotos[proto] += from->other_ipprotos[proto];
+	}
+}
+
+static void merge_percpu_globalcounters(struct global_counters *merged,
+					const struct global_counters *percpu,
+					int n_cpus)
+{
+	int cpu;
+
+	memset(merged, 0, sizeof(*merged));
+
+	for (cpu = 0; cpu < n_cpus; cpu++) {
+		update_globalcounters(merged, &percpu[cpu]);
+	}
+}
+
+static void diff_globalcounters(struct global_counters *diff,
+				const struct global_counters *prev,
+				const struct global_counters *next)
+{
+	int proto;
+
+	diff->nonip_pkts = next->nonip_pkts - prev->nonip_pkts;
+	diff->nonip_bytes = next->nonip_bytes - prev->nonip_bytes;
+	diff->tcp_pkts = next->tcp_pkts - prev->tcp_pkts;
+	diff->tcp_bytes = next->tcp_bytes - prev->tcp_bytes;
+	diff->udp_pkts = next->udp_pkts - prev->udp_pkts;
+	diff->udp_bytes = next->udp_bytes - prev->udp_bytes;
+	diff->icmp_pkts = next->icmp_pkts - prev->icmp_pkts;
+	diff->icmp_bytes = next->icmp_bytes - prev->icmp_bytes;
+	diff->icmp6_pkts = next->icmp6_pkts - prev->icmp6_pkts;
+	diff->icmp6_bytes = next->icmp6_bytes - prev->icmp6_bytes;
+
+	for (proto = 0; proto < N_IPPROTOS; proto++) {
+		diff->other_ipprotos[proto] = next->other_ipprotos[proto] -
+					      prev->other_ipprotos[proto];
+	}
+}
+
+static int report_globalcounters(struct output_context *out_ctx,
+				 struct aggregation_context *agg_ctx)
+{
+	int n_cpus = libbpf_num_possible_cpus();
+	__u64 t = get_time_ns(CLOCK_MONOTONIC);
+	struct global_counters tot_cnt, diff;
+	struct global_counters *cpu_cnt;
+	__u32 key = 0;
+	int err;
+
+	cpu_cnt = calloc(n_cpus, sizeof(*cpu_cnt));
+	if (!cpu_cnt)
+		return -errno;
+
+	err = bpf_map_lookup_elem(agg_ctx->maps.map_globcnt_fd, &key, cpu_cnt);
+	if (err)
+		goto exit;
+
+	merge_percpu_globalcounters(&tot_cnt, cpu_cnt, n_cpus);
+	diff_globalcounters(&diff, &agg_ctx->prev_counters, &tot_cnt);
+	agg_ctx->prev_counters = tot_cnt;
+
+	print_globalcounters(out_ctx, t, &diff);
+
+exit:
+	free(cpu_cnt);
+	return err;
+}
+
 static __u64 sum_trafficcounts_pkts(const struct traffic_counters *counters)
 {
 	return counters->tcp_ts_pkts + counters->tcp_nots_pkts +
@@ -1296,21 +1529,6 @@ static void print_aggstats_standard(FILE *stream, __u64 t,
 
 exit:
 	fprintf(stream, "\n");
-}
-
-static void print_pktbytes_tuple_json(json_writer_t *jctx,
-				      const char *tuple_name, __u64 pkts,
-				      __u64 bytes)
-{
-	// Don't include empty pkt/byte tuples
-	if (pkts == 0)
-		return;
-
-	jsonw_name(jctx, tuple_name);
-	jsonw_start_object(jctx);
-	jsonw_u64_field(jctx, "packets", pkts);
-	jsonw_u64_field(jctx, "bytes", bytes);
-	jsonw_end_object(jctx);
 }
 
 static void print_trafficcount_json(json_writer_t *jctx,
@@ -1535,25 +1753,31 @@ exit:
 }
 
 static int report_aggregated_stats(struct output_context *out_ctx,
-				   struct aggregation_maps *maps,
+				   struct aggregation_context *agg_ctx,
 				   struct aggregation_config *agg_conf)
 {
 	__u64 t = get_time_ns(CLOCK_MONOTONIC);
 	int err, map_idx;
 
-	map_idx = switch_agg_map(maps->map_active_fd);
+	map_idx = switch_agg_map(agg_ctx->maps.map_active_fd);
 	if (map_idx < 0)
 		return map_idx;
 
-	err = report_aggregated_stats_map(out_ctx, maps->map_v4_fd[map_idx],
+	err = report_aggregated_stats_map(out_ctx,
+					  agg_ctx->maps.map_v4_fd[map_idx],
 					  AF_INET, agg_conf->ipv4_prefix_len, t,
 					  agg_conf);
 	if (err)
 		return err;
 
-	err = report_aggregated_stats_map(out_ctx, maps->map_v6_fd[map_idx],
+	err = report_aggregated_stats_map(out_ctx,
+					  agg_ctx->maps.map_v6_fd[map_idx],
 					  AF_INET6, agg_conf->ipv6_prefix_len,
 					  t, agg_conf);
+	if (err)
+		return err;
+
+	err = report_globalcounters(out_ctx, agg_ctx);
 	return err;
 }
 
@@ -1921,15 +2145,17 @@ int fetch_aggregation_map_fds(struct bpf_object *obj,
 		bpf_object__find_map_fd_by_name(obj, "map_v6_agg1");
 	maps->map_v6_fd[1] =
 		bpf_object__find_map_fd_by_name(obj, "map_v6_agg2");
+	maps->map_globcnt_fd =
+		bpf_object__find_map_fd_by_name(obj, "map_global_counters");
 
 	if (maps->map_active_fd < 0 || maps->map_v4_fd[0] < 0 ||
 	    maps->map_v4_fd[1] < 0 || maps->map_v6_fd[0] < 0 ||
-	    maps->map_v6_fd[1] < 0) {
+	    maps->map_v6_fd[1] < 0 || maps->map_globcnt_fd < 0) {
 		fprintf(stderr,
-			"Unable to find aggregation maps (%d/%d/%d/%d/%d).\n",
+			"Unable to find aggregation maps (%d/%d/%d/%d/%d/%d).\n",
 			maps->map_active_fd, maps->map_v4_fd[0],
 			maps->map_v4_fd[1], maps->map_v6_fd[0],
-			maps->map_v6_fd[1]);
+			maps->map_v6_fd[1], maps->map_globcnt_fd);
 		return -ENOENT;
 	}
 
@@ -2006,14 +2232,17 @@ static int init_aggregation_timer(struct bpf_object *obj,
 {
 	int err, fd;
 
-	err = fetch_aggregation_map_fds(obj, &config->agg_maps);
+	memset(&config->agg_ctx.prev_counters, 0,
+	       sizeof(config->agg_ctx.prev_counters));
+
+	err = fetch_aggregation_map_fds(obj, &config->agg_ctx.maps);
 	if (err) {
 		fprintf(stderr, "Failed fetching aggregation maps: %s\n",
 			get_libbpf_strerror(err));
 		return err;
 	}
 
-	err = init_agg_backup_entries(&config->agg_maps,
+	err = init_agg_backup_entries(&config->agg_ctx.maps,
 				      config->agg_conf.ipv4_prefix_len > 0,
 				      config->agg_conf.ipv6_prefix_len > 0);
 	if (err) {
@@ -2037,7 +2266,7 @@ static int init_aggregation_timer(struct bpf_object *obj,
 
 static int handle_aggregation_timer(int timer_fd,
 				    struct output_context *out_ctx,
-				    struct aggregation_maps *maps,
+				    struct aggregation_context *agg_ctx,
 				    struct aggregation_config *agg_conf)
 {
 	__u64 timer_exps;
@@ -2055,7 +2284,7 @@ static int handle_aggregation_timer(int timer_fd,
 			timer_exps - 1);
 	}
 
-	err = report_aggregated_stats(out_ctx, maps, agg_conf);
+	err = report_aggregated_stats(out_ctx, agg_ctx, agg_conf);
 	if (err) {
 		fprintf(stderr, "Failed reporting aggregated RTTs: %s\n",
 			get_libbpf_strerror(err));
@@ -2162,7 +2391,7 @@ static int epoll_poll_events(int epfd, struct pping_config *config,
 		case PPING_EPEVENT_TYPE_AGGTIMER:
 			err = handle_aggregation_timer(
 				events[i].data.u64 & PPING_EPEVENT_MASK,
-				config->out_ctx, &config->agg_maps,
+				config->out_ctx, &config->agg_ctx,
 				&config->agg_conf);
 			break;
 		case PPING_EPEVENT_TYPE_SIGNAL:
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -1149,6 +1149,12 @@ static bool ecncounters_empty(const struct ecn_counters *counters)
 	return memcmp(counters, &empty, sizeof(empty)) == 0;
 }
 
+static bool pping_errors_empty(const struct pping_error_counters *errors)
+{
+	static const struct pping_error_counters empty = { 0 };
+
+	return memcmp(errors, &empty, sizeof(empty)) == 0;
+}
 
 static void print_counter_standard(FILE *stream, const char *name, __u64 val,
 				   bool *first)
@@ -1198,6 +1204,22 @@ static void print_ecncounters_standard(FILE *stream,
 }
 
 static void
+print_ppingerrors_standard(FILE *stream,
+			   const struct pping_error_counters *errors)
+{
+	bool first = true;
+
+	fprintf(stream, "errors=(");
+	print_counter_standard(stream, "store-packet-ts", errors->pktts_store,
+			       &first);
+	print_counter_standard(stream, "create-flow-state", errors->flow_create,
+			       &first);
+	print_counter_standard(stream, "create-agg-subnet-state",
+			       errors->agg_subnet_create, &first);
+	fprintf(stream, ")");
+}
+
+static void
 print_globalcounters_standard(FILE *stream, __u64 t_monotonic,
 			      const struct global_counters *counters)
 {
@@ -1236,6 +1258,11 @@ print_globalcounters_standard(FILE *stream, __u64 t_monotonic,
 		print_ecncounters_standard(stream, &counters->ecn);
 	}
 
+	if (!pping_errors_empty(&counters->err)) {
+		fprintf(stream, ", ");
+		print_ppingerrors_standard(stream, &counters->err);
+	}
+
 	fprintf(stream, "\n");
 }
 
@@ -1268,6 +1295,17 @@ static void print_ecncounters_json(json_writer_t *jctx,
 	print_counter_json(jctx, "ECT1", counters->ect1);
 	print_counter_json(jctx, "ECT0", counters->ect0);
 	print_counter_json(jctx, "CE", counters->ce);
+	jsonw_end_object(jctx);
+}
+
+static void print_ppingerrors_json(json_writer_t *jctx,
+				   const struct pping_error_counters *errors)
+{
+	jsonw_start_object(jctx);
+	print_counter_json(jctx, "store_packet_ts", errors->pktts_store);
+	print_counter_json(jctx, "create_flow_state", errors->flow_create);
+	print_counter_json(jctx, "create_agg_subnet_state",
+			   errors->agg_subnet_create);
 	jsonw_end_object(jctx);
 }
 
@@ -1309,6 +1347,9 @@ static void print_globalcounters_json(json_writer_t *jctx, __u64 t_monotonic,
 	jsonw_name(jctx, "ecn_counters");
 	print_ecncounters_json(jctx, &counters->ecn);
 
+	jsonw_name(jctx, "errors");
+	print_ppingerrors_json(jctx, &counters->err);
+
 	jsonw_end_object(jctx);
 }
 
@@ -1332,6 +1373,14 @@ static void update_ecncounters(struct ecn_counters *to,
 	to->ce += from->ce;
 }
 
+static void update_pping_errors(struct pping_error_counters *to,
+				const struct pping_error_counters *from)
+{
+	to->pktts_store += from->pktts_store;
+	to->flow_create += from->flow_create;
+	to->agg_subnet_create += from->agg_subnet_create;
+}
+
 static void update_globalcounters(struct global_counters *to,
 				  const struct global_counters *from)
 {
@@ -1353,6 +1402,7 @@ static void update_globalcounters(struct global_counters *to,
 	}
 
 	update_ecncounters(&to->ecn, &from->ecn);
+	update_pping_errors(&to->err, &from->err);
 }
 
 static void merge_percpu_globalcounters(struct global_counters *merged,
@@ -1378,6 +1428,16 @@ static void diff_ecncounters(struct ecn_counters *diff,
 	diff->ce = next->ce - prev->ce;
 }
 
+static void diff_pping_errors(struct pping_error_counters *diff,
+			      const struct pping_error_counters *prev,
+			      const struct pping_error_counters *next)
+{
+	diff->pktts_store = next->pktts_store - prev->pktts_store;
+	diff->flow_create = next->flow_create - prev->flow_create;
+	diff->agg_subnet_create =
+		next->agg_subnet_create - prev->agg_subnet_create;
+}
+
 static void diff_globalcounters(struct global_counters *diff,
 				const struct global_counters *prev,
 				const struct global_counters *next)
@@ -1401,6 +1461,7 @@ static void diff_globalcounters(struct global_counters *diff,
 	}
 
 	diff_ecncounters(&diff->ecn, &prev->ecn, &next->ecn);
+	diff_pping_errors(&diff->err, &prev->err, &next->err);
 }
 
 static int report_globalcounters(struct output_context *out_ctx,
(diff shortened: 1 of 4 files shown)


simosund:
Hi @tohojo, or anyone else who wants to provide some input. This PR is not ready for review, as I have yet to fix the map utalization stats as described above. But thought I would offer an opportunity of a "pre-review" of my plans to potentially save us both some work before I go ahead and implement something stupid.

So, the map utilization stats you can currently by compiling with the DEBUG are: 
- number of cleanup cycles 
- number of entries in the map (that the cleanup process looped through)
- runtime (for the periodical cleaning task)
- number of entries automatically deleted by the BPF programs (TSecr match or closed flow after seeing FIN/RST)
- number of entries that were timed out by the cleanup process. 

This output is synchronized with the periodical map cleanup process itself, so you get the stats once every time the cleanup runs. However, if we're gonna make this a proper feature (rather than something for internal debugging) I'm not sure if this is ideal, and perhaps it's better to let the user specify an arbitrary interval to get the map utilization stats.

Pros with separating the map utilization stats from the periodical cleanup process:
- More customizable as rate of output it not tied to how frequently the maps are cleaned
- May be easier to integrate stats on the prefix-maps, as those are not handled by the periodical cleanup and are instead cleared by the periodical aggregation reporting (which runs in the same thread like all the other reporting, unlike the periodical cleanup which runs in a separate thread). 
- May require less synchronization between the "main"/reporting thread and the periodical cleanup thread. Note that the current DEBUG output actually avoids explicit synchronization between the threads by having the BPF-iter programs that the periodical cleanup thread invokes send a message with the stats through the perf-buffer which the main-thread picks up and reports (although this message risks getting lost if a lot of individual RTTs are also sent over the perf-buffer).
- Some of the stats are or can be updated live by the BPF programs, so they're not inherently tied to cleanup process. 
 
Cons: 
- Will require a bit more rework of the current DEBUG code to get to work (although depending on if you think the current strategy of sending these stats over the perf-buffer as well as the current set of metrics, the code may have to be reworked a fair bit anyways)
- The number of entries in the map is currently determined by the cleanup process itself, as it simply counts how many entries it loops through. So to report the size at an arbitrary time that does not coincide with the cleanup process, we can either simply provide the number of entries seen at the last cleanup, or perhaps better add some counter for number of current entries which the BPF programs increment/decrement whenever they add/delete an entry. 
- Some of the stats are inherently tied to the cleanup process (ex. cleanup runtime and number of entries timed out), and will thus only get updated when the cleanup process runs anyways.

tohojo:
I think your plan sounds reasonable, and I agree that the stats
reporting shouldn't be tied to the cleanup process. For counting the
number of entries I think it's better to increment/decrement counters
when entries are added/deleted; as for cleanup stats, just updating the
counters whenever you run the cleanup is fine. If someone runs the
reporting more often than the cleanup they'll just get the same numbers
twice which is consistent with other such stats (e.g., if you run
reporting twice with no packets being processed in-between, you'll also
get the same numbers twice).

simosund:
Hi @tohojo, I think this PR is ready for review now.

There is one higher-level aspect which I'm not really sure on, and that's how to report the map utilization stats for the aggregation maps. I currently more or less try to hide the fact that there are two different instances of the aggregation maps (the active and inactive one, which are switched every time the aggregated stats are reported), by only reporting the currently active aggregation map (and use the same name regardless of which instance it is). This way it may look like there's only one instance of the aggregation map, but as there are actually two different instances the stats may become a bit confusing as they don't necessarily line up between reports (e.g. it could say it created 1000 entries in one report, and another 1000 entries in the next report (without deleting any), and still only report a total of 1000 entries in the map). 

I'm now starting to think that this was probably a bad idea, and that it would probably be better to just report both instances of the aggregation maps (so all 4 in total, 2 for IPv4 and 2 for IPv6), even if it means exposing the user to some implementation details (to make sense of these stats you need to be somewhat aware of the implementation details anyways). Especially as you may chose different intervals for the aggregation and map utilization reporting, so that they may not align very well. Let me know what you think, and if you want I can change this behavior before you do a proper review of the code (should be relatively minor code changes).

If you against all don't find anything that requires changes, please don't merge it quite yet as I also want to run a couple of performance tests to see how these changes impact performance (especially for traffic without TCP timestamps, as that's where I expect the largest relative increase in overhead costs as those packets are now added to the per-prefix aggregated stats).

simosund:
Oh, and in my recent force-push I just fixed an overflow bug which could cause the per-prefix reported byte counts to become waaaay to large. Specifically in the `sum_agg_pktcnts_bytes()` and `sum_agg_pktcnts_packets()` I for some reason used an `int` to sum a bunch of `__u64` counters together (and then return the result as a `__u64` again), so just changed `sum` to be a `__u64` as it ought to be.

sferlin:
I think these ^ (with additional tests to confirm your hypothesis of the impact) are very good and nice results. Very good for the analysis adding the counters and nice because, a good sign, the penalty is consistent across the #flows i.e. you do not see any strange deviation from the course of the curve. Do you see the consistency in the penalty with Ipv6 vs Ipv4 as well?

As for iperf3 with UDP I think it has to do TCP first to punch holes in firewalls. You might see this short-lived TCP disappear and UDP following on the same port after that(?)

simosund:
Did a couple of additional experiments to get a better picture of how much the different features in this PR contribute to the additional overhead. I have removed the raw forwarding baseline (as it can be seen in previous plots) to better highlight the differences between different configurations of ePPing. So I now have the following setups:
- Before this update, when aggregating per-prefix once per second (pre a1)
- After this update, when aggregating per-prefix once per second (post a1)
- As above, but in addition also reporting global counts of IP-protocols once per second (post a1 g1)
- As above, but in addition also reporting map utilization stats once per second (post a1 g1 m1)

So for TCP traffic it looks like this: 
epping_agg_performance_tcp_nobaseline

I did not really expect performance when just reporting the aggregated per-prefix stats to decrease (for TCP traffic). It seems strange that the small amount of additional logic to determine which of 3 different sets of counters should be incremented has a noticeable effect. I suspect it might be related to how I moved a large data structure off from the stack and into a per-CPU map to avoid exceeding the 512-byte stack limit (see 0bed749). The global counters do not seem to add much additional overhead at all, while the map utilization stats do add some (also somewhat surprising considering that the global counters have to be updated for every packet, whereas the map utilization stats are only updated when map entries are created/deleted).

For UDP it looks like this:
epping_agg_performance_udp_nobaseline
For UDP it is not surprising that the performance when just aggregating per-prefix drops a lot, as UDP traffic was not included in these per-prefix stats before. Once again global counters seem to have virtually no impact, while map utilization has a slight increase in overhead (and now the map utilization stats should not really need to be updated at all...so guess it's maybe just due to additional overhead from userspace reporting?).

Should note that these results are based on just 3 runs (although each run lasts for 100 seconds, so each point in these graphs basically shows the average across a total of 5 minutes, and the errorbars show standard deviation rather than confidence interval), and there is a little variation between each run, so take them with a grain of salt.

simosund:
Converted to draft while I update the code to report stats for both instances of the aggregation maps. Once that is done I will mark this PR as ready for review again.

After with some discussion with Anna today I've decided that I will probably also add some counters for ECN to the global counters before we start another run with Robert, so that you get the total counts for packets with the 4 different ECN code points (00, 01, 10 and 11). But that I can potentially do in a separate PR (although should be relatively straight forward, so will probably just be a single commit).

pping/pping_kern.c
```diff
@@ -1340,6 +1373,35 @@ static void send_flow_timeout_message(void *ctx, struct network_tuple *flow,
 	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &fe, sizeof(fe));
 }
 
+static void mapiter_update_maputil_stats(enum pping_map map, __u64 seq,
+					 bool finished, __u64 time)
+{
+	/* The map iteration always seems to end with a final "empty" iteration
+	   (no key/value), i.e. the "finished" round. This final round seems to
```
tohojo, line 1366: This seems to be deliberate, to allow the iterator to do any final summation actions at the end of an iteration. However, the cleanup you're doing in these iterators don't really need to do that, so why not just have them exit early if key or value are zero, instead of this hack? You even have such an early return in `flowmap_cleanup()`, it's just not early enough :)

pping/pping.c
```diff
@@ -180,6 +195,7 @@ static const struct option long_options[] = {
 	{ "aggregate-timeout",    required_argument, NULL, AGG_ARG_TIMEOUT }, // Interval for timing out subnet entries in seconds (default 30s)
 	{ "write",                required_argument, NULL, 'w' }, // Write output to file (instead of stdout)
 	{ "global-counters",      required_argument, NULL, 'g' }, // Output global packets counters every X seconds
+	{ "map-stats",            required_argument, NULL,  ARG_MAP_STATS }, // Output map utilization statistics every X seconds
```
tohojo, line 200: Same comment here ; none of the map utilisation stats are collected in the fast path, so do we really need to make the collection optional? I'd just only output them in JSON mode so they don't clutter up the normal interactive output.

pping/pping.c
```diff
@@ -1554,6 +1559,14 @@ static int report_aggregated_stats_map(struct output_context *out_ctx,
 		err = bpf_map_delete_batch(map_fd, del_keys, &del_idx,
 					   &batch_opts);
 
+	if (maputil) {
```
tohojo, line 1550: AFAICT, this is never NULL, so why the check?

pping/pping.c
```diff
@@ -1563,24 +1576,36 @@ static int report_aggregated_stats_map(struct output_context *out_ctx,
...
 {
+	struct map_util_stats *ipv4_util = NULL, *ipv6_util = NULL;
 	__u64 t = get_time_ns(CLOCK_MONOTONIC);
 	int err, map_idx;
 
 	map_idx = switch_agg_map(maps->map_active_fd);
 	if (map_idx < 0)
 		return map_idx;
 
+	if (maputil_ctx) {
```
tohojo, line 1578: Same here, this is never NULL

pping/pping.c
```diff
@@ -1342,13 +1414,13 @@ static int switch_agg_map(int map_active_fd)
 	return prev_map;
 }
 
-static void report_aggregated_rtt_mapentry(
+static void report_aggregated_stats_mapentry(
 	struct output_context *out_ctx, struct ipprefix_key *prefix,
-	struct aggregated_rtt_stats *percpu_stats, int n_cpus, int af,
+	struct aggregated_stats *percpu_stats, int n_cpus, int af,
 	__u8 prefix_len, __u64 t_monotonic, struct aggregation_config *agg_conf,
 	bool *del_entry)
```
tohojo, line 1448: Every time you write a function with 9 parameters it's time to take a good hard look in the mirror and think about whether the complexity of your code is getting out of hand ;)

For this one, for instance:
- del_entry could be a return value
- n_cpus doesn't really need to be passed in (just call libbpf_num_possible_cpus() inside the function)
- prefix len is already storend inside agg_conf, and you're passing in the af

Probable better to do a separate cleanup PR afterwards, but just wanted to point this out now since you're touching this code; but it would be good to try to do a refactor pass with a look to decreasing complexity and improving readability.

tohojo (changes requested):
I think there's probably enough for you to chew on in this one now. Note that I reviewed each commit separately, but I'm not sure github will present comments in that order (have I mentioned that I dislike the github reviewing interface immensely? ;)), so any comments that refer back to each other may be out of order...

pping/pping.c
```diff
@@ -1119,59 +1119,96 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
...
 
-static __u64 aggregated_rtt_stats_maxbins(struct aggregated_rtt_stats *stats,
-					  __u64 bin_width, __u64 n_bins)
+static bool aggregated_stats_empty(struct aggregated_stats *stats)
 {
-	return stats->max / bin_width < n_bins ? stats->max / bin_width + 1 :
-						 n_bins;
+	return sum_agg_pktcnts_pkts(stats, true) +
+		       sum_agg_pktcnts_pkts(stats, false) ==
+	       0;
```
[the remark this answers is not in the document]
simosund: Comparing an entire `aggregated_stats` struct won't work because of the `last_updated` member. But if I refactor as you suggest below (as it shows up on the github interface at least) comparing against just the `packet_counters` structs should work quite well.

pping/pping_kern.c
```diff
@@ -1121,28 +1134,43 @@ static void pping_match_packet(struct flow_state *f_state, void *ctx,
...
+{
+	if (!stats)
+		return;
+
+	int dir_idx = as_tx ? TX_IDX : RX_IDX;
+	int group_idx =
+		p_info->pid.flow.proto == IPPROTO_TCP && p_info->rtt_trackable ?
+			AGG_PKTCNT_TCPTS :
+		p_info->pid.flow.proto == IPPROTO_TCP ? AGG_PKTCNT_TCPNOTS :
+							AGG_PKTCNT_OTHER;
```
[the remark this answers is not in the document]
simosund: Talking of rx and tx, I now realize that it's not entierly clear which is which in this context. I've seen it from the perspective of the subnet, i.e. what's reported as rx is packets received by the subnet (packet dst match subnet). But perhaps it makes more sense to report it from the perspective of the capture point, i.e. rx would be packets the capture point has "received" from the subnet (packet src match subnet). I guess the latter would be more consitent with how the RTTs for the subnet are from the capture point's perspective, i.e. RTT for packets going to (and ACKs comming back from) that subnet (unless you use the `--aggregate-reverse` flag).

What do you think? As we're changing the data format anyways by adding the different packet counters, now would probably be the time to change the meaning of rx/tx. Or perhaps that should be called something else entierly? To and from?

pping/pping.h
```diff
@@ -272,4 +275,9 @@ struct aggregated_stats {
 	__u32 rtt_bins[RTT_AGG_NR_BINS];
 };
 
+struct global_packet_counters {
+	struct packet_counters non_ip;
+	struct packet_counters ip_protos[GLOBCOUNT_N_IPPROTS];
```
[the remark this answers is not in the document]
simosund: Didn't think of it from a caching perspective (as from a memory perspective I don't consider 4k problematic considering we're spending 4G on the aggregation maps).

I'm not sure I quite follow your idea for having an array of packet counts for the esoteric protocols. I take it you don't mean an full 256-member array as now, because then that would obviously waste just as much space (although be nicer to the cache by simply not using it for the common protocols). So do you mean a much smaller array for some selection of esoteric protocols (which we must then remap from their IP-protocol number to a more compact index). Or just have an array with space for X ( X << 256) protocols and then create the protocol <-> index mapping dynamically?

As for using u32 counters I assume you mean only for the esoteric protocols? As with ~1Mpps a u32 would overflow in slightly over an hour for the packet-count, and potentially just a few seconds for the byte count, and we never reset these counters.

pping/pping.c
```diff
@@ -174,6 +179,7 @@ static const struct option long_options[] = {
 	{ "aggregate-reverse",    no_argument,       NULL, ARG_AGG_REVERSE }, // Aggregate RTTs by dst IP of reply packet (instead of src like default)
 	{ "aggregate-timeout",    required_argument, NULL, AGG_ARG_TIMEOUT }, // Interval for timing out subnet entries in seconds (default 30s)
 	{ "write",                required_argument, NULL, 'w' }, // Write output to file (instead of stdout)
+	{ "global-counters",      required_argument, NULL, 'g' }, // Output global packets counters every X seconds
```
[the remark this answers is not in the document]
simosund: Because the user may not care for these stats, which is then just bloating their output. And unless we add some default reporting interval, they're gonna have to supply the option in order to specify the interval anyways. 

One option could be to use the same reporting interval as the aggregation, but what to do for non-aggregated RTT-reporting then? Not allow it?

As it implementation wise is independent of any other type of reporting, I didn't see any clear reason to restrict it. So I guess my counter question is, why not make it configurable?

pping/pping.c
simosund: Well, I would expect that most users wouldn't want these stats, they're replacing having to compile the program with a DEBUG flag. And with the change to how both instances of the aggregation maps are reported now, it does expose implementation details which is probably not ideal for default output.

If these stats are forced, not having them clutter up the "standard" interactive output seems sensible (although personally I will miss it as having some live output option makes it a bit easier to test different things). But not sure if it cluttering up their JSON output is great either, as if you have quite little traffic (e.g. deployment on personal equipment rather than ISP) then these map utilization stats may very well make up the majority of your entires (6 maps being reported every interval)

And again, unless we want to hardcode a reporting interval (or tie it something else), the users must provide an interval anyways.

pping/pping.c
simosund: Was thinking of making it optional to pass a map utilization context in case the user didn't request those stats. But then found that adding a check for if the user had actually requested the map utilization stats at the calling sites for these functions was a bit ugly, so just always passed it regardless (considering the overhead for incrementing a couple of counters should be relatively small). 

Will remove the NULL-option.

pping/pping.c
simosund: I agree that there are defintely some things that could use a refactor (especially the monstorsity `load_attach_bpfprogs()` comes to mind, even if it just has two parameters =)). But as you say, that probably belongs in a separate PR (which I will get around to....later..).

Otherwise, agree with the points. Passing on `n_cpus` was some premature optimization (so bad idea), as I haven't bothered to look up what `libbpf_num_possible_cpus()` actually does and figured if it called /proc or something that was perhaps not optimal to do thousands of times in a loop.

And sure, I touched the code, but it was all just renaming aggregated RTTs to aggregated stats (which I hope you're fine with keeping in this PR, considering that this PR is partially repsonsible for making the aggregated stats more than just RTTs).

pping/pping.h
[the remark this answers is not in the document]
simosund: > I'm not sure I quite follow your idea for having an array of packet counts for the esoteric protocols. I take it you don't mean an full 256-member array as now, because then that would obviously waste just as much space (although be nicer to the cache by simply not using it for the common protocols). So do you mean a much smaller array for some selection of esoteric protocols (which we must then remap from their IP-protocol number to a more compact index). Or just have an array with space for X ( X << 256) protocols and then create the protocol <-> index mapping dynamically?

Never mind, think I understand now. You meant having a 256-member array for rarer protocols, but only packet counts for them (not byte counts)?

Still not sure if I agree with turning them into u32s though. Even if there's not too much traffic, say 1000 pps, of some protocol other than basically TCP/UDP then those counters will still overflow if you run the tool for around a month. So I guess I would at least have to add some kind of wrap-around calculation then (based on the assumption that if a later report has lower numbers than a previous one, it has overflown exactly once). Is that added complexity really worth it to potentially improve cache-usage for what we expect to be relatively rarely accessed counters?

pping/pping_kern.c
simosund: As I tried to explain in that comment, that final round is needed to update the map utilization stats, specifically the runtime (the number of entries iterated through I guess I could just update with +1 for each "non-empty" iteration instead). 

But I guess it's perhaps better to measure that runtime from user space instead (assuming reading the link is blocking)? Or just skip it entirely as I'm not sure if either way of measuring it is particularly accurate? Although it would be nice to get a sense of how much processing we're spending on just cleaning the maps.

And both `flowmap_cleanup()` and `tsmap_cleanup()` early-abort right after this map utilization update if key/value are NULL, which is as early as possible with the current way of measuring the run time.

pping/pping_kern.c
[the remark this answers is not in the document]
tohojo: Hmm, yeah, I agree that it makes more sense to consider rx as "received from that subnet". Definitely make it consistent with everything else! :)

And no, I don't have an idea for a better name...

pping/pping.h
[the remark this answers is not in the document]
tohojo: > Never mind, think I understand now. You meant having a 256-member array for rarer protocols, but only packet counts for them (not byte counts)?

Yes, exactly. And just keep the full array, but don't use the entries for UDP and TCP.

> 1000 pps... overflow if you run the tool for around a month.

Do you really expect to see 1000 pps of non-UDP/TCP traffic anywhere? I would be surprised. I would suggest not handling this until it becomes an actual problem - you can always increase the size of the counters later :)

pping/pping.c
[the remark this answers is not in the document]
tohojo: Making things configurable has a cost: It makes the code more complex (more branches), and it makes the tool more difficult to use (user has to understand and decide on more options). Especially testing becomes a lot more complicated the more options there are, as you get combinatorial explosion of options that you all need to make sure works correctly in all combinations of enabled/disabled. Especially options that are off by default are really easy to accidentally break, unless you have a very robust test suite (which, well...) :)

So it's better to think of configuration knobs as a last resort: you're basically telling the user "I couldn't figure out if this feature was really useful, so I'm offloading that decision onto you". Don't do that unless you really have to.

My suggestion would be to tie these to the regular reporting interval, by some (big-ish?) multiple. That way you could even get rid of the separate timer and just go `if (report_count % counter_interval == 0) print_counters();`.

pping/pping_kern.c
tohojo: Yeah, just measuring the runtime in userspace would probably be simpler :)

pping/pping.c
tohojo: See above: try your best to implement a default always-on way of reporting that is useful and drop the config option. If it later turns out that there are some *specific* use cases where this doesn't work well, **then** you can add a config option. But don't start out with that :)

pping/pping.c
tohojo: > (which I will get around to....later..)

I'm sure you're aware that this "...later..." has a tendency to turn into "never" for these kinds of things. Which is part of the reason why I'm pointing it out now; and if it does indeed turn into "never" then the next step will have to be to block merging of any more features until things have been cleaned up :)

pping/pping.c
simosund: Then I will have to rethink these map utilization stats quite a bit, because in my mind this is basically debug information, and having debug info always on doesn't make much sense to me. I didn't give it a short option as I assumed it would rarely be used.

The only stat here I think would be of general use in normal output is the failure to create map entries, as that actually impacts the network stats they get from the tool (will basically miss some RTT samples). Otherwise I'm not sure why most users would care about any other map utilization stats, as it's just a form of self-monitoring (and in general I don't think users should have to be aware of that the tool uses a bunch of BPF maps, even less what the various maps actually are).

I guess I could create some form of heuristic for when the map utilization stats might be of interest, e.g. when there were failures to create map entries or a map is close to being full some stats for it might be relevant. Although I don't think that type of conditional output will be any less complex than having a user toggle (and definitely not any easier to test). 

Will get back with a suggestion for how these stats can be reported once I have some more concrete idea for how to make them both user-friendly enough to be part of normal output while also being detailed enough that they can be of use to me to figure out if there's any issues with the program.

tohojo: If some of them are strictly debug-only, then that could be a fair argument for hiding them behind a switch, but then I think it would make more sense with a more generic `--debug` switch? I'd be more sympathetic to a performance argument for this though, and AFAICT the graphs you posted above pertain to the global counters not the map cleanup stats, right?

Also, as a general point, I'm not saying **never** make something an option. What I'm pointing out here is more that there's often a (quite natural) tendency to just go with making something optional or configurable by default, and then using that as an excuse to think less hard about the feature itself "because users will only turn it on if they really need it". Whereas if the configuration knob is the last resort, we're forced to think through all the alternatives first, which will most likely improve things overall. In this case, for instance:

- Which stats are really useful to have around all the time, and which are only useful for specific debugging scenarios? Are they really useful for that debugging, or should we just not gather them at all?
- If the concern is emitting too much data, maybe a generic mechanism that suppresses counters that don't change is actually a better solution overall?
- Can we improve the performance of the stats gathering in general, given that they will be always on?
- Etc...

All this to say, yes, please do think about all this and make a concrete suggestion. And if that suggestion still includes some kind of configuration knob, if it's well-reasoned maybe that is OK ;)

simosund: > AFAICT the graphs you posted above pertain to the global counters not the map cleanup stats, right?

The first set of plots, with just "pre-update" and "post-update", the post-update show with everything turned on (i.e. aggregation + global counters + map stats). In the second set of plots the "post a1 g1 m1" is with everything turned on while "post a1 g1" is aggregation + global counters, but without the map stats. 

So the difference between the green and the red line in the second set of plots seem to show the impact of the map stats, but don't have any version with just aggregation + map stats (without global counters), so it could potentially be related to some combinatorial thing (i.e. just map stats is fine, but when combined with global counters as well cache efficiency starts to drop. As I don't see why map stats would have any real impact for pure UDP traffic where none of the map stats are ever updated, although I guess it could also be from the additional user space processing which has to contend with all the packet processing as the tests are done on a single core.

pping/pping.c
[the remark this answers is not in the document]
simosund: > My suggestion would be to tie these to the regular reporting interval, by some (big-ish?) multiple. That way you could even get rid of the separate timer and just go if (report_count % counter_interval == 0) print_counters();

Getting rid of the additional timer would be nice. However, unless you object I will probably go ahead and use the multiple 1, i.e. get it once for every aggregation report. The amount of output from the global counters is quite small compared to the aggregated output (a few protocol counters vs packet/byte counters + rtt stats + rtt histogram per subnet). So from a user perspective these global counters just becomes a part of the aggregated output then (i.e. you get global protocol counts + various per-subnet stats for each interval). 

It does mean the global counters will no longer work outside of the aggregated mode, as there is no interval to tie it to for the individual RTTs (and not sure that adding them for every X individual RTTs is necessarily a good idea). But think that's probably alright, as the global counts is also a form of aggregation, so it can just be seen as a part of the aggregation feature.

pping/pping.c
abrunstrom: I am no expert here, but a generic `--debug` switch for all internal information (to the extent that information is needed) that separates it from more generally useful always-on information makes a lot of sense to me.

simosund: So have tried thinking about some different alternatives for these map utilization stats now. In the end I don't think I've landed too far away from what I originally proposed, i.e. most of this should be default-off, but will try to lay out my reasoning a bit more.

In the end, the only "map utilization" stat (of those I currently collect) that I think would be generally useful for most users, and thus should be always on, is for failing to create an entry. Any time we fail to create an entry, that signals an actual issue which may result in missing some data (failing to create a timestamp entry will miss the RTT for that specific timestamp, failing to create a flow-state entry will miss all RTTs for that flow until its state can be created, and failing to create an aggregation entry will result in stats being added to the fallback /0 subnet instead of /x subnet it was really from). 

All the other map stats are really mainly useful for me to figure out if we have sized the maps sensibly and see if they're actually used how I think they're used. E.g. for timestamp map I expect a fair amount of churn, with lots of entries being quickly added an deleted, and hopefully at most 50% of the entries being expired, while for the aggregation maps I expect quite little churn and both instances to be fairly similar to each other. Always reporting these seems like it would just be confusing to most users, who don't have any knowledge of which maps are used for what, and what the normal/expected values would be. To me it's a bit like if a normal user space application would always include stats about memory allocations for its own various internal buffers in its output. 

So reporting for the "creation failed" should probably be separate from any reporting of other map utilization stats. As failing to create a map entry is a form of error, it should probably be highlighted/separated from normal reporting output in some way. I'm not entirely decided on what the best way to achieve this separation is. One alternative is to keep them as warnings that are written to stderr, similar to the current "map full" warnings. The drawback with such warnings is that they don't integrate nicely with the normal output, e.g. if writing to file the warnings will still just go to stderr and not be part of that file, and will probably just be human-readable text rather than ex. writing a JSON-entry to stderr. Another option is to still keep them in the normal output stream (stdout or file), but still try to indicate that they are a type of warning by ex. including the word "warning" in the output. Ideas are very much welcome here. These "warnings" should only be reported when they happen as they should be an exception and not the norm (and if they are the norm then ePPing is probably not working too well in your environment). Periodic reporting with "there were no failures" is fairly useless, instead only report when there are failures (and assume there were no failures otherwise).

For all the other map stats, I think they should be opt-in. I'm fine with calling it `--debug` instead of `--map-stats`, but unless we in the future add some other type of debug information it will in practice just be map utilization stats. Number of entries in the map I think makes sense to report periodically (even if they're unchanged since last report), but other counters (such as entries created or expired) can be omitted if they're unchanged to reduce the amount of output a bit. Sending the reporting interval as a value to `--debug` seems a bit strange, and won't necessarily make any sense for future additions of debug information (e.g. might have some other debug information which are more event-based), so we can probably just hardcode this to some relatively low value (leaning towards every second, but every 10s would probably also work).

That the "failure to create" entries are reported separately from other map utilization stats don't necessarily mean that the collection of those stats have to be implemented as two independent systems though. Both could could be achieved with a system similar to what I've already implemented for this PR (per-CPU map with map utilization struct that is updated by eBPF programs and then periodically fetched from user space where diff since previous state is calculated), where we just have a switch in user space that determines if any stats other than failure ones are actually reported.

I do think all the set of map utilization stats I added serve a purpose:
- **Number of entries in map:** Useful to see how much of the available map space we're actually using, as well as how stable or bursty it is. This is especially interesting to look at when there are map creation failures. Is the map sitting just below the map size for long periods or time, or is it just a short burst which cause map creation failures?
- **Entries created, deleted and expired:** First off, I need these to actually calculate the number of entries in the map when using per-CPU entries (as an entry may be deleted by a different core than created it, so if just incrementing/decrementing a single "entries" counter instead might result in negative values on some cores). These are also more informative than just the number of entries, as they also tell me the "churn" rate in the maps. Especially the timestamp map has many very short-lived entries (they only live for the duration of an RTT), where a single flow in theory can create and delete 1000 timestamps per second, while only taking up a single entry slot in the map at any given time. The separation between "deleted" (by eBPF program matching timestamp or detecting FIN/RST and closing flow) and "expired" (period cleanup decided entry was stale and deleted it) also gives a measure of how well the program is working. Ideally, we want most entries to be deleted rather than expired. If a large fraction of timestamp entries are expired then we're doing a lot of unnecessary work that doesn't give us an RTT in the end, and if a lot of flow entries are expired then our flow closing detection is not working very well.
- **Cleanup cycles, cleanup iterations and cleanup time:** Useful to get a sense of how much overhead the periodical cleanup contributes to. Having the number of entries iterated through in addition to cleanup time is useful to assess how the cleanup time scales with the number of entries (would think fairly linearly, but caching and other factors could perhaps make it less linear), which in turn also helps to assess how increasing the map size may impact the overhead. The number of entries that map cleanup iterates through I would generally expect to be close to the current number of entries in the map, but if the map size varies rapidly then the current size of the map may be quite different from the size of the map when the periodical cleanup looped through it, or there may be some bug/unforeseen issue which causes the cleanup to often iterate through the entries multiple times (old userspace iteration would restart from beginning if an entry was deleted) or skip some entries (e.g. skips to next hash-bucket rather than next entry in bucket if bucket is modified in some way). And as this map utilization report is not tied to the map cleanup interval, it's good to know how many clean-cycles the iteration-count and runtime is for. 

So this is my suggestion for how to collect and report map utilization stats. Some other alternatives include:
- Always report map utilization stats. To make this reasonable I think we have to massively simplify the stats that are reported, as I would not expect the average user to understand my current set of metrics, nor the set of maps. For example, we might just report the the "utilization" of map slots, i.e. the number of entries in the map compared to the max map size (0-100%). It's arguably not a good idea to report stats for the individual maps, as which maps we use is an implementation detail, and this especially goes for two sets of maps (active/inactive) used for the aggregation maps, so maybe just report some average across multiple maps. But simplifying the stats like this make them much less useful to me, and still not that relevant for others.
- Collect map utilization stats in the eBPF programs, but don't build any user space component to report them. Instead simply rely on bpftool to dump the maps whenever you want the stats. This makes the user space of ePPing less complex, and also allows for some flexibility (you don't need to report at any fixed interval etc.). The drawback is that this results in more raw data as it will give you the per-CPU counts (rather then summed ones ePPing provides) and totals rather then diffs (as JSON is a text-format, larger numbers take up more space than smaller ones). It also just shifts the work for dealing with the map entries done by ePPing into whatever script you create the parse this data.
- Drop this feature. As I have argued, most of the map utilization stats will not be useful for most users, it's something that's primarily useful for me, myself, and I. As I'm currently the one doing most of the development anyways, I can just add whatever debug stats I need ad-hoc while adding a feature (and then remove them from the final PR). In this case it's just that I would like to get these stats from a more realistic deployment (like Robert's) rather than just my development environment or the testbed. We could get similar stats by just compiling with DEBUG, just that the output format for that is pretty terrible (text without timestamps to stderr) and less complete. 

If anyone manages to get through this essay and have any input I'll happily discuss different options =) Otherwise, I'll just go ahead and implement something similar-ish to what I've described and I guess we'll just end up having the discussion as part of the next review round.

tohojo: > We could get similar stats by just compiling with DEBUG, just that the output format for that is pretty terrible (text without timestamps to stderr) and less complete.

If it's really only something that's useful for your own one-off debugging, how about still keeping it behind a compile-time flag, but making it part of the regular JSON stats output (like in this PR) when it is enabled?

simosund: I have a hard time seeing who other than me would have use for the map utilization stats as of now, so I'm fine with leaving this behind a compile-time flag. If it turns out that it could actually be useful for others in the future I guess we can change that then.

As I will need some additional code in user space (compared to pre-PR debug stats) to gather stats for the aggregation maps + runtime for the cleanup I will have to think a bit about how to approach it to avoid cluttering the code with too many `#ifdef`s, but can probably use a similar strategy as the pre-PR debug code in the BPF progs, simply have separate debug functions which become empty if the DEBUG flag is not set.

As for notifying users about failure the failure to create map entries, any opinion on how to handle that? Keep it as warnings written to stderr, or include it as part of the normal output stream? I'm leaning towards stderr as they should generally be quite rare (for the ~1 month at Robert we got a total of about 100 such warnings), so they shouldn't clutter up the system journal too much. 

Thinking it's perhaps better to drop this rework of the map utilization stats + map full warnings from this PR and have a separate PR for that.

tohojo: > I have a hard time seeing who other than me would have use for the map utilization stats as of now, so I'm fine with leaving this behind a compile-time flag. If it turns out that it could actually be useful for others in the future I guess we can change that then.

SGTM

> As I will need some additional code in user space (compared to pre-PR debug stats) to gather stats for the aggregation maps + runtime for the cleanup I will have to think a bit about how to approach it to avoid cluttering the code with too many `#ifdef`s, but can probably use a similar strategy as the pre-PR debug code in the BPF progs, simply have separate debug functions which become empty if the DEBUG flag is not set.

Yeah, sounds reasonable. On the BPF side you can also use const-valued variables and rely on the verifier's dead code elimination if that makes things easier...

> As for notifying users about failure the failure to create map entries, any opinion on how to handle that? Keep it as warnings written to stderr, or include it as part of the normal output stream? I'm leaning towards stderr as they should generally be quite rare (for the ~1 month at Robert we got a total of about 100 such warnings), so they shouldn't clutter up the system journal too much.

How about *also* keeping a global counter in the global stats struct, so you get both the warning, but if pping is run in a way that no one reads the warnings, you'll still see the counter value in the stats so you can keep track?

> Thinking it's perhaps better to drop this rework of the map utilization stats + map full warnings from this PR and have a separate PR for that.

Yup, good idea!

pping/pping.h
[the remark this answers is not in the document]
simosund: > Do you really expect to see 1000 pps of non-UDP/TCP traffic anywhere?

I don't really expect it to be common, but I'm not confident enough in my knowledge of networks to feel reasonably certain that I can rule it out. Thinking that for example some type of tunneling protocol (if used) could end up generating a fair amount of traffic. I also recently came across [another eBPF powered latency measurement tool](https://dl.acm.org/doi/10.1145/3609021.3609302) (basically in-line traceroute) which would likely generate a fair amount of ICMP traffic. 

Oh, and talking of ICMP, I did add separate counters packet/byte counter for ICMP (and ICMPv6) as well as I expect that to be somewhat common, and is a protocol we support passively measuring latency for, so felt weird to exclude that from the protocol counters. 

> I would suggest not handling this until it becomes an actual problem - you can always increase the size of the counters later :)
Despite my lack of conviction, I've stuck to u32s for now. I don't think it will pose a problem for any <= month long measurement at Robert, so I don't think it will mess up any measurement data we collect in the near term.

pping/pping_kern.c
```diff
@@ -314,6 +321,28 @@ get_dualflow_key_from_packet(struct packet_info *p_info)
...
+	__u32 key = 0;
+	struct global_packet_counters *glob_count;
+
+	glob_count = bpf_map_lookup_elem(&map_global_counters, &key);
+	if (!glob_count) // Should never happen
+		return;
+
+	if (ipproto < 0) {
+		glob_count->non_ip.packet_count++;
+		glob_count->non_ip.byte_count += pkt_len;
```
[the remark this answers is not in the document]
simosund: Just had to check that the `parse_ip6hdr()` function wouldn't potentially return an extension header if it ran out of loops (as we've decreased the loop count in ePPing a bit to combat the complexity). But it does indeed fail if it can't parse the actual IPv6 header, so it *should* be safe to reuse 0 for non-IP instead, so I've applied that.

pping/pping.c
```diff
@@ -1119,59 +1120,291 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
...
+	for (proto = 0; proto < N_IPPROTOS; proto++) {
+		diff->other_ipprotos[proto] = next->other_ipprotos[proto] -
+					      prev->other_ipprotos[proto];
 	}
 }
 
-static void clear_aggregated_rtts(struct aggregated_rtt_stats *stats)
+static int report_globalcounters(struct output_context *out_ctx, int map_fd)
+{
+	static struct global_counters prev_cnt = { 0 };
```
tohojo, line 1295: Making this a static variable is not a good idea: it adds hidden state to this one function, which makes it harder to reason about the function behaviour (it's something you have to remember that the function does). Better to explicitly pass in the "prev" struct. Or maybe store it in out_ctx, since you're passing that in anyway?

pping/pping.c
```diff
@@ -1119,59 +1120,291 @@ static void handle_missed_events(void *ctx, int cpu, __u64 lost_cnt)
...
+	print_pktbytes_tuple_json(jctx, "ICMP", counters->icmp_pkts,
+				  counters->icmp_bytes);
+	print_pktbytes_tuple_json(jctx, "ICMPv6", counters->icmp6_pkts,
+				  counters->icmp6_bytes);
+
+	for (proto = 0; proto < N_IPPROTOS; proto++) {
+		if (counters->other_ipprotos[proto] > 0)
+			print_pktbytes_tuple_json(
+				jctx, proto_to_str(proto),
+				counters->other_ipprotos[proto], 0);
```
tohojo, line 1217: Hmm, so this will give you output like `"12": {"packets": 1}` for all the "misc" protocols? I'm wondering if that is a bit verbose, but maybe it's better to be consistent.

That `proto_to_str()` function returning a pointer to a random global variable is incredibly thread-unsafe, though (and even unsafe against being called multiple times before the previous value is used). The way you're using it here, that's fine, but it's better to change it to make the caller supply the buffer (surprising side effects in utility functions like this is a recipe for future pain).

tohojo (changes requested):
Just a few nits, related to your fondness for using global variables in weird places (`static` variables inside a function is just a differently named global variable). But otherwise looks good, and scoping the PR down makes it easier to review as well, which is nice :)

pping/pping.c
simosund: > Better to explicitly pass in the "prev" struct. Or maybe store it in out_ctx, since you're passing that in anyway?

While avoiding extra parameters would be nice, I'm not sure it conceptually makes sense to stuff the the global counters stats in the output context? Isn't really connected to how the output should be handled? And as the output context is recreated when a file is reopened (on HUP-signal), I'll also have to modify the `open_output()` and `reopen_output()` functions to copy over that counter state, which doesn't feel like something those functions should be touching IMO. 

I'm thinking I'll create a new "aggregation context" struct which will just consist of the current aggregation maps struct + the global counters state, and pass this aggregation context to the `report_aggregated_stats()` function (which can then pass in the aggregation context to `report_globalcounters() to get both map-fd + prev in one argument).

pping/pping.c
simosund: > Hmm, so this will give you output like "12": {"packets": 1} for all the "misc" protocols? I'm wondering if that is a bit verbose, but maybe it's better to be consistent.

Yes, it's a bit verbose, but don't think that's too problematic as it's only reported once per aggregation interval, and we don't expect too many misc protocols. If you have some good idea for a more compact format that is preferably roughly as self-descriptive and somewhat consistent I'll happily consider it, but think it will make little different on the overall output footprint considering we're at every such interval also reporting potentially thousands of per-subnet entries (which in this PR has also gotten much more verbose due to having separate TCP-with-TS/TCP-without-TS/other packet/byte counters).

> That proto_to_str() function returning a pointer to a random global variable is incredibly thread-unsafe, though (and even unsafe against being called multiple times before the previous value is used). The way you're using it here, that's fine, but it's better to change it to make the caller supply the buffer (surprising side effects in utility functions like this is a recipe for future pain).

So, just to be clear, you think I should change the `proto_to_str()` function in this PR? I guess it will then have to copy over the hardcoded strings ("TCP" etc) to that passed buffer then which is slightly less efficient for the common use cases (especially when it's used for individual RTT-reporting), but guess that overhead should be rather minuscule compared to all the work with formatting the output strings anyways.

simosund: While I have it on my mind, I guess you're not particularly fond of my (somewhat awkwardly named) `get_libbpf_strerror()` wrapper either then. So guessing replacing that with adding buffer's + calls `libbpf_strerror()` should be added to that future refactoring PR.

tohojo: > So, just to be clear, you think I should change the proto_to_str() function in this PR? 

Yeah, just add it as a separate commit at the start of the series.

> I guess it will then have to copy over the hardcoded strings ("TCP" etc) to that passed buffer then which is slightly less efficient for the common use cases 

The impact of a five-byte strcpy() will be negligible :)

> While I have it on my mind, I guess you're not particularly fond of my (somewhat awkwardly named) get_libbpf_strerror() wrapper either then. 

Good guess ;)

simosund:
Hi @tohojo,
I've addressed the two things you pointed out in your last review. So please have another look when you have time.

tohojo (changes requested):
I think the code looks good now, so only have a bit of bikeshedding on the error counters:

Basically, I don't like 'map_failures' name - that's exposing low-level details. Let's name them by the high-level errors they are a symptom of, and just have the containing JSON object be called 'errors'. So maybe something like:
```
  "errors": {
    "store_packet_ts": 499,
    "create_flow_state": 1,
    "create_agg_subnet_state": 122923
  }
```

(feel free to come up with shorter substitutes along those lines)

simosund:
> Basically, I don't like 'map_failures' name - that's exposing low-level details. Let's name them by the high-level errors they are a symptom of, and just have the containing JSON object be called 'errors'.

Fair point. I guess you were mainly refering to the output, but went ahead and renamed it internally as well from `map_failures` to a more generic `pping_error_counters`. While the 3 error counters for now are all related to failure to create map entries, it could also be extended with other types of errors down the line.

So please have another look at the last commit.

pping/pping.c
```diff
@@ -1271,6 +1298,17 @@ static void print_ecncounters_json(json_writer_t *jctx,
...
 }
 
+static void print_ppingerrors_json(json_writer_t *jctx,
+				   const struct pping_error_counters *errors)
+{
+	jsonw_start_object(jctx);
+	print_counter_json(jctx, "store_packet_ts", errors->pktts_store);
+	print_counter_json(jctx, "create-flow-state", errors->flow_create);
+	print_counter_json(jctx, "create-agg-subnet-state",
+			   errors->agg_subnet_create);
```
tohojo, line 1308: You should probably make up your mind about whether you're going with dashes or underscores here ;)

simosund: Whoops, good catch! Should be all underscores (as that's what all other JSON entries use). Will fix.

tohojo (approved):

PR #103, opened 2023-10-04
Add xdp-synproxy to bpf-examples

vincentmli:
this code is from kernel bpf selftests xdp synproxy, removed the tc part for simplicity, shows an exmaple of using libxdp to attach xdp synproxy program on network interface.

if port is not in allowed ports, the packet will be dropped by xdp synproxy by default, this would break tcp connections to ports that user does not want to do synproxy, change the default to allow connection pass through.

### DIFF
--- a/xdp-synproxy/xdp_synproxy.c
+++ b/xdp-synproxy/xdp_synproxy.c
@@ -1,0 +1,388 @@
+// SPDX-License-Identifier: LGPL-2.1 OR BSD-2-Clause
+/* Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved. */
+
+#include <stdnoreturn.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <unistd.h>
+#include <getopt.h>
+#include <signal.h>
+#include <sys/types.h>
+#include <bpf/bpf.h>
+#include <bpf/libbpf.h>
+#include <xdp/libxdp.h>
+#include <net/if.h>
+#include <linux/if_link.h>
+#include <linux/limits.h>
+#include <sys/resource.h>
+
+#include "logging.h"
+#include "xdp_synproxy_kern.skel.h"
+
+
+#define STRERR_BUFSIZE 1024
+
+static unsigned int ifindex;
+static __u32 attached_prog_id;
+
+static noreturn void cleanup()
+{
+        char errmsg[STRERR_BUFSIZE];
+        struct xdp_program *p = NULL;
+        int err;
+
+	if (attached_prog_id == 0)
+		exit(0);
+
+	p = xdp_program__from_id(attached_prog_id);
+
+	err = xdp_program__detach(p, ifindex, 0, 0);
+	xdp_program__close(p);
+        if (err) {
+		libxdp_strerror(err, errmsg, sizeof(errmsg));
+		pr_warn("Couldn't detach BPF program: %s(%d)\n", errmsg, err);
+		p = NULL;
+		exit(1);
+        }
+
+	exit(0);
+}
+
+static noreturn void help(const char *progname)
+{
+	fprintf(stderr, "Usage: %s [--iface <iface>|--prog <prog_id>] [--mss4 <mss ipv4> --mss6 <mss ipv6> --wscale <wscale> --ttl <ttl>] [--ports <port1>,<port2>,...]\n",
+		progname);
+	exit(1);
+}
+
+static unsigned long parse_arg_ul(const char *progname, const char *arg, unsigned long limit)
+{
+	unsigned long res;
+	char *endptr;
+
+	errno = 0;
+	res = strtoul(arg, &endptr, 10);
+	if (errno != 0 || *endptr != '\0' || arg[0] == '\0' || res > limit)
+		help(progname);
+
+	return res;
+}
+
+static void parse_options(int argc, char *argv[], unsigned int *ifindex, __u32 *prog_id,
+			  __u64 *tcpipopts, char **ports)
+{
+	static struct option long_options[] = {
+		{ "help", no_argument, NULL, 'h' },
+		{ "iface", required_argument, NULL, 'i' },
+		{ "prog", required_argument, NULL, 'x' },
+		{ "mss4", required_argument, NULL, 4 },
+		{ "mss6", required_argument, NULL, 6 },
+		{ "wscale", required_argument, NULL, 'w' },
+		{ "ttl", required_argument, NULL, 't' },
+		{ "ports", required_argument, NULL, 'p' },
+		{ NULL, 0, NULL, 0 },
+	};
+	unsigned long mss4, wscale, ttl;
+	unsigned long long mss6;
+	unsigned int tcpipopts_mask = 0;
+
+	if (argc < 2)
+		help(argv[0]);
+
+	*ifindex = 0;
+	*prog_id = 0;
+	*tcpipopts = 0;
+	*ports = NULL;
+
+	while (true) {
+		int opt;
+
+		opt = getopt_long(argc, argv, "", long_options, NULL);
+		if (opt == -1)
+			break;
+
+		switch (opt) {
+		case 'h':
+			help(argv[0]);
+			break;
+		case 'i':
+			*ifindex = if_nametoindex(optarg);
+			if (*ifindex == 0)
+				help(argv[0]);
+			break;
+		case 'x':
+			*prog_id = parse_arg_ul(argv[0], optarg, UINT32_MAX);
+			if (*prog_id == 0)
+				help(argv[0]);
+			break;
+		case 4:
+			mss4 = parse_arg_ul(argv[0], optarg, UINT16_MAX);
+			tcpipopts_mask |= 1 << 0;
+			break;
+		case 6:
+			mss6 = parse_arg_ul(argv[0], optarg, UINT16_MAX);
+			tcpipopts_mask |= 1 << 1;
+			break;
+		case 'w':
+			wscale = parse_arg_ul(argv[0], optarg, 14);
+			tcpipopts_mask |= 1 << 2;
+			break;
+		case 't':
+			ttl = parse_arg_ul(argv[0], optarg, UINT8_MAX);
+			tcpipopts_mask |= 1 << 3;
+			break;
+		case 'p':
+			*ports = optarg;
+			break;
+		default:
+			help(argv[0]);
+		}
+	}
+	if (optind < argc)
+		help(argv[0]);
+
+	if (tcpipopts_mask == 0xf) {
+		if (mss4 == 0 || mss6 == 0 || wscale == 0 || ttl == 0)
+			help(argv[0]);
+		*tcpipopts = (mss6 << 32) | (ttl << 24) | (wscale << 16) | mss4;
+	} else if (tcpipopts_mask != 0) {
+		help(argv[0]);
+	}
+
+	if (*ifindex != 0 && *prog_id != 0)
+		help(argv[0]);
+	if (*ifindex == 0 && *prog_id == 0)
+		help(argv[0]);
+}
+
+static int syncookie_attach(unsigned int ifindex)
+{
+	char errmsg[STRERR_BUFSIZE];
+	struct xdp_program *p = NULL;
+	struct xdp_synproxy_kern *skel;
+	int err;
+
+	char *xdp_program_name = "syncookie_xdp";
+	DECLARE_LIBXDP_OPTS(xdp_program_opts, xdp_opts, 0);
+
+	skel = xdp_synproxy_kern__open();
+        if (!skel) {
+		err = -errno;
+		pr_warn("Couldn't open XDP program: %s\n", strerror(-err));
+		return err;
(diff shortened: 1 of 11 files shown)


vincentmli:
Hi @tohojo:

Here is the PR to add xdp-synproxy to bpf-examples repo, please review :)

vincentmli:
Hi @tohojo do you have time to review this PR? :)

xdp-synproxy/xdp_synproxy.c
```diff
@@ -0,0 +1,391 @@
...
+	static struct option long_options[] = {
+		{ "help", no_argument, NULL, 'h' },
+		{ "iface", required_argument, NULL, 'i' },
+		{ "prog", required_argument, NULL, 'x' },
+		{ "mss4", required_argument, NULL, 4 },
+		{ "mss6", required_argument, NULL, 6 },
+		{ "wscale", required_argument, NULL, 'w' },
+		{ "ttl", required_argument, NULL, 't' },
+		{ "ports", required_argument, NULL, 'p' },
+		{ "file", required_argument, NULL, 'f' },
```
tohojo, line 84: Is supplying the filename on the command line really a good idea? Seems a bit cumbersome, and since there's no checks on the supplied objects, things can quickly get out of sync...

xdp-synproxy/xdp_synproxy_kern.c
```diff
@@ -0,0 +1,807 @@
...
+	if (xdp) {
+		if (bpf_xdp_adjust_tail(ctx, TCP_MAXLEN - hdr->tcp_len))
+			return XDP_ABORTED;
+	}
+
+	return XDP_TX;
+}
+
+static __always_inline int syncookie_part2(void *ctx, void *data, void *data_end,
+					   struct header_pointers *hdr, bool xdp)
```
tohojo, line 756: The `xdp` arg seems to be leftover from the in-kernel version that also supports TC? Let's just get rid of that...

xdp-synproxy/xdp_synproxy.c
```diff
@@ -0,0 +1,391 @@
...
+        }
+	xdp_program__close(p);
+
+	exit(0);
+}
+
+static noreturn void help(const char *progname)
+{
+	fprintf(stderr, "Usage: %s [--iface <iface>|--prog <prog_id>] [--mss4 <mss ipv4> --mss6 <mss ipv6> --wscale <wscale> --ttl <ttl>] [--ports <port1>,<port2>,...] [--file <path-to-xdp_synproxy_kern.o>]\n",
+		progname);
```
tohojo, line 55: So the command line interface seems a little strange to me. It attaches on the first run, but you can also supply a prog ID to replace the ports and options while it's running? But the only way to detach is to kill the original instance? IDK, maybe this is fine for an initial version, since we're just porting this over from the selftests...

tohojo (changes requested):

xdp-synproxy/xdp_synproxy.c
vincentmli: the in kernel version search the object file, my reasoning is this allow user to specify whatever location user prefer to put the object file in to give user the flexibility, also when run in container or in k8s as daemonset, this allow user to customize the object file location to fit their container environment. I will add check to see if the file exist in location that is specified by the argument, how does that sound?

xdp-synproxy/xdp_synproxy_kern.c
vincentmli: ok, got it

xdp-synproxy/xdp_synproxy.c
vincentmli: as an initial version and to simplify, I will just remove the prog ID argument. I think the original version intention is to have some kind of map persistency for the generated syncookie counter, so each time when the program is re-loaded/attached, the counter is not reset to 0, and probably as you pointed out to reconfigure ports and options maybe.

tohojo: Sure, let's just keep it this way for now...

xdp-synproxy/xdp_synproxy.c
vincentmli: ah, I think I got it, let me know otherwise :)

```
diff --git a/xdp-synproxy/xdp_synproxy.c b/xdp-synproxy/xdp_synproxy.c
index ed073c1..18390f3 100644
--- a/xdp-synproxy/xdp_synproxy.c
+++ b/xdp-synproxy/xdp_synproxy.c
@@ -167,9 +167,11 @@ static int syncookie_attach(char *file, unsigned int ifindex)
        int err;
 
        char *xdp_filename = file;
+       char *xdp_program_name = "syncookie_xdp";
        DECLARE_LIBXDP_OPTS(xdp_program_opts, xdp_opts, 0);
 
        xdp_opts.open_filename = xdp_filename;
+       xdp_opts.prog_name = xdp_program_name;
        p = xdp_program__create(&xdp_opts);
        err = libxdp_get_error(p);
        if (err) {
```

tohojo: Yup, exactly :)

vincentmli: thanks! pushed the new change.

vincentmli:
@tohojo I like your idea of using skeleton to save the user space program from loading external object file, so I updated the code with initial skeleton changes, the code could still be optimized like the map access through skeleton, maybe we save that for next revision and get this example in first, what do you think?

xdp-synproxy/xdp_synproxy.c
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	xdp_opts.obj = skel->obj;
+	xdp_opts.prog_name = xdp_program_name;
+	p = xdp_program__create(&xdp_opts);
+	err = libxdp_get_error(p);
+	if (err) {
+		libxdp_strerror(err, errmsg, sizeof(errmsg));
+		pr_warn("Couldn't create xdp program: %s(%d)\n", errmsg, err);
+		return err;
```
tohojo, line 184: This needs to close the skeleton; probably a second error label at the bottom is easiest?

xdp-synproxy/xdp_synproxy.c
```diff
@@ -0,0 +1,388 @@
...
+	if (attached_prog_id == 0)
+		exit(0);
+
+	p = xdp_program__from_id(attached_prog_id);
+
+	err = xdp_program__detach(p, ifindex, 0, 0);
+        if (err) {
+		libxdp_strerror(err, errmsg, sizeof(errmsg));
+                pr_warn("Couldn't attach BPF program: %s(%d)\n", errmsg, err);
+                p = NULL;
```
tohojo, line 45: This should be an xdp_program__close().

tohojo: (Or just call close before checking `err`)

tohojo (changes requested):
A few error path nits

xdp-synproxy/xdp_synproxy.c
vincentmli: pushed the change, hope I did not miss anything :)

tohojo: You didn't fix this one...

xdp-synproxy/xdp_synproxy.c
```diff
@@ -0,0 +1,390 @@
...
+	if (err) {
+		p = NULL;
+		libxdp_strerror(err, errmsg, sizeof(errmsg));
+		pr_warn("Couldn't create xdp program: %s(%d)\n", errmsg, err);
+		goto out;
+	}
+
+	err = xdp_program__attach(p, ifindex, 0, 0);
+	if (err) {
+		xdp_program__close(p);
```
tohojo, line 190: Instead you moved this one so there's now no xdp_program__close on the success path...

tohojo (changes requested):
Not sure if it's the Github review interface making things up, but no, that was not the fix I had in mind :)

xdp-synproxy/xdp_synproxy.c
vincentmli: sorry for my misunderstanding, do we need xdp_program__close on success path (err == 0) ?  if the program is attached successfully, xdp_program__close will not cause the program to be detached?

vincentmli: sorry I am not clear on the skeleton destroy and program fd close imapct, so let me ask this to clarify  or if you could explain a little bit on the reason :)

1, if program is attached successfully
a, skeleton need to be destroyed ?
b, program fd needed to be closed?
if both yes, why?

2, if program is not attached successfully,
a, program fd needed to be closed first? yes
b, skeleton need to be destroyed ? yes

am I correct?

xdp-synproxy/xdp_synproxy.c
vincentmli: you were talking about code in cleanup here, and I am fixing code at syncookie_attatch, my bad :)

vincentmli: so the cleanup triggered after successful attach, but when SIGINT/SIGTERM is received, program is detached, if the detach failed, we xdp_program__close before checking the error message, correct? do we need to destroy the skeletion in cleanup?

vincentmli: I likely may still miss something, let me try again with this, so regardless if the program is detached successfully or not, the xdp program will be closed

```
diff --git a/xdp-synproxy/xdp_synproxy.c b/xdp-synproxy/xdp_synproxy.c
index 821608b..8183951 100644
--- a/xdp-synproxy/xdp_synproxy.c
+++ b/xdp-synproxy/xdp_synproxy.c
@@ -40,13 +40,14 @@ static noreturn void cleanup()
 
        err = xdp_program__detach(p, ifindex, 0, 0);
         if (err) {
+               xdp_program__close(p);
                libxdp_strerror(err, errmsg, sizeof(errmsg));
-                pr_warn("Couldn't attach BPF program: %s(%d)\n", errmsg, err);
+                pr_warn("Couldn't detach BPF program: %s(%d)\n", errmsg, err);
                 p = NULL;
-                exit(1);
+               exit(1);
         }
-       xdp_program__close(p);
 
+       xdp_program__close(p);
        exit(0);
 }
```

xdp-synproxy/xdp_synproxy.c
vincentmli: you commented for cleanup, I missed it

tohojo: But the xdp_program__close() was already below the out label, so the goto below took care of that.

xdp-synproxy/xdp_synproxy.c
```diff
@@ -0,0 +1,390 @@
...
+		goto out;
+	}
+	attached_prog_id = xdp_program__id(p);
+
+	err = 0;
+
+	signal(SIGINT, cleanup);
+	signal(SIGTERM, cleanup);
+
+	return err;
```
tohojo, line 202: Drop this return, we want the success path to also run the cleanup functions below.

xdp-synproxy/xdp_synproxy.c
tohojo: Yes, Every xdp_prog pointer needs to be passed to xdp_program__close() before exiting, regardless of whether it's going through an error path or a success path. Same thing with the skeleton, when created, it needs to be destroyed properly again.

There is no side effect to the xdp_program__close(), the program is only detached by an explicit xdp_program__detach().

xdp-synproxy/xdp_synproxy.c
tohojo: So what I meant was that you can combine the out path, so you change the end of the function, to be like:

```c
out_close:
    xdp_program__close(p);
out:
    xdp_synproxy_kern__destroy();
    return err;
```

and the you change this `goto out` below to be `goto out_close`. That way you hit both cleanup functions if the attach fails, and also in the success path below (if you get rid of the first return, see comment below).

tohojo (changes requested):

vincentmli:
could not comment on your comment from github, you cleared confusion 

"
Yes, Every xdp_prog pointer needs to be passed to xdp_program__close() before exiting, regardless of whether it's going through an error path or a success path. Same thing with the skeleton, when created, it needs to be destroyed properly again.

There is no side effect to the xdp_program__close(), the program is only detached by an explicit xdp_program__detach()."

should skeleton also be destroyed after program attachment, regardless if the attachment succeeded or not?  I pushed the change, bare with me, we are getting closer

xdp-synproxy/xdp_synproxy.c
vincentmli: my new change combined close and destroy into one same out label, I removed the return.

vincentmli:
so xdp_program and skeleton are sort of preparation resources for program attachment, after program attachment, regardless  attachment success or not, the xdp_program and skeleton are no longer needed and can be freed/destroyed, am I understanding it correct ?

tohojo:
Yup, Those data structures are for keeping the userspace data around, and once the program has been loaded into the kernel we no longer need them (since we're not using the skeleton for accessing the maps).

tohojo (approved):
Looks like we got all the combinations of error and success paths covered now :)

PR #112, opened 2024-01-05
track tcp payload offset as scalar in xdp_synproxy

vincentmli:
sync kernel commit 977bc146d (selftests/bpf: track tcp payload offset as scalar in xdp_synproxy).

without above fix, when run xdp-synproxy in kernel 6.7, it failed with verifier: "BPF program is too large. Processed 1000001 insn"

### DIFF
--- a/xdp-synproxy/xdp_synproxy_kern.c
+++ b/xdp-synproxy/xdp_synproxy_kern.c
@@ -61,6 +61,8 @@
 #define DEFAULT_TTL 64
 #define MAX_ALLOWED_PORTS 8
 
+#define MAX_PACKET_OFF 0xffff
+
 #define swap(a, b) \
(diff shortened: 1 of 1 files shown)


tohojo:
Please include the full commit message, including authorship and s-o-b line from the kernel commit, so it's clear that you didn't write this yourself. But do include information about any changes you made when porting the commit to this repository.

I'd suggest using the formatting we use for RHEL backports; which basically amounts to copy-pasting the output of 'git show' starting with the commit ID and ending with the commit message, keeping the indentation, and adding you own comments before and s-o-b line after. Here's an example: https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/commit/eb6bf454f7f59cd5b86a11301498db254b37ca9a

vincentmli:
Hi @tohojo I changed the commit message according to your suggestion, please review :)

tohojo:
That's better, thanks! :)

PR #123, opened 2024-12-26
Ignore all executions

Rtoax:
Ignore executable files to prevent them from being added to the git tree by mistake


Rtoax:
> Hmm, we generally do this by having separate .gitignore files in the subdirectories (see, e.g., the nat64 directory). Could you please split them out into separate files for consistency?

Ok！

Rtoax:
> Hmm, we generally do this by having separate .gitignore files in the subdirectories (see, e.g., the nat64 directory). Could you please split them out into separate files for consistency?

Done, thanks.

tohojo (approved):
Better, thanks! :)

PR #125, opened 2025-01-10
Add netstacklat tool

simosund:
Initial implementation of the netstacklat tool, which is based on Jesper Brouers's bpftrace script for measuring latency up to certain certain points in the Linux network stack.

The eBPF portion of the program is designed to be compatible with [ebpf_exporter](https://github.com/cloudflare/ebpf_exporter). In theory one only has to drop in the netstack.bpf.c file into the `ebpf_exporter/examples` directory and add the corresponding `yaml` config file (which I will probably add to this repository as well later), although have not yet verified that it actually works as intended there.

In addition to finishing up existing commits, some points I wish to address before attempting to merge this request is:
- Figure out how to use vmlinux.h properly in this setup (or if it's better to get definition of sk_buffer some other way)
- Add option to the userspace program to enable software timestamping on select interface(s) to make it a bit easier to use? Currently you're unlikely to actually get any useful data out of this unless you enable software timestamping on relevant interfaces externally (as skb->tstamp will just be 0).
- Verify that it works with ebpf_exporter and add yaml config file.


simosund:
So, I now belive I have a full "minimal working example" that quite closely replicates the bpftrace script and prints out its output in a similar format. There are still many missing features that would be nice to have, such as:
- Parse some command-line arguments, letting the user configure e.g. the reporting interval (current hard coded to 10s)
- Automatically enable software RX timestamping (on all interfaces or interfaces provided as command line arguments)
- Automatically figure out the TAI-offset at start up (currently hard coded as 37 seconds)
- Figure out clock basis for `skb->tstamp` (currently assumes `CLOCK_REALTIME`, but on at least one machine that did not seem to be the case), and possibly if it's valid (right now it assumes any non-zero timestamps are valid)

But figure that this should be sufficent for an early review at least.

One issue worth highlighting is that it's currently not fully compatible with ebpf-exporter in the sense that you can't just drop in the `netstacklat.bpf.c` file. First of there are some shared defines and enums in `netstacklat.h` that `netstacklat.bpf.c` depends on, so unless you also copy over that header (which seems to go against the style of the existing examples at [ebpf-exporter/examples](https://github.com/cloudflare/ebpf_exporter/tree/master/examples)) you need to first insert those into the `netstacklat.bpf.c` file. Furthermore, `netstacklat.bpf.c` does `#include "vmlinux_local.h"` as expected for this bpf-examples repository, but the ebpf-exporter examples seem to expect `#include <vmlinux.h>` instead. Not sure if there's any clever pre-processor magic or other tricks you could use to solve this.

For licensing, I've put my own files under GPL-2.0 as most other examples here. However, I've also included the same version of `bits.bpf.h` header from [bcc/libbpf-tools](https://github.com/iovisor/bcc/tree/master/libbpf-tools) (same version as ebpf-exporter uses), which then retains its dual LGPL/BSD license, and `maps.bpf.h` from ebpf-exporter that I put under MIT as that seems to be what ebpf-exporter in general uses.

As I mentioned during the meeting, I intentionally opted to use separate BPF maps for each histogram rather than multiplex multiple histograms in the same map as most of the ebpf-exporter examples seem to do, e.g. [here](https://github.com/cloudflare/ebpf_exporter/blob/99d2752f9e0a095b57f53e5da6856ad143b0e443/examples/biolatency.bpf.c#L35-L40). By doing so I can simply use array maps instead of hashmaps and thereby reduce the overhead for map lookups a bit.

simosund:
Oh, and ping @netoptimizer if Github hasn't already done so (not intending to rush you, just want to make sure you're aware this PR is ready to be reviewed).

simosund:
While I've done a couple of force pushes, there's been no major change to the initial 3 commits (which was what I originally requested the review for). I've just done some minor cleanup there (renamed an enum value from `NETSTACK_HOOK_TCP_V4_DO_RCV` to `NETSTACKLAT_HOOK_TCP_V4_DO_RCV` to be consistent with the others, and prefixed the commit messages with `netstacklat:`).

I have now also added 3 additional commits implementing most of the features I mentioned were missing in my previous comment. They are all relatively self-contained I can move them out into a separate PR(s) and/or reorder them if requested. 
There are however some aspects of those later PRs that I would also like to get some feedback on, so I've included them here for the moment (as it seems weird starting up a PR for additions to another PR that hasn't been merged yet). 

For the commit about enabling the RX software timestamps, it relies on the user space component to do that, so it won't really work when just using the `netstacklat.bpf.c` file together with ebpf-exporter. While @tohojo pointed out that it seems like they're perhaps [working](https://lore.kernel.org/all/20250112113748.73504-1-kerneljasonxing@gmail.com/) on some way of allowing you to enable `SO_TIMESTAMPING` from eBPF programs (a bit unclear if that's what it actually does though, and only seems to be for the TX path for now), I don't think we can rely on that as it will be quite some time before that makes it into most shipped kernels.

An alternative approach would be to add an eBPF program at some early part of the network stack that checks if the skb->tstamp field is set, and if not sets it itself. Although I guess not all hooks are necessarily allowed to write to that field, and I'm not sure if there may be any consequences higher up the stack for setting the tstamp field when its not expected to be set.

For the commit about initializing the TAI-offset, it has a similar limitation in that it relies on the user space program, so won't work for ebpf-exporter. As an alternative, it may be possible to get the TAI offset from an eBPF program instead. E.g. the following bpftrace script can get it:
```console
sudo bpftrace -e '
// On kernels >= 6.13, remove this struct definition (no longer an anonymous struct)
struct tk_data {
	seqcount_raw_spinlock_t	seq;
	struct timekeeper	timekeeper;
};

BEGIN
{
	printf("Getting clock offsets from core...\n");

	$tk = (struct tk_data *)kaddr("tk_core");
	printf("MONOTONIC -> REALTIME (offs_real): %ld\n",
	       $tk->timekeeper.offs_real);
	printf("MONOTONIC -> TAI (offs_tai): %ld\n",
	       $tk->timekeeper.offs_tai);
	printf("REALTIME -> TAI (tai_offset): %ld\n",
	       $tk->timekeeper.tai_offset);

	exit();
}'
```
However, I'm not really sure how to translate that to C/libbpf in a good way. First it needs to be attached to some suitable hook that ideally only runs once before all the other hooks trigger so that we don't have to probe that part of kernel memory every time we need to use the TAI offset. And then I'm not really sure if there's really anything corresponding the the bpftrace `kaddr()` function you can use in C/libbpf. Through a brief look it seems to lookup the /proc/kallsyms "file" and get the kernel address from there. 

Furthermore, the TAI offset is currently only set at startup, so if the program is running when a leap second occurs, the TAI offset will be out of sync, and we will get massively inflated latencies. We could periodically update it, although that still runs the risk of it being out of sync in-between updates. We could also try to hook into whatever part of the kernel that actually updates that offset, so that we change it as soon as the kernel updates it. But in order to update it, it can obviously no longer be a constant, so it will potentially add a small amount of overhead when using the TAI offset.

simosund:
When it comes to output format, I have for now approximately replicated the output format from @netoptimizer's bpftrace script. So it may for example look like this:
```console
Mon Jan 27 20:02:01 2025
tcp_v4_do_rcv:
[1.02us, 2.05us)       1054 |                                                |
[2.05us, 4.1us)      181795 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[4.1us, 8.19us)      181529 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[8.19us, 16.4us)      68458 |@@@@@@@@@@@@@@@@@@                              |
[16.4us, 32.8us)       1187 |                                                |
[32.8us, 65.5us)        990 |                                                |
[65.5us, 131us)         453 |                                                |
[131us, 262us)            8 |                                                |
Sum: 2373204797

tcp_data_queue:
[2.05us, 4.1us)      169968 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@   |
[4.1us, 8.19us)      179638 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[8.19us, 16.4us)      51028 |@@@@@@@@@@@@@@                                  |
[16.4us, 32.8us)       1176 |                                                |
[32.8us, 65.5us)        900 |                                                |
[65.5us, 131us)         433 |                                                |
[131us, 262us)            8 |                                                |
Sum: 2320708288

udp_queue_rcv_one_skb:
[1.02us, 2.05us)          8 |@@@@@@@@@@@@@@@@@                               |
[2.05us, 4.1us)          23 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[4.1us, 8.19us)          12 |@@@@@@@@@@@@@@@@@@@@@@@@@                       |
[8.19us, 16.4us)         14 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                   |
[16.4us, 32.8us)          7 |@@@@@@@@@@@@@@@                                 |
[32.8us, 65.5us)          2 |@@@@                                            |
Sum: 571657
```

(For reference, the corresponding output from the bpftrace script is:
```console
20:02:01
@cnt[kfunc:vmlinux:udp_queue_rcv_one_skb]: 66
@cnt[kfunc:vmlinux:tcp_data_queue]: 404016
@cnt[kfunc:vmlinux:tcp_v4_do_rcv]: 436508
@total[kfunc:vmlinux:udp_queue_rcv_one_skb]: 66
@total[kfunc:vmlinux:tcp_data_queue]: 404022
@total[kfunc:vmlinux:tcp_v4_do_rcv]: 436514
@stats[kfunc:vmlinux:tcp_v4_do_rcv]: count 436515, average 5148, total 2247484837
@stats[kfunc:vmlinux:tcp_data_queue]: count 404024, average 5489, total 2217754970
@stats[kfunc:vmlinux:udp_queue_rcv_one_skb]: count 66, average 6507, total 429464

@hist_ns[kfunc:vmlinux:udp_queue_rcv_one_skb]: 
[1K, 2K)              10 |@@@@@@@@@@@@@@@@@@@@                                |
[2K, 4K)              26 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[4K, 8K)              11 |@@@@@@@@@@@@@@@@@@@@@@                              |
[8K, 16K)             13 |@@@@@@@@@@@@@@@@@@@@@@@@@@                          |
[16K, 32K)             5 |@@@@@@@@@@                                          |
[32K, 64K)             1 |@@                                                  |

@hist_ns[kfunc:vmlinux:tcp_data_queue]: 
[1K, 2K)              25 |                                                    |
[2K, 4K)          175047 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ |
[4K, 8K)          177977 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[8K, 16K)          48539 |@@@@@@@@@@@@@@                                      |
[16K, 32K)          1135 |                                                    |
[32K, 64K)           880 |                                                    |
[64K, 128K)          420 |                                                    |
[128K, 256K)           7 |                                                    |

@hist_ns[kfunc:vmlinux:tcp_v4_do_rcv]: 
[256, 512)             1 |                                                    |
[512, 1K)              4 |                                                    |
[1K, 2K)           28061 |@@@@@@@                                             |
[2K, 4K)          158756 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@        |
[4K, 8K)          185669 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[8K, 16K)          61565 |@@@@@@@@@@@@@@@@@                                   |
[16K, 32K)          1072 |                                                    |
[32K, 64K)           946 |                                                    |
[64K, 128K)          442 |                                                    |
[128K, 256K)           8 |                                                    |

```
)

So it "visualizes" the data in a format that makes it fairly easy to glance at, but not really ideal if you want to parse it and feed it into some other system. But if you want to feed the data into some other system it's perhaps better to use it together with ebpf-exporter in the first place.

simosund:
I've now added the ebpf-exporter YAML config file. 

Furthermore, in the "Make build process compatible with ebpf-exporter" commit I have provided a suggestion for how to make netstacklat.bpf.c compile correctly both here in the bpf-examples repository as well as in when copied into the `ebpf_exporter/examples` directory. You will still have to also copy over the `netstacklat.h` file, which breaks the style compared to the other exampels in `ebpf_exporter/examples` a bit, but works without having to actually modify any file.

netstacklat/maps.bpf.h
```diff
@@ -0,0 +1,84 @@
+/* SPDX-License-Identifier: MIT */
+/* From https://github.com/cloudflare/ebpf_exporter/blob/99d2752f9e0a095b57f53e5da6856ad143b0e443/examples/maps.bpf.h */
```
tohojo, line 2: Okay, so a bunch of macros that don't properly delimit their contents, even going so far as having later macros directly refer to the expanded values of earlier ones? Yikes, no thanks! Please get rid of these and reimplement them using proper inline functions, or at least macros that are sane :)

netstacklat/netstacklat.bpf.c
```diff
@@ -0,0 +1,100 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+#ifdef BPF_EXAMPLES
```
tohojo, line 2: I think this is probably going a bit too far in the compatibility. Let's just drop this commit for now, and circle back to it once we have an idea for how things will look wrt bpf-exporter.

tohojo (changes requested):
I think generally this looks sane, with the main exception being the macros from bpf_exporter. Those we should definitely not import!

simosund:
I've now done a few sanity tests to validate that it seems to able to capture latency increases in the network stack. Simply used 10 parallel iperf flows to generate traffic and did a test without iptables rules, and another test after adding ~25k pointless iptables rules. 

These first graphs are without the iptables rules.
image

These graphs are with the 25k iptables rules.
image

Note that the median and 95th percentile values should likely be around twice as high due to the issue with how ebpf-exporter bins the values that I previously mentioned. 

While these graphs may not be the easiest to read, the latency one can note that the latency increases quite a bit after adding the 25k iptables rules, with e.g. the median roughly doubling and the mean for tcp_v4_do_rcv() increasing from around 12us to 200us (the mean for tcp_data_queue() actually drops from around 12us to 6us though). The heatmap also shows that the distribution for tcp_v4_do_rcv() shifts from a rather unimodal distribution to a much more multimodal one. Interestingly enough, the relationship between how frequently tcp_v4_do_rcv() and tcp_data_queue() also shifts, where tcp_v4_do_rcv() only runs around 15% more frequently than tcp_data_queue() without all the iptables rules, whereas it runs about twice as frequently with the 25k rules (where the rate for tcp_data_queue() seems to match the actual packet rate).

These tests, however, also somewhat highligted that the log2 histograms are indeed very coarse grained, as a latency increase will generally only be noticible if it roughly doubles the latency (as that's what's required to end up in another histogram bin). And in this case the impact of adding those 25k iptables rules could for example much easier be noticed by the packet rate dropping from around 260 kpps to 24 kpps. Instead of log2 histograms we could of course use something more fine-grained, but already at our current 30 buckets we already exceed what's typically used with Prometheus histograms. 

In the last year or so Prometheus has added support for "[native histograms](https://prometheus.io/docs/specs/native_histograms/)" that can handle a large amount of bins much more efficently. However, that is still an experimental feature and not supported by ebpf-exporter.

netstacklat/maps.bpf.h
simosund: Hmm, well for just using it here in bpf-examples it would be easy to implement these as functions (not sure if I even need to specify them as inlined, generally leave that decision to the compiler).

However, if the goal is to be compatible with ebpf-exporter, maybe it makes more sense to try and fix this upstream (and then import the fixed version of these)? For that it will probably still have to be macros as the type of the key is unknown, it's just assumed to be a struct that has the member `bucket` (which they simply access with .bucket). Not very experienced in writing anything but the most trivial macros, so might have to bounce this back and forth a few times before I get this right if we go down this path.

Could of coure still be handled with functions that accept a void pointer to the key, as long as we decide that the bucket member has to be the first member. However, ebpf-exporter currently seem to assume that the bucket is the the last member of the key (ran into some issues when I tried placing something after the bucket member), so that would likely require changes to some interal aspects of ebpf-exporter as well as updating all of their examples.

simosund:
Have also looked a little bit at the performance. Based on the bpf stats (`sysctl -w kernel.bpf_stats_enabled=1`), the overhead from the eBPF probes for the current implementation looks as follows:
```csv
id   , name                      ,        count, avg runtime (ns), frequency (Hz), est. CPU util (%)
2093 , netstacklat_tcp_v4_do_rcv ,     93309605,           176.03,      310702.88,         5.469346%
2095 , netstacklat_tcp_data_queue,     82011162,           134.08,      273081.26,         3.661600%
2096 , netstacklat_udp_queue_rcv ,          255,           762.80,           0.85,         0.000065%
0    , Total                     ,    175321022,           156.41,      583784.99,         9.131011%
```
Where count is the number of times the hook ran, avg runtime is the processing time per hook activation, frequency is how frequently the hook was triggered, and estimated CPU utilization is the single core CPU utilization for the program given at that runtime and frequency (i.e. runtime * frequency).

So the tcp_v4_do_rcv() hook takes up around 175ns per activation, while the tcp_data_queue() hook takes up around 135ns per activation while under load (the 762ns for udp_queue_rcv() is a bit inflated as it's not really under any load, if I redid the tests with a UDP traffic load I expect it would drop to a similar level). 

I also tested with a version where I multiplex all the histograms in a single hashmap (how ebpf-exporter typically does it, and produces a more easily handled Prometheus metric), which gave the following results:
```csv
id   , name                      ,        count, avg runtime (ns), frequency (Hz), est. CPU util (%)
2057 , netstacklat_tcp_v4_do_rcv ,     94031163,           266.09,      313093.05,         8.331175%
2059 , netstacklat_tcp_data_queue,     82230810,           192.18,      273801.73,         5.261953%
2060 , netstacklat_udp_queue_rcv ,          131,           832.94,           0.44,         0.000036%
0    , Total                     ,    176262104,           231.61,      586895.21,        13.593164%
```
So about 60ns (40%) more for tcp_data_queue() and 90ns (50%) more for tcp_v4_do_rcv(). 

I also tested a dummy version that just picked a fixed value for the delay instead of getting it from reading the skb timestamp and comparing to the current time to get a sort of lower floor for what the maximum performance could reasonably be (only overhead for adding the value to the histogram, using the separate array maps), which gave the following:
```csv
id   , name                      ,        count, avg runtime (ns), frequency (Hz), est. CPU util (%)
276  , netstacklat_tcp_v4_do_rcv ,     89714195,            70.09,      297714.38,         2.086633%
278  , netstacklat_tcp_data_queue,     76804520,            55.17,      254873.94,         1.406174%
279  , netstacklat_udp_queue_rcv ,          379,           373.21,           1.26,         0.000047%
0    , Total                     ,    166519094,            63.21,      552589.58,         3.492854%
``` 
So just parsing the the skb timestamp and getting the current time seems to make up the majority of the overhead in the current implementation.

One interesting observation here is that the overhead from the tcp_v4_do_rcv() hook (per activation) seems to be consistently higher than that for tcp_data_queue(), despite all hooks running essentially the same code and all of them being fentry hooks. Furthermore, the it seems that tcp_v4_do_rcv() hook takes roughly about 30% longer than tcp_data_queue() in all these configurations, rather than being a fixed overhead (e.g. 40ns).

As some reference points, I also ran similar tests with epping and the [Kyanos](https://kyanos.io/) tool, result are as follows:
epping:
```csv
id   , name            ,        count, avg runtime (ns), frequency (Hz), est. CPU util (%)
189  , pping_tc_egress ,     79354117,           583.30,      251995.04,        14.698873%
191  , pping_tc_ingress,     15509659,           829.35,       49252.10,         4.084723%
192  , tsmap_cleanup   ,       966946,           427.65,        3219.48,         0.137680%
193  , flowmap_cleanup ,         3289,           323.04,          10.95,         0.000354%
0    , Total           ,     95834011,           621.54,      304477.57,        18.921630%
```

kyanos:
```csv
id   , name                                      ,        count, avg runtime (ns), frequency (Hz), est. CPU util (%)
2326 , dev_hard_start_xmit                       ,     91032535,           837.27,      302098.44,        25.293666%
2333 , security_socket_recvmsg_enter             ,     37333410,            91.60,      123893.78,         1.134898%
2334 , security_socket_sendmsg_enter             ,     28416279,            87.66,       94301.60,         0.826667%
2337 , sock_alloc_ret                            ,         1737,           713.72,           5.76,         0.000411%
2338 , tcp_destroy_sock                          ,           98,          2661.88,           0.33,         0.000087%
2341 , tcp_v4_do_rcv                             ,     86890360,           658.30,      288352.32,        18.982194%
2343 , tracepoint__netif_receive_skb             ,     91034180,           771.08,      302103.90,        23.294533%
2344 , tracepoint__sched__sched_process_exec     ,           13,         12189.92,           0.04,         0.000053%
2346 , tracepoint__skb_copy_datagram_iovec       ,     94955297,           441.97,      315116.43,        13.927324%
2347 , tracepoint__syscalls__sys_enter_accept4   ,          140,          1562.58,           0.46,         0.000073%
2348 , tracepoint__syscalls__sys_enter_close     ,        43195,           444.89,         143.35,         0.006377%
2349 , tracepoint__syscalls__sys_enter_connect   ,          363,           921.93,           1.20,         0.000111%
2350 , tracepoint__syscalls__sys_enter_read      ,       105619,           660.55,         350.50,         0.023153%
2352 , tracepoint__syscalls__sys_enter_recvfrom  ,     37107108,           350.08,      123142.78,         4.311032%
2353 , tracepoint__syscalls__sys_enter_recvmsg   ,       221419,           779.13,         734.80,         0.057250%
2355 , tracepoint__syscalls__sys_enter_sendmsg   ,        71460,           829.18,         237.15,         0.019664%
2356 , tracepoint__syscalls__sys_enter_sendto    ,         1426,           699.82,           4.73,         0.000331%
2357 , tracepoint__syscalls__sys_enter_write     ,     28387953,           357.42,       94207.60,         3.367143%
2359 , tracepoint__syscalls__sys_exit_accept4    ,          140,          2785.14,           0.46,         0.000129%
2360 , tracepoint__syscalls__sys_exit_close      ,        43195,           375.93,         143.35,         0.005389%
2361 , tracepoint__syscalls__sys_exit_connect    ,          363,          2790.62,           1.20,         0.000336%
2362 , tracepoint__syscalls__sys_exit_read       ,       105620,           474.41,         350.51,         0.016628%
2364 , tracepoint__syscalls__sys_exit_recvfrom   ,     37107108,           850.75,      123142.78,        10.476430%
2365 , tracepoint__syscalls__sys_exit_recvmsg    ,       221419,          1367.78,         734.80,         0.100504%
2367 , tracepoint__syscalls__sys_exit_sendmsg    ,        71460,          1540.79,         237.15,         0.036539%
2368 , tracepoint__syscalls__sys_exit_sendto     ,         1426,          1829.03,           4.73,         0.000866%
2369 , tracepoint__syscalls__sys_exit_write      ,     28387952,           924.71,       94207.59,         8.711466%
0    , Total                                     ,    561541275,           593.46,     1863517.74,       110.593253%
```

So in summary, current overhead is fairly low, but still not insignificant. Only adding a few more hook points will likely result in overhead similar to that of epping (in a setup where epping does 3-4 hashmap lookups per packet), although we still have a lot of leg room left until we reach as high overhead as kyanos (but kyanos can also deliver much more detailed data).

simosund:
With a bunch of help from @skyfmmf I was also able to get lattrace running (although without some hooks that are not available on my kernel). So as an additional reference, here is a similar performance test for lattrace:
```csv
id   , name                     ,        count, avg runtime (ns), frequency (Hz), est. CPU util (%)
836  , trace_alloc_skb          ,     27571636,          1124.57,      229129.33,        25.767200%
839  , tp_napi_gro_receive_entry,          900,          1759.16,           7.48,         0.001316%
840  , trace_clone_skb          ,     20855241,          1138.62,      173313.89,        19.733806%
841  , tp_consume_skb           ,     10600335,          1203.15,       88092.26,        10.598854%
842  , tp_kfree_skb             ,         1303,          2020.93,          10.83,         0.002188%
843  , trace_kfree_skb          ,     25886742,          1023.87,      215127.31,        22.026283%
846  , start_conntrack          ,     27532368,           974.68,      228803.00,        22.300972%
847  , end_conntrack            ,     27532368,           919.37,      228803.00,        21.035489%
850  , netif_receive            ,     27532563,           946.98,      228804.62,        21.667320%
851  , netdev_queue             ,     27531724,          1040.48,      228797.65,        23.806037%
852  , tcp_receive              ,     20853296,          1008.86,      173297.72,        17.483325%
853  , udp_receive              ,          112,          1420.97,           0.93,         0.000132%
0    , Total                    ,    215898588,          1027.89,     1794188.00,       184.422921%
```

So about ~1000ns per hook for lattrace compared to around ~150ns per hook for netstacklat (although the work lattrace actually does varies a little bit based on the hook, while netstacklat essentially runs the same code for every hook).

simosund:
Hi @tohojo @netoptimizer, I've now updated this PR to address Toke's previous comments, so please re-review. 

I explain my rationale regarding how I replaced the ebpf-exporter macros in the initial commit while trying to keep it as compatible as possible. I've also done some other minor changes, e.g. renamed the maps to follow the [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/) (as the map names are directly converted to metric names by ebpf-exporter).

simosund:
Can also mention that as a bit of a side-project also wanted to explore whether my idea regarding doing a binary search for finding the correct histogram bins was viable, which can be found my [netstacklat-binarysearch](https://github.com/simosund/bpf-examples/tree/netstacklat-binarysearch) branch.

It supports much higher resolution histograms, like e.g. using the exponential base 2^(1/8) instead of 2 (corresponding to schema/scale 3 of the Prometheus/OpenTelemetry histogram bucketing scheme). So instead of:
```console
$ sudo netstacklat --histgram-exp-base 2
tcp_data_queue:
(256ns, 512ns]       589843 |                                                |
(512ns, 1.02us]    60651018 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
(1.02us, 2.05us]    4659890 |@@@@                                            |
(2.05us, 4.1us]     1709485 |@                                               |
(4.1us, 8.19us]     2736061 |@@                                              |
(8.19us, 16.4us]    2152283 |@@                                              |
(16.4us, 32.8us]    2192421 |@@                                              |
(32.8us, 65.5us]    1127523 |@                                               |
(65.5us, 131us]       77367 |                                                |
(131us, 262us]        61477 |                                                |
(262us, 524us]        70728 |                                                |
(524us, 1.05ms]      120540 |                                                |
(1.05ms, 2.1ms]       95772 |                                                |
(2.1ms, 4.19ms]       93887 |                                                |
(4.19ms, 8.39ms]      33511 |                                                |
(8.39ms, 16.8ms]       3352 |                                                |
(16.8ms, 33.6ms]         47 |                                                |
count: 76375205, average: 12.98us
```

You can get something like this:
```console
$ sudo netstacklat --histgram-exp-base 1.09050773267 --histogram-bins 256
tcp_data_queue:
(331ns, 362ns]            6 |                                                |
(362ns, 394ns]         1777 |                                                |
(394ns, 430ns]         8627 |                                                |
(430ns, 469ns]         9955 |                                                |
(469ns, 512ns]       172417 |@                                               |
(512ns, 558ns]      1126030 |@@@@                                            |
(558ns, 608ns]      4774267 |@@@@@@@@@@@@@@@@@                               |
(608ns, 663ns]      8391747 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                  |
(663ns, 724ns]     10553781 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@           |
(724ns, 789ns]     13523505 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
(789ns, 861ns]     10200965 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@            |
(861ns, 939ns]      6171181 |@@@@@@@@@@@@@@@@@@@@@@                          |
(939ns, 1.02us]     3884627 |@@@@@@@@@@@@@@                                  |
(1.02us, 1.12us]    2440684 |@@@@@@@@@                                       |
(1.12us, 1.22us]    1475033 |@@@@@                                           |
(1.22us, 1.33us]     916458 |@@@                                             |
(1.33us, 1.45us]     578992 |@@                                              |
(1.45us, 1.58us]     415995 |@                                               |
(1.58us, 1.72us]     352531 |@                                               |
(1.72us, 1.88us]     320667 |@                                               |
(1.88us, 2.05us]     298565 |@                                               |
(2.05us, 2.23us]     275586 |@                                               |
(2.23us, 2.44us]     258181 |@                                               |
(2.44us, 2.65us]     242228 |@                                               |
(2.65us, 2.9us]      239862 |@                                               |
(2.9us, 3.16us]      217436 |@                                               |
(3.16us, 3.44us]     197560 |@                                               |
(3.44us, 3.76us]     217922 |@                                               |
(3.76us, 4.1us]      257362 |@                                               |
(4.1us, 4.47us]      249781 |@                                               |
(4.47us, 4.87us]     256871 |@                                               |
(4.87us, 5.31us]     287401 |@                                               |
(5.31us, 5.79us]     307688 |@                                               |
(5.79us, 6.32us]     333881 |@                                               |
(6.32us, 6.89us]     375341 |@                                               |
(6.89us, 7.51us]     446672 |@@                                              |
(7.51us, 8.19us]     489102 |@@                                              |
(8.19us, 8.93us]     414785 |@                                               |
(8.93us, 9.74us]     327430 |@                                               |
(9.74us, 10.6us]     286621 |@                                               |
(10.6us, 11.6us]     272253 |@                                               |
(11.6us, 12.6us]     277478 |@                                               |
(12.6us, 13.8us]     284322 |@                                               |
(13.8us, 15us]       278177 |@                                               |
(15us, 16.4us]       264872 |@                                               |
(16.4us, 17.9us]     261153 |@                                               |
(17.9us, 19.5us]     277002 |@                                               |
(19.5us, 21.2us]     267395 |@                                               |
(21.2us, 23.2us]     262806 |@                                               |
(23.2us, 25.3us]     276017 |@                                               |
(25.3us, 27.6us]     291072 |@                                               |
(27.6us, 30us]       315936 |@                                               |
(30us, 32.8us]       339252 |@                                               |
(32.8us, 35.7us]     329475 |@                                               |
(35.7us, 39us]       276416 |@                                               |
(39us, 42.5us]       201996 |@                                               |
(42.5us, 46.3us]     133870 |                                                |
(46.3us, 50.5us]      86340 |                                                |
(50.5us, 55.1us]      58442 |                                                |
(55.1us, 60.1us]      40683 |                                                |
(60.1us, 65.5us]      26194 |                                                |
(65.5us, 71.5us]      16038 |                                                |
(71.5us, 77.9us]      11234 |                                                |
(77.9us, 85us]         9562 |                                                |
(85us, 92.7us]         8762 |                                                |
(92.7us, 101us]        8900 |                                                |
(101us, 110us]         8927 |                                                |
(110us, 120us]         8781 |                                                |
(120us, 131us]         9119 |                                                |
(131us, 143us]         9436 |                                                |
(143us, 156us]         9522 |                                                |
(156us, 170us]         9200 |                                                |
(170us, 185us]         8495 |                                                |
(185us, 202us]         7655 |                                                |
(202us, 220us]         6739 |                                                |
(220us, 240us]         6612 |                                                |
(240us, 262us]         6640 |                                                |
(262us, 286us]         7080 |                                                |
(286us, 312us]         7571 |                                                |
(312us, 340us]         7907 |                                                |
(340us, 371us]         8525 |                                                |
(371us, 404us]         9278 |                                                |
(404us, 441us]        10301 |                                                |
(441us, 481us]        10933 |                                                |
(481us, 524us]        11935 |                                                |
(524us, 572us]        12821 |                                                |
(572us, 623us]        14058 |                                                |
(623us, 680us]        14771 |                                                |
(680us, 741us]        15148 |                                                |
(741us, 809us]        16151 |                                                |
(809us, 882us]        16425 |                                                |
(882us, 962us]        16207 |                                                |
(962us, 1.05ms]       15798 |                                                |
(1.05ms, 1.14ms]      15109 |                                                |
(1.14ms, 1.25ms]      14756 |                                                |
(1.25ms, 1.36ms]      13631 |                                                |
(1.36ms, 1.48ms]      13644 |                                                |
(1.48ms, 1.62ms]      12638 |                                                |
(1.62ms, 1.76ms]      10497 |                                                |
(1.76ms, 1.92ms]      10261 |                                                |
(1.92ms, 2.1ms]       11232 |                                                |
(2.1ms, 2.29ms]       12180 |                                                |
(2.29ms, 2.49ms]      12824 |                                                |
(2.49ms, 2.72ms]      13276 |                                                |
(2.72ms, 2.97ms]      11857 |                                                |
(2.97ms, 3.23ms]      10195 |                                                |
(3.23ms, 3.53ms]      11539 |                                                |
(3.53ms, 3.85ms]      13821 |                                                |
(3.85ms, 4.19ms]      13919 |                                                |
(4.19ms, 4.57ms]       6205 |                                                |
(4.57ms, 4.99ms]       6264 |                                                |
(4.99ms, 5.44ms]       5116 |                                                |
(5.44ms, 5.93ms]       3677 |                                                |
(5.93ms, 6.47ms]       3522 |                                                |
(6.47ms, 7.05ms]       4314 |                                                |
(7.05ms, 7.69ms]       3087 |                                                |
(7.69ms, 8.39ms]       2182 |                                                |
(8.39ms, 9.15ms]       1058 |                                                |
(9.15ms, 9.98ms]        680 |                                                |
(9.98ms, 10.9ms]        353 |                                                |
(10.9ms, 11.9ms]        231 |                                                |
(11.9ms, 12.9ms]        132 |                                                |
(12.9ms, 14.1ms]         64 |                                                |
(14.1ms, 15.4ms]         22 |                                                |
(15.4ms, 16.8ms]         59 |                                                |
(16.8ms, 18.3ms]          3 |                                                |
(18.3ms, 20ms]           15 |                                                |
(20ms, 21.8ms]          100 |                                                |
(21.8ms, 23.7ms]         93 |                                                |
(23.7ms, 25.9ms]        132 |                                                |
(25.9ms, 28.2ms]         19 |                                                |
(28.2ms, 30.8ms]          3 |                                                |
count: 76699912, average: 13.41us
```

It does add some overhead, although it's not catastrophically high (performance comparison can be found in the commit message). 

However, it also makes it a bit of a pain to use together with ebpf-exporter, as it's the userspace program that calculates and provides the bin-limits that the eBPF program uses, so without it you have to manually configure all the bin limits.

Probably not the way to go, especially as we realistically can't use that many buckets with ebpf-exporter anyways. But still thought it was interesting to explore, and if nothing else I learned that the verifier is actually pretty good at figuring out if loops will terminate (have a while loop in there that's not completely obvious that it will terminate).

simosund:
In the last force push (2025-02-20T09:30 CET, commit starting at 1118029) I essentially squashed in a small change to the structure of the eBPF code that makes it easier to reuse the core functionality (recording latency until now) for other types of hooks (see commit c649f20 on the [netstacklat-more-probes](https://github.com/simosund/bpf-examples/tree/netstacklat-more-probes) branch for more details).

simosund:
Converted to draft while I prepare an initital up-to-date version of the tool that can be merged. Will reopen once that is ready for review.

simosund:
Hi @tohojo, @netoptimizer. I've now updated this PR contain an up-to-date version of netstacklat with all features that I consider ready to be merged. Please review when you have time.

This excludes the tracking of standing queues (as that is still rather experimental and based on my own testinging doesn't work super well), and the rework to multiplex all histograms in a single hashmap instead of separate array maps (still needs some more refinement, and more extensive evaluation of overhead cost). These can be found in my  branches [netstacklat-queue-detection](https://github.com/simosund/bpf-examples/tree/netstacklat-queue-detection) and [netstacklat-multiplex-histograms](https://github.com/simosund/bpf-examples/tree/netstacklat-multiplex-histograms) (have not yet rebased them to build on top of this updated branch).

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1049 @@
...
+
+		optlen = snprintf(
+			optstr, sizeof(optstr), "%c%s", long_options[i].val,
+			long_options[i].has_arg == optional_argument ? "::" :
+			long_options[i].has_arg == required_argument ? ":" :
+								       "");
+		if (strlen + optlen < size) {
+			strncpy(buf + strlen, optstr, optlen + 1);
+		}
+		strlen += optlen;
```
tohojo, line 104: I don't think this works correctly if it actually runs out of space. It's incrementing strlen regardless of whether it runs out of space, and it also keeps looping in this case. The use of `snprintf()` without checking the return value works, but requires a bit of thought to check that it's correct (due to the fixed-size inputs).

I almost think the open-coded version is more readable:
```c
int generate_optstr(char *buf, size_t size) {

	char *end = buf + size - 1, *p = buf;
	int i;

        for (i = 0; long_options[i].name; i++) {
		if (long_options[i].flag || !isalnum(long_options[i].val))
			continue;

                *p++ = (unsigned char)long_options[i].val;
                if (p >= end)
			break;

                if (long_options[i].has_arg) {
			*p++ = ':';
			if (p >= end)
                          break;
		}

                if (long_options[i].has_arg == optional_argument) {
			*p++ = ':';
			if (p >= end)
                          break;
		}
        }
        *p++ = '\0';

        return (p - buf);
}
```

...but up to you, as long as it behaves correctly if it runs out of space :)

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1049 @@
...
+	__u32 pids[MAX_FILTER_PIDS];
+};
+
+static const struct option long_options[] = {
+	{ "help",            no_argument,       NULL, 'h' },
+	{ "report-interval", required_argument, NULL, 'r' },
+	{ "list-probes",     no_argument,       NULL, 'l' },
+	{ "enable-probe",    required_argument, NULL, 'e' },
+	{ "disable-probe",   required_argument, NULL, 'd' },
+	{ "filter-pid",      required_argument, NULL, 'p' },
```
tohojo, line 71: And likewise for this (`filter-pids`)...

tohojo: Also, I would expect the PIDs I supply as "filter-pids" arguments to be *removed* from the output, not the reverse. For this usage, maybe call it `only-pids` instead (or something similar)?

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1049 @@
...
+			hooks_off = true;
+			break;
+		case 'p': // filter-pids
+			err = parse_pids(ARRAY_SIZE(conf->pids) - conf->npids,
+					 conf->pids + conf->npids, optarg,
+					 optval_to_longopt(opt)->name);
+			if (err < 0)
+				return err;
+
+			conf->npids += err;
```
tohojo, line 455: If a variable is used both as a return value and an error value, it's better to name it `ret`. This looks like you're assigning an error value to something meaningful.

netstacklat/netstacklat.c
simosund: Depends on what you consider correct behavior when running out of space. My intended behavior is that it only writes in the buffer up to the provided size, but still returns the full length of the string that would be written if it did not have to be truncated, similar to e.g. `snprintf`, allowing you to detect that the string was truncated (and that I also check when it's used). As the `strncpy` operation actually writing to the provided buffer is guarded by size check I think it should not overflow the buffer, but to figure out what the full length of the string should be without the truncation I need to continue the loop.

But perhaps it's better to just simplify so it errors out in case the buffer is not sufficiently large. In that case I agree that your version is easier to follow. However, I'll then change so that it errors out instead of just silently truncating the output, as I do not like it silently failing (or truncating). Especially as your version may end up truncating it in the middle of an option, e.g. missing the ":" in an option with a required argument, converting it to a no-argument option (whereas my version would truncate it at an option boundary, so it may miss options but not change their meanings).

netstacklat/netstacklat.c
simosund: I assume your objection here is only regarding the name of the option, not its function. As having to supply the all the PIDs you're not interested in seems much more cumbersome than specifying the PIDs you're actually interested in most of the time. I can see how "filter" can be a bit ambiguous as it can be used both for excluding (filtering out) unwanted data or including (filtering for) wanted data. 

How about simply calling the option `pid` (or `pids` if we want to go with the plural version) similar to many other tools that offer similar functionality, e.g. `perf record` ?

netstacklat/netstacklat.bpf.c
```diff
@@ -0,0 +1,338 @@
...
+	__u8 mono_delivery_time: 1;
+} __attribute__((preserve_access_index));
+
+/*
+ * To be compatible with ebpf-exporter, all histograms need a key whose final
+ * member is the histogram bucket index
+ */
+struct hist_key {
+	u32 bucket;
+};
```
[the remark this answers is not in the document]
simosund: The [documentation](https://docs.kernel.org/bpf/map_array.html) for array maps says that the keys are always u32, so I assumed it wouldn't be valid to specify another key type for array maps. I just now tried what happens if I do `__type(key, struct hist_key)` instead, and then get "libbpf: Error in bpf_create_map_xattr(netstack_latency_ip_start_seconds):Invalid argument(-22). Retrying without BTF". It still works in the end, but without BTF, so not ideal.

I guess that whole struct is a bit weird, but it's just to try and maintain compatibility with ebpf-exporter macros (that you weren't a fan of), where the key must be a struct with a `bucket` member. In the current version I could drop that struct (and the corresponding parameter in the `increment_exp2_histogram_nosync()` function) if we do not care to maintain complete compatibility with those macros. It would still work together with ebpf-exporter, you just wouldn't be able to switch out my `increment_exp2_histogram_nosync()` function with the [corresponding ebpf-exporter macro](https://github.com/cloudflare/ebpf_exporter/blob/master/examples/maps.bpf.h#L78-L79).

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1049 @@
...
+	int npids;
+	__u32 pids[MAX_FILTER_PIDS];
+};
+
+static const struct option long_options[] = {
+	{ "help",            no_argument,       NULL, 'h' },
+	{ "report-interval", required_argument, NULL, 'r' },
+	{ "list-probes",     no_argument,       NULL, 'l' },
+	{ "enable-probe",    required_argument, NULL, 'e' },
+	{ "disable-probe",   required_argument, NULL, 'd' },
```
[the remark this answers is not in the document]
tohojo: Yeah, the comments and the commit message. But since the code expects multiple comma-separated arguments, it should definitely be plural. If it's singular I would expect to have to supply the argument multiple times (i.e., `--enable-probe ip-start --enable-probe tcp-start` vs `--enable-probes ip-start,tcp-start`)

netstacklat/netstacklat.c
tohojo: Right, I see. Okay, in that case I agree that your implementation is essentially correct. However, erroring out does seem like the less error-prone thing to do (and yeah, my version doesn't do that correctly, nice catch) :)

netstacklat/netstacklat.c
tohojo: Yeah, change the name. `--pids` is fine :)

netstacklat/netstacklat.c
simosund: Agree, I had just been a bit lazy and reused an existing variable.

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1049 @@
...
+/*
+ * Used to pack both a "type" and a value into the epoll_event.data.u64 member.
+ * The topmost bits indicates the type (SIG, TIMER, etc) while the remaining
+ * bits can be used for the value. The MASK can be used to filter out the
+ * type/value.
+ */
+#define NETSTACKLAT_EPOLL_SIG (1ULL << 63)
+#define NETSTACKLAT_EPOLL_TIMER (1ULL << 62)
+#define NETSTACKLAT_EPOLL_MASK \
+	(~(NETSTACKLAT_EPOLL_SIG | NETSTACKLAT_EPOLL_TIMER))
```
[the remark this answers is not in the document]
simosund: Agreed, although in this case you're partly reviewing yourself as it's based on your suggestions from the epping PR #59 =)

netstacklat/netstacklat.c
simosund: Have adopted your version now, but switching out all the breaks for `return -E2BIG`. 

Also changed the `p >= end` checks to `p > end` as with the post-increment it checks the position after the place it wrote to, and as `end` is the last valid position it should be OK to write something to the second last valid position (saving the last valid position for the terminating NULL).

In case it runs out of space it will not null-terminate the string, but figured it should be OK to just consider the contents of buf invalid in case it returns an error.

tohojo: Hmm, so the idea with the `>=` check is that there should also be room for the NULL byte at the end. A well-behaved string function should always ensure that it never leaves a buffer with an unterminated string (that is one of the problems with `snprintf()` and friends).

So you'll need a `goto err` that sets `p = '\0';` before returning :)

simosund: Have added an `err` section so that `buf` will still contain a valid string if we error out, even if it's not necessarily a correct optstring (as it in addition to missing some options it may also change the meanings of some options, as previously noted).

Have moved the checks to before we actually try to write anything to `p` so the `p <= end` logic properly works for reserving the last position for the NULL. Doesn't visually look as nice, but that way the check doesn't depend on if this is the last time we will append (a non-NULL byte) to buf or not (because if it's the last time, then having the next position be the last valid position is ok, which is why I changed it to `>` before so it wouldn't fail if the buffer was exactly large enough). That also ensures it works correctly for `size` 1 (which the previous post-checks would never be able to catch in time).

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1070 @@
...
+
+	if (size <= 0)
+		return -E2BIG;
+
+	for (i = 0; long_options[i].name; i++) {
+		if (long_options[i].flag || !isalnum(long_options[i].val))
+			continue;
+
+		if (p >= end) {
+			err = -E2BIG;
```
tohojo, line 100: ...you can get rid of the assignments here...

netstacklat/netstacklat.c
```diff
@@ -0,0 +1,1070 @@
...
+
+		if (long_options[i].has_arg == optional_argument) {
+			if (p >= end) {
+				err = -E2BIG;
+				goto err;
+			}
+			*p++ = ':';
+		}
+	}
+	*p++ = '\0';
```
tohojo, line 121: And turn the end of the function into the following:
```c
	ret = (p - buf) + 1;
out:
	*p = '\0';
	return ret;
```

tohojo (changes requested):
Looks good, just a few nits to save a few lines and make the flow a bit nicer...

PR #126, opened 2025-02-10
Add bpf2go example + Devcontainer example

David-VTUK:
[bpf2go](https://github.com/cilium/ebpf/tree/main/cmd/bpf2go) enables embedding and interacting with eBPF programs in Go applications. It does this by compiling, loading and managing eBPF programs by generating Go bindings from eBPF object files.

This is a simple example that includes both components:

* A eBPF kernel application that inspects incoming packets via `xdp`, recording the `protocol` value in the IP header and storing it in a map.

* A Go user space application that loads the eBPF application, reads from the map and outputs information via `stdout`, translating the IANA protocol number to its respective name. 

The `README.md` for the example contains further information

Additionally, a devcontainers example directory is added to help speed up eBPF development by facilitating repeatable, consistent environments to develop eBPF applications in.

Signed-off-by: David Holder [david.holder@gmail.com](mailto:david.holder@gmail.com)


xdp-bpf2go-example/main.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	// Load the compiled eBPF ELF and load it into the kernel.
+	var objs packetProtocolObjects
+	if err := loadPacketProtocolObjects(&objs, nil); err != nil {
+		log.Fatal("Loading eBPF objects:", err)
+	}
+	defer objs.Close()
+
+	ifname := "eno2" // Change this to an interface on your machine.
```
tohojo, line 29: Couldn't the interface name be given on the command line? That would make it possible to run the example without changing the code...

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
+// go:build ignore
+
+#include <linux/bpf.h>
+#include <bpf/bpf_helpers.h>
+#include <linux/if_ether.h>
+#include <linux/ip.h>
```
tohojo, line 6: We generally follow the Linux kernel coding style for the BPF code in this repo. I think we should do this here as well for consistency. So please change the indent to be a single tab character; and a few more comments below:

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+} protocol_count SEC(".maps");
+
+SEC("xdp")
+int get_packet_protocol(struct xdp_md *ctx) {
+
+  void *data_end = (void *)(long)ctx->data_end;
+  void *data = (void *)(long)ctx->data;
+
+  // Parse Ethernet header
+  struct ethhdr *eth = data;
```
tohojo, line 22: Variable declarations should all be at the top of the function.

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+  if ((void *)(eth + 1) > data_end) {
```
tohojo, line 23: No need to cast to void - this could just be `if (eth + 1 > data_end)`

Also, please drop the braces, they are not needed here.

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+    return XDP_PASS;
+  }
+
+  // Check if the packet is an IP packet
+  if (eth->h_proto != __constant_htons(ETH_P_IP)) {
```
tohojo, line 28: We use `bpf_htons()` in other places - that automatically selects the constant version as appropriate.

Also, drop the braces

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+    return XDP_PASS;
+  }
+
+  // Parse IP header
+  struct iphdr *ip = data + sizeof(struct ethhdr);
+  if ((void *)(ip + 1) > data_end) {
```
tohojo, line 34: As above - no need for the void cast or the braces.

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+    return XDP_PASS;
+  }
+
+  // Check if the packet is an IP packet
+  if (eth->h_proto != __constant_htons(ETH_P_IP)) {
+    return XDP_PASS;
+  }
+
+  // Parse IP header
+  struct iphdr *ip = data + sizeof(struct ethhdr);
```
tohojo, line 33: Variable declaration at the top of the file, please.

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+  if ((void *)(ip + 1) > data_end) {
+    return XDP_PASS;
+  }
+
+  __u32 key = ip->protocol; // Using IP protocol as the key
+  __u64 *count = bpf_map_lookup_elem(&protocol_count, &key);
```
tohojo, line 39: Variable declarations at the top of the file.

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,47 @@
...
+  if (count) {
```
tohojo, line 40: No need for the braces.

xdp-bpf2go-example/main.go
```diff
@@ -0,0 +1,232 @@
...
+
+	log.Printf("Analysing packets on %s..", ifname)
+
+	tick := time.Tick(time.Second)
+	stop := make(chan os.Signal, 5)
+	signal.Notify(stop, os.Interrupt)
+	for {
+		select {
+		case <-tick:
+			//    log.Print(objs.ProtocolCount)
```
tohojo, line 53: Leftover debug comment?

xdp-bpf2go-example/main.go
```diff
@@ -0,0 +1,232 @@
...
+	"os"
+	"os/signal"
+	"time"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/link"
+	"github.com/cilium/ebpf/rlimit"
+)
+
+func main() {
```
tohojo, line 16: Should the application check if it's being run as root? Otherwise users will just get a cryptic permission denied error when the memlock fails below...

tohojo (changes requested):
Generally looks good, thanks for the contribution! Only a few minor-ish comments on the code, see below.

Also, please squash the second and third commit into the first one when you respin :)

xdp-bpf2go-example/packetProtocol.c
David-VTUK: If I don't case this to a void pointer I get the following compilation error:

```
packetProtocol.c:27:17: warning: comparison of distinct pointer types lacks a cast
   27 |   if ((eth + 1) > data_end)
      |                 ^
```

I image this is because of:

```
 void *data = (void *)(long)ctx->data;
```

This is probably due to which compiler go invokes when running `go build` in this project.

xdp-bpf2go-example/packetProtocol.c
```diff
@@ -0,0 +1,44 @@
...
+
+SEC("xdp")
+int get_packet_protocol(struct xdp_md *ctx) {
+
+  void *data_end = (void *)(long)ctx->data_end;
+  void *data = (void *)(long)ctx->data;
+  struct ethhdr *eth = data;
+  struct iphdr *ip = data + sizeof(struct ethhdr);
+  __u32 key = ip->protocol; // Using IP protocol as the key
+  __u64 *count = bpf_map_lookup_elem(&protocol_count, &key);
```
tohojo, line 24: I don't think this works? You're dereferencing the `ip` variable before the range check below, which will be rejected by the verifier.

You can declare the variable without assigning to it at the top of the function (just `__u32 key`), and then to the assignment `key = ip->protocol` below the range check at the bottom. Same thing for the assignment to `count` (and arguably to `ip`.

Also, please fix the indentation (to be a single tab character per indentation level) as is the custom in the kernel sources.

David-VTUK: It certainly works but I agree with you, I'll change accordingly.

Apologies I thought I already changed the indentation but my ide formatted on save.

PR #127, opened 2025-03-06
configure: Add libcap check

chentao-kernel:
The xdpsock needs libcap, otherwise it will compile failed. So we can add the libcap check in configure.

xdpsock.c:27:10: fatal error: sys/capability.h: No such file or directory
   27 | #include <sys/capability.h>
      |          ^~~~~~~~~~~~~~~~~~

The test result:

dylane@2404:~/sdb/bpf-examples$ ./configure
clang: 18.1.3 (1ubuntu1)
libcap support: yes
libmnl support: yes
libbpf support: submodule
ELF support: yes
zlib support: yes
libxdp support: submodule

### DIFF
--- a/configure
+++ b/configure
@@ -84,6 +84,20 @@ check_zlib()
     fi
 }
 
+check_libcap()
+{
+    if ${PKG_CONFIG} libcap --exists; then
+        echo "HAVE_LIBCAP:=y" >>$CONFIG
+        echo "yes"
+
+        echo 'CFLAGS += -DHAVE_LIBCAP' `${PKG_CONFIG} libcap --cflags` >> $CONFIG
+        echo 'LDLIBS += ' `${PKG_CONFIG} libcap --libs` >>$CONFIG
+    else
+        echo "missing - this is required"
+        return 1
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -84,6 +84,20 @@ check_zlib()
     fi
 }
 
+check_libcap()
+{
+    if ${PKG_CONFIG} libcap --exists; then
+        echo "HAVE_LIBMNL:=y" >>$CONFIG
```
tohojo, line 90: copy-paste error?

configure
```diff
@@ -84,6 +84,20 @@ check_zlib()
...
+        echo "yes"
+
+        echo 'CFLAGS += -DHAVE_LIBMNL' `${PKG_CONFIG} libcap --cflags` >> $CONFIG
```
tohojo, line 93: ...and here...

tohojo (changes requested):

configure
chentao-kernel: will fix it

chentao-kernel: > copy-paste error?

done, thanks

PR #130, opened 2025-06-24
trie-free-latency: Demo showing the latency of freeing LPM_TRIE maps

mfleming:
The performance of freeing LPM_TRIE maps suffers heavily for maps with a large number of entries. Here's a little demo that measures the duration of trie_free() and allows the user to control how many entries are in the map, as well as how many iterations of the demo are run to collect latency statistics.

This is a reproduction of an issue we've seen at Cloudflare and a public demo to aid with the upstream discussion at https://lore.kernel.org/lkml/20250616095532.47020-1-matt@readmodwrite.com/

### DIFF
--- a/LPM-map-bench/.gitignore
+++ b/LPM-map-bench/.gitignore
@@ -1,0 +1,1 @@
+trie-free-latency
--- a/LPM-map-bench/Makefile
+++ b/LPM-map-bench/Makefile
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
+
+USER_TARGETS   := trie-free-latency
+BPF_TARGETS    := trie-free-latency.bpf stats.bpf
+BPF_SKEL_OBJ   := trie-free-latency.bpf.o stats.bpf.o
+
+EXTRA_DEPS += trie.h
+LDLIBS     += -lm
+
+LIB_DIR = ../lib
+
+include $(LIB_DIR)/common.mk
+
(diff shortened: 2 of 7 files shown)


netoptimizer:
Thanks for working on this "homework" assignment ;-)

trie-free-latency/stats.bpf.c
```diff
@@ -0,0 +1,64 @@
...
+{
+	__u64 *tsp;
+
+	tsp = bpf_map_lookup_elem(&latency_start, &map);
+	if (tsp) {
+		__u64 delta_ns = bpf_ktime_get_ns() - *tsp;
+		struct latency_record r = {
+			.val = delta_ns,
+		};
+		char *name = BPF_CORE_READ(map, name);
```
netoptimizer, line 48: On KASAN kernels the map name will not be available at `fexit`

- Thus, this code will not work on KASAN

mfleming: Thanks! I've updated the code to record the map when it's available (fentry of `trie_free`) and stash in the `latency_start` map.

netoptimizer:
I'm considering if we should rename the example director `trie-free-latency` 

- to something with `LPM` in the name, to make the connection with the `BPF_MAP_TYPE_LPM_TRIE`

What about these proposals:

- `LPM-map-tests` or `LPM-map-bench`

I'm considering adding another example in this directory that can monitor the runtime performance/latency of LPM map lookups

- I already have a `bpftrace` script, that measures lookup latency, but it added too much overhead (to production)
- Thus, I want to write a C/BPF version with minimal overhead

PR #133, opened 2025-09-29
netstacklat - Handle TCP head-of-line blocking

simosund:
Currently netstacklat may report latency for head-of-line (HOL) blocked TCP reads. TCP HOL-blocking can occur when there is packet loss or reordering, when the TCP stack has to delay the delivery of a packet to the socket receive queue until the preceding TCP segments have been received.

As this HOL-blocking is dependent on factors external to the local host (i.e. the packet loss and latency of the entire network path from the sender), this is not the type of latency we typically want to highlight with netstacklat. This PR addresses this issue by not recording the latency for reads that may experience HOL-blocking.

See the commit messages for further details on the how this is accomplished and some limitations.

### DIFF
--- a/netstacklat/netstacklat.bpf.c
+++ b/netstacklat/netstacklat.bpf.c
@@ -188,11 +188,17 @@ static bool filter_network_ns(struct sk_buff *skb, struct sock *sk)
 	return get_network_ns(skb, sk) == user_config.network_ns;
 }
 
+static bool filter_network(struct sk_buff *skb, struct sock *sk)
+{
+	if (!filter_ifindex(skb ? skb->skb_iif : sk ? sk->sk_rx_dst_ifindex : 0))
+		return false;
+
+	return filter_network_ns(skb, sk);
+}
 
 static void record_skb_latency(struct sk_buff *skb, struct sock *sk, enum netstacklat_hook hook)
 {
 	struct hist_key key = { .hook = hook };
-	u32 ifindex;
 
 	if (bpf_core_field_exists(skb->tstamp_type)) {
 		/*
@@ -217,15 +223,11 @@ static void record_skb_latency(struct sk_buff *skb, struct sock *sk, enum netsta
 			return;
 	}
 
-	ifindex = skb->skb_iif;
-	if (!filter_ifindex(ifindex))
-		return;
-
-	if (!filter_network_ns(skb, sk))
+	if (!filter_network(skb, sk))
 		return;
 
 	if (user_config.groupby_ifindex)
-		key.ifindex = ifindex;
+		key.ifindex = skb->skb_iif;
 
 	record_latency_since(skb->tstamp, &key);
 }
@@ -305,7 +307,6 @@ static void record_socket_latency(struct sock *sk, struct sk_buff *skb,
 {
 	struct hist_key key = { .hook = hook };
 	u64 cgroup = 0;
-	u32 ifindex;
 
 	if (!filter_min_sockqueue_len(sk))
 		return;
@@ -316,15 +317,11 @@ static void record_socket_latency(struct sock *sk, struct sk_buff *skb,
 	if (!filter_current_task(cgroup))
 		return;
 
-	ifindex = skb ? skb->skb_iif : sk->sk_rx_dst_ifindex;
-	if (!filter_ifindex(ifindex))
-		return;
-
-	if (!filter_network_ns(skb, sk))
+	if (!filter_network(skb, sk))
 		return;
 
 	if (user_config.groupby_ifindex)
-		key.ifindex = ifindex;
+		key.ifindex = skb ? skb->skb_iif : sk->sk_rx_dst_ifindex;
 	if (user_config.groupby_cgroup)
 		key.cgroup = cgroup;
 
--- a/netstacklat/netstacklat.bpf.c
+++ b/netstacklat/netstacklat.bpf.c
@@ -11,6 +11,10 @@
 
 #define READ_ONCE(x) (*(volatile typeof(x) *)&(x))
 
+// Mimic macros from /include/net/tcp.h
+#define tcp_sk(ptr) container_of(ptr, struct tcp_sock, inet_conn.icsk_inet.sk)
+#define TCP_SKB_CB(__skb)	((struct tcp_skb_cb *)&((__skb)->cb[0]))
+
 char LICENSE[] SEC("license") = "GPL";
 
 
@@ -23,6 +27,7 @@ volatile const struct netstacklat_bpf_config user_config = {
 	.filter_cgroup = false,
 	.groupby_ifindex = false,
 	.groupby_cgroup = false,
+	.include_hol_blocked = false,
 };
 
 /*
@@ -37,6 +42,13 @@ struct sk_buff___old {
 	};
 	__u8 mono_delivery_time: 1;
 } __attribute__((preserve_access_index));
+
+struct tcp_sock_ooo_range {
+	u32 prev_n_ooopkts;
+	u32 ooo_seq_end;
+	/* indicates if ooo_seq_end is still valid (as 0 can be valid seq) */
+	bool active;
+};
 
 struct {
 	__uint(type, BPF_MAP_TYPE_PERCPU_HASH);
@@ -65,6 +77,22 @@ struct {
 	__type(key, u64);
 	__type(value, u64);
 } netstack_cgroupfilter SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_SK_STORAGE);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+	__type(key, int);
+	__type(value, struct tcp_sock_ooo_range);
+} netstack_tcp_ooo_range SEC(".maps");
+
+/*
+ * Is a < b considering u32 wrap around?
+ * Based on the before() function in /include/net/tcp.h
+ */
+static bool u32_lt(u32 a, u32 b)
+{
+	return (s32)(a - b) < 0;
+}
 
 static u64 *lookup_or_zeroinit_histentry(void *map, const struct hist_key *key)
 {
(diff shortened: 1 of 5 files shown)


simosund:
@netoptimizer In my last commit I've added some sanity checks like we talked about on the meeting. I am however not sure if these sanity checks are a very good idea, as they do come with a noticeable increase in overhead and still offer somewhat limited protection.

### Performance impact
I've rerun my test with various degrees of packet loss, and the results are available in [this dashboard](https://snapshots.raintank.io/dashboard/snapshot/zKw4xyfSLQSPRGLppWRJ0xnW6UdkCbHI). The first run is with current master (no ooo-filter at all), the second with "alt3" as of commit dff6bc54988ea (without sanity checks, same as alt3 in my previous test), and the last is alt3 with the sanity checks added (in commit 59e923df588d6).

Looking at performance for the tcp-socket-read hook (the only one that should be affected), it looks like this:

#### Program runtime
retrans_checksoverhead_runtime

*NOTE:* The spikes to ~700ns here are from periods of very little traffic, so focus on the parts between the spikes.

Runtime without ooo-filter varies from around 250 - 290ns, runtime with ooo-filter varies from around 300 - 390ns, and runtime with ooo-filter + added sanity checks varies from around 350 - 470ns. So the sanity checks add about as much additional overhead as the entire filtering mechanism itself. Or phrased a differently, the sanity checks roughly double the overhead of the entire filtering mechanism as a whole.

#### Overall CPU util
retrans_checksoverhead_cpu

The largest CPU utilization overall seems to occur at the test with 0.1% packet loss (second spike in each group). For this test the CPU utilization without ooo-filtering is about 0.34%, about 0.46% with ooo-filtering and around 0.57% with ooo-filtering + sanity checks. So once again, the sanity checks roughly double the overhead of the ooo-filtering.

### Limitations with the sanity checks
The sanity checks are described in the commit message (and the code). There's basically two types of them.
- A sanity check that the sequence number (`skb->cb.end_seq`) read from `tp->ooo_last_skb` is within the current receive window
- Sanity checks that `tp->rcv_nxt` and `tp->copied_seq` did not decrease since last read

The sanity check that the ooo sequence number is within the receive window adds some overhead as it requires probing the tp->rcv_wup and tp->rcv_wnd fields as well. It also adds a potential new error source as now we need to trust that they are right, although it's less likely they're catastrophically wrong than the skb read (as they don't risk being read from a freed skb). 

I'm also not entirely sure that the end of the ooo-sequence number must always be within the current receive window, even if that should usually be the case. Partly because depending on system configuration it appears that the receive window can shrink (see the [tcp_select_window()](https://elixir.bootlin.com/linux/v6.17.4/source/net/ipv4/tcp_output.c#L279)), in which the current receive window might actually be smaller than the receive window was when the skb was added to the ooo-queue. And partly because as far as I can tell from the code in [tcp_data_queue()](https://elixir.bootlin.com/linux/v6.17.4/source/net/ipv4/tcp_input.c#L5155) it only drops packets where the starting sequence (cb.seq) is ahead of the receive window, meaning that segments where the end sequence (cb.end_seq) is ahead of the receive window may still be added to the error queue.

I'm also not sure how much the "doesn't decrease between reads" checks on rcv_nxt and copied_seq really help, as even if we get a complete garbage value (which we generally shouldn't here though, as the socket should still be valid while our eBPF program runs and 32-bit writes I assume are "atomic" on most architectures) there's only about a 50% chance we will detect the error. Or put another way, it won't detect errors below ~2GB. It might be possible to reel in these limits further if there's some good way to determine the maximum amount of data that can be data that can be read in a single TCP socket read, but even if possible that will likely require probe reading additional socket members (adding additional overhead and error sources).

### Conclusion
The sanity check on the data in the ooo_last_skb would be really good to have (as that is read from a potentially invalid pointer), but it does add some overhead (and new potential error sources) due to having to probe additional fields, and I'm not sure the concept behind it (comparing it against the current receive window) is entirely sound.

The sanity checks on `rcv_nxt` and `coped_seq` are quite weak. They also add some overhead as now they always have to be probed to keep track of their state, while before they only needed to be probed conditionally when needed.

simosund:
Hi again @netoptimizer,
I've now split up these sanity checks into two different commits. The first one adds the sanity check for `tp->ooo_last_skb->cb.end_seq` value (that it's non-zero and within the current receive window), while the second commit adds the monotonic growth checks for `tp->rcv_nxt` and `tp->copied_seq`. TLDR, I recommend including the first these commits, and skipping the second.

I think it makes sense to adapt the first of these sanity checks, as that's the only one were we have to follow an additional pointer to a potentially no longer valid SKB. As I mentioned previously, I think there might be some circumstances where `end_seq` of the ooo-queue can actually end up ahead of the current receive window, so this check might be a bit too strict. In some cases the receive window can shrink, but only if `net.ipv4.tcp_shrink_window` is set (1). However, the default is for it to be unset (0) and I doubt it's particularly common that people change it (although please check its value on your systems in case it's more common than I think). As I also noted previously, as far as I can tell SKBs where the `end_seq` but not `seq` (i.e. the start seq) is ahead of receive window may be added to the ooo-queue. However, after also going through the code in `tcp_rcv_established()` (which is the one that calls `tcp_data_queue()`, which in turns calls `tcp_data_queue_ofo()`) I saw there were some additional checks that ultimately culminates in `tcp_sequence()`, where an SKB with an `end_seq` that overruns the current receive window is only accepted if the receive queue is currently empty. So requires very specific conditions for that to go wrong (ooo-segment where start-seq is within the receive window but end-seq is beyond it arrives while the receive queue is empty). 

I'm more skeptical to the sanity checks on `rcv_nxt` and `copied_seq`. As I've previously mentioned they're rather weak checks, and require that we track and update their state every time. They're also for values that are direct members of the tcp_sock, so the values we get for them should still be valid (although possibly updated before or in-between we read them), as long as only a part of the value is updated at a time (e.g. the first/last 16 bits). But 32-bit writes should effectively be "atomic" (as it either you'll read the full value before an update, or the full value after an update) on most architectures, right?

## Performance overhead
Through the test below I also show that the the `tp->ooo_last_skb->cb.end_seq` sanity check seems to have neglible additional overhead compared to alt3 ooo-filtering without any sanity checks. While also adding in the sanity checks for `rcv_nxt` and `copied_seq` roughly doubles the overhead of the ooo-filtering. 

I've done an additional set of tests were I run:
- no ooo-filtering (current master)
- alt3 ooo-filtering without any sanity checks ()
- alt3 ooo-filtering with `tp->ooo_last_skb->cb.end_seq` sanity check ()
- alt3 ooo-filtering with both `tp->ooo_last_skb->cb.end_seq` + `rcv_nxt` and `copied_seq` sanity checks

As usual, full dashboard with interactive (mouseover) plots are available [here](https://snapshots.raintank.io/dashboard/snapshot/YuyP8OiUotmywX3CyxG0TX8mV8XNYCCS). For convenience, I've inserted the performance-related figures below:

Screenshot 2025-10-27 at 18-42-03 View panel - netstacklat - TCP HoL blocking - none alt3 alt3 ooo-seq-check alt3 ooo-seq-check rcv_nxt_copied-seq-check - Grafana

Screenshot 2025-10-27 at 18-43-27 View panel - netstacklat - TCP HoL blocking - none alt3 alt3 ooo-seq-check alt3 ooo-seq-check rcv_nxt_copied-seq-check - Grafana

simosund:
Right, forgot that this hadn't been merged yet. I don't think anyone has formally reviewed it, but I assume @netoptimizer in practice ended up reviewing it while applying it to the Cloudflare build?

As I mention in the discussion above, I also think we should skip the last commit, as those sanity checks add considerable overhead while it's doubtful they would really catch many real issues. So unless anyone objects about that, I should drop that before we merge.

tohojo:
I'm OK with dropping the last commit if it doesn't matter in practice. Ping @netoptimizer re: the review status? :)

tohojo:
Any updates on this, @simosund ? As I recall, you were planning to drop the last commit from the series?

simosund:
Yes, I was initially waiting on @netoptimizer to go through it, but I'm fairly confident that he never applied the last commit either, and in my own evaluation I did not use the last commit. So let's just drop it.

From my understanding the plan was for @netoptimizer to review it, but as he's currently on vacation I guess the status there is unchanged.

netstacklat/netstacklat.bpf.c
```diff
@@ -300,12 +330,156 @@ static bool filter_min_sockqueue_len(struct sock *sk)
...
+		bpf_printk("failed to read tcp_sock->rcv_wnd, err=%d", err);
+		goto exit;
+	}
+
+	window = rcv_wup + rcv_wnd;
+	if (u32_lt(window, rcv_nxt))
+		window = rcv_nxt;
+
+exit:
+	*seq = window;
```
tohojo, line 359: nit: the usual pattern is to only write to the return variable if there is no error.

netstacklat/netstacklat.bpf.c
```diff
@@ -300,12 +330,156 @@ static bool filter_min_sockqueue_len(struct sock *sk)
...
+		 * be the result of reading from an invalid SKB pointer
+		 */
+		if (cb.end_seq == 0 || u32_lt(cur_rcv_window, cb.end_seq))
+			max_seq = cur_rcv_window;
+		else
+			max_seq = cb.end_seq;
+	}
+
+exit:
+	*seq = max_seq;
```
tohojo: Same here

tohojo (changes requested):
A few nits, otherwise I think we should just merge this :)

netstacklat/netstacklat.bpf.c
simosund: Fixed. As there's nothing that needs to be done on errors now, I've replaced the goto exit with early returns.

netstacklat/netstacklat.bpf.c
simosund: Fixed. As there's nothing that needs to be done on errors now, I've replaced the goto exit with early returns. Furthermore, seq is now set directly instead of using the intermediate max_seq variable.

tohojo (approved):
Alright, let's merge this then :)

PR #142, opened 2026-06-24
pping: add json-lines (jsonl) output format

tobiaspal:
Adds `--format jsonl`: one JSON object per line with no enclosing array, so the output is an append-only log that can be tailed/streamed and stays valid as it grows. The existing `json` format wraps everything in a single root array that is only parseable once the process closes it, and concatenates into invalid JSON if the output is appended across restarts. Same per-event objects and fields as `json`; the `json` format itself is unchanged.

- Commit 1 adds the format — a `line_delimited` mode on the json_writer — plus README and `--format` help.
- Commit 2 line-buffers the stream so each event is flushed as it is written.

Tested against live TCP traffic (IPv4 and IPv6); each emitted line validates as standalone JSON with `jq`.

### DIFF
--- a/lib/util/json_writer.c
+++ b/lib/util/json_writer.c
@@ -25,6 +25,7 @@ struct json_writer {
 	FILE		*out;	/* output file */
 	unsigned	depth;  /* nesting */
 	bool		pretty; /* optional whitepace */
+	bool		line_delimited; /* json-lines mode */
 	char		sep;	/* either nul or comma */
 };
 
@@ -100,6 +101,7 @@ json_writer_t *jsonw_new(FILE *f)
 		self->out = f;
 		self->depth = 0;
 		self->pretty = false;
+		self->line_delimited = false;
 		self->sep = '\0';
 	}
 	return self;
@@ -120,6 +122,15 @@ void jsonw_destroy(json_writer_t **self_p)
 void jsonw_pretty(json_writer_t *self, bool on)
 {
 	self->pretty = on;
+	if (on)
+		self->line_delimited = false;
+}
+
+void jsonw_line_delimited(json_writer_t *self, bool on)
+{
+	self->line_delimited = on;
+	if (on)
+		self->pretty = false;
 }
 
 /* Basic blocks */
@@ -139,7 +150,14 @@ static void jsonw_end(json_writer_t *self, int c)
 	if (self->sep != '\0')
 		jsonw_eol(self);
 	putc(c, self->out);
-	self->sep = ',';
+	if (self->line_delimited && self->depth == 0) {
+		/* Terminate each top-level object with a trailing newline */
+		putc('\n', self->out);
+		fflush(self->out);
+		self->sep = '\0';
+	} else {
+		self->sep = ',';
+	}
 }
 
 
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -73,6 +73,7 @@ static const char *__doc__ =
 enum pping_output_format {
 	PPING_OUTPUT_STANDARD,
 	PPING_OUTPUT_JSON,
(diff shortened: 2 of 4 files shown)


lib/util/json_writer.c
```diff
@@ -122,6 +124,11 @@ void jsonw_pretty(json_writer_t *self, bool on)
 	self->pretty = on;
 }
 
+void jsonw_line_delimited(json_writer_t *self, bool on)
+{
+	self->line_delimited = on;
```
tohojo, line 131: This should probably turn `pretty` off (and vice versa); it does not appear those two will work together?

lib/util/json_writer.c
```diff
@@ -139,7 +146,13 @@ static void jsonw_end(json_writer_t *self, int c)
 	if (self->sep != '\0')
 		jsonw_eol(self);
 	putc(c, self->out);
-	self->sep = ',';
+	if (self->line_delimited && self->depth == 0) {
+		/* End each top-level object with its own newline so json-lines consumers see a complete line as soon as it is written, not when the next object starts */
```
tohojo: Long comment, please line-break.

lib/util/json_writer.c
```diff
@@ -139,7 +146,13 @@ static void jsonw_end(json_writer_t *self, int c)
...
+		putc('\n', self->out);
```
tohojo, line 155: Let's just add an explicit `fflush()` here instead of changing the line buffering at the caller.

pping/pping.c
```diff
@@ -2106,11 +2112,18 @@ static struct output_context *open_output(const char *filename,
...
-	if (out_ctx->format == PPING_OUTPUT_JSON) {
+	if (out_ctx->format == PPING_OUTPUT_JSON ||
+	    out_ctx->format == PPING_OUTPUT_JSONL) {
 		out_ctx->jctx = jsonw_new(out_ctx->stream);
 		if (!out_ctx->jctx)
 			goto err;
-		jsonw_start_array(out_ctx->jctx);
+		if (out_ctx->format == PPING_OUTPUT_JSONL) {
+			/* Line-buffer so each event reaches the file as it is written, not in block-sized chunks */
+			setvbuf(out_ctx->stream, NULL, _IOLBF, BUFSIZ);
```
tohojo: See above wrt `fflush()` instead of this.

tohojo (changes requested):

pping/pping.c
```diff
@@ -73,7 +73,8 @@ static const char *__doc__ =
 enum pping_output_format {
 	PPING_OUTPUT_STANDARD,
 	PPING_OUTPUT_JSON,
-	PPING_OUTPUT_PPVIZ
+	PPING_OUTPUT_PPVIZ,
+	PPING_OUTPUT_JSONL
```
tohojo: This is not UAPI, we can add stuff in the middle. So let's just have JSON and JSONL next to each other like everywhere else.

tohojo (changes requested):
The enum thing below is really a nit, but there's no reason to have the `fflush()` be it's own commit at this point, so please squash into a single commit and fix the enum order while you're at it. Otherwise LGTM :)

pping/pping.c
tobiaspal: Done — moved `PPING_OUTPUT_JSONL` between `JSON` and `PPVIZ`, and reordered the matching `parse_arguments` branch to keep them adjacent.

tobiaspal:
Squashed to a single commit (the `fflush()` folded into the format commit) and force-pushed.
Appreciate the nitpicks too, let's make this as good as possible.

tohojo (approved):

PR #143, opened 2026-06-30
pping: support PPP/PPPoE and other L3 interfaces (no Ethernet header)

tobiaspal:
pping is exactly the tool I wanted, but I couldn't use it: my ISP uplink is PPPoE, which carries no Ethernet header, so pping produced no RTTs on it. This change makes it work on L3 interfaces like that.

The BPF parser assumed an Ethernet header on every packet, so on `ARPHRD_PPP`/`ARPHRD_NONE` interfaces (PPP/PPPoE uplinks, tun, and similar) it consumed 14 bytes of the IP header and dropped every packet as non-IP, so no RTT was produced.

- `parse_l3_proto()` reads the IP version nibble to pick IPv4/IPv6 without advancing the cursor, so the existing `parse_iphdr`/`parse_ip6hdr` run unchanged.
- L3 interfaces are auto-detected via `SIOCGIFHWADDR` (`ARPHRD_PPP`/`NONE`); `--eth-header`/`--ip-header` override the detection.

Tested on a live 3-uplink PPPoE router: all three links autodetect as L3 and produce TCP RTTs in jsonl.

A few things I'd appreciate your take on:

1. **Are the override flags worth keeping?** With `SIOCGIFHWADDR` autodetection they are only an escape hatch -- `--ip-header` could force L3 for a link type we don't recognise, but `--eth-header` is fairly theoretical (nothing realistic forces an Ethernet header over an `ARPHRD_PPP`/`NONE` detection). Happy to drop one or both and just trust the autodetect.
2. **`--eth-header`/`--ip-header` are long-only** (option codes >= 128, kept under 256 so `print_usage()`'s `isalnum()` stays valid). I did this so the flags are self-contained: dropping them touches only the enum, two `long_options` rows, and two `switch` cases -- nothing in the getopt short-string -- which makes their cost easy to see. If we keep them I can add short options.
3. If the overrides stay, I'll follow up to make `--eth-header`/`--ip-header` mutually exclusive (today both can be passed and the last wins) and to have an explicit flag take precedence over the autodetect regardless of argument order.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -6,6 +6,10 @@ static const char *__doc__ =
 #include <bpf/libbpf.h>
 #include <linux/if_link.h>
 #include <net/if.h> // For if_nametoindex
+#include <net/if_arp.h> // For ARPHRD_PPP/ARPHRD_NONE (detecting L3 interfaces)
+#include <sys/ioctl.h> // For ioctl()
+#include <linux/sockios.h> // For SIOCGIFHWADDR
+#include <sys/socket.h> // For socket()
 #include <arpa/inet.h> // For inet_ntoa and ntohs
 
 #include <stdio.h>
@@ -273,6 +277,35 @@ static int parse_bounded_long(long long *res, const char *str, long long low,
 	return 0;
 }
 
+/*
+ * Interfaces such as PPP/PPPoE and tun carry no Ethernet header: at the
+ * tc/XDP hook the packet starts at the IP header. Detect them via their ARP
+ * hardware type so the BPF programs skip Ethernet parsing.
+ */
+static bool iface_is_l3(const char *ifname)
+{
+	struct ifreq ifr = { 0 };
+	int fd, ret;
+
+	fd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (fd < 0)
+		return false;
+
+	memcpy(ifr.ifr_name, ifname, strlen(ifname) + 1);
+	ret = ioctl(fd, SIOCGIFHWADDR, &ifr);
+	close(fd);
+	if (ret < 0)
+		return false;
+
+	switch (ifr.ifr_hwaddr.sa_family) {
+	case ARPHRD_PPP:
+	case ARPHRD_NONE:
+		return true;
+	default:
+		return false;
+	}
+}
+
 static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 {
 	int err, opt, len;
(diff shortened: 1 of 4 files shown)


pping/pping.c
```diff
@@ -273,6 +285,35 @@ static int parse_bounded_long(long long *res, const char *str, long long low,
...
+static bool iface_is_l3(const char *ifname)
+{
+	struct ifreq ifr = { 0 };
+	int fd, ret;
+
+	fd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (fd < 0)
+		return false;
+
+	strncpy(ifr.ifr_name, ifname, IFNAMSIZ - 1);
```
tohojo: This gives the following compiler warning on my system:
```
In function ‘iface_is_l3’,
    inlined from ‘parse_arguments’ at pping.c:357:8,
    inlined from ‘main’ at pping.c:2662:8:
pping.c:302:9: warning: ‘strncpy’ output may be truncated copying 15 bytes from a string of length 15 [-Wstringop-truncation]
  302 |         strncpy(ifr.ifr_name, ifname, IFNAMSIZ - 1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

We already checked that the source is null-terminated, so maybe just turn this into a memcpy()?

pping/pping.c
```diff
@@ -311,6 +353,13 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 					get_libbpf_strerror(err));
 				return err;
 			}
+
+			if (iface_is_l3(config->ifname)) {
+				config->bpf_config.eth_header = false;
+				fprintf(stderr,
+					"%s is an L3 interface (no Ethernet header); parsing from the IP header\n",
+					config->ifname);
```
tohojo: Instead of this separate print, how about we incorporate it into the existing "Starting ePPing" line (adding the mode after the interface name)? Something like:

```
Starting ePPing in standard mode tracking TCP on wg0 (IP header mode)
```

or 

```
Starting ePPing in standard mode tracking TCP on eth0 (Ethernet header mode)
```

tohojo (changes requested):
Very nice! I tested this on a Wireguard interface, and that works as expected.

A few nits below. And I'd say skip the overrides for now; if it turns out to be problematic we can always add them back later, but I don't see how it should, really...

pping/pping.c
tobiaspal: Oh yes, much better

pping/pping.c
tobiaspal: Done. `memcpy(ifr.ifr_name, ifname, strlen(ifname) + 1)` now; the name is length-checked and null-terminated at `-i`.

pping/pping.c
tobiaspal: Done, folded into the startup line: `... tracking TCP on pppoe-wan_3bb (IP header mode)`.

tobiaspal:
v2, force-pushed: dropped `--eth-header`/`--ip-header` (autodetect only), switched `strncpy` to `memcpy`, and folded the header mode into the startup line.

tohojo (approved):
Looks good, thanks!

PR #144, opened 2026-07-01
pping: don't re-create the shared clsact qdisc for the egress hook

tobiaspal:
On startup `pping` prints:

```
libbpf: Kernel error message: Exclusivity flag on, cannot modify
```

This is confusing as a user: is it bad? is something not working? did I do something wrong? do I need to act on it, and how? I spent a while tracking it down. It turns out to be harmless, but that isn't obvious to a newcomer, so I tried to fix it.

It's also mentioned in passing in [#86](https://github.com/xdp-project/bpf-examples/issues/86#issuecomment-1549936336), where @simosund lists this message among the "harmless error messages" that are "maybe not ideal".

In tc mode the ingress and egress programs share one clsact qdisc, but each `tc_attach()` issued its own `bpf_tc_hook_create()`. The second (egress) create hits the existing qdisc, which the kernel rejects with `-EEXIST` and libbpf logs as `Exclusivity flag on, cannot modify` on every startup. The `-EEXIST` is already tolerated, but the redundant create still runs and produces the log.

`tc_attach()` now takes a `create_hook` argument, so only the first attach for an interface creates the qdisc (ingress in tc mode, or egress when ingress used XDP). The other attach uses the existing hook.

Tested on a live PPPoE router in tc-ingress mode: the `Exclusivity flag on` line is gone from startup, RTT collection unaffected.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -601,8 +601,12 @@ static int xdp_detach(struct xdp_program *prog, int ifindex,
 }
 
 /*
- * Will attempt to attach program at section sec in obj to ifindex at
- * attach_point.
+ * Will attach program prog_name in obj to ifindex at attach_point.
+ * The ingress and egress programs share a single clsact qdisc (it carries both
+ * hook points), so only the first attach for an interface should create it:
+ * pass create_hook=true for that one and false for the other. This avoids
+ * re-issuing a create for the already-existing qdisc, which the kernel rejects
+ * with -EEXIST and libbpf logs as "Exclusivity flag on, cannot modify".
  * On success, will fill in the passed opts, optionally set new_hook depending
  * if it created a new hook or not, and return the id of the attached program.
  * On failure it will return a negative error code.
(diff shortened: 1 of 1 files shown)


tohojo:
So this helps the first time you run pping on the interface, but not on
subsequent runs. If we're fixing this I think we should make sure we solve this
case as well. Options:

- Just silence the libbpf error instead of trying to avoid it (using a log
  handler).

- Relax the '--force' requirement for removing the clsact qdisc on shutdown (so
  always remove it if we created it).

- Move to TCX attachment.


Probably the second option is the easiest...

tobiaspal:
Yes, good point, I started using `--force` early on because I wanted the lingering qdiscs cleaned up.

Went with option 2 (added as a second commit). The shutdown teardown now removes the shared clsact qdisc whenever pping created it (`created_tc_hook`) instead of only under `--force`, so it no longer lingers between runs. This allows me to stop using the `--force` option.

But attaching onto an existing qdisc still trips the ingress `-EEXIST`, and libbpf still logs it there — these patches remove the message for pping's own qdisc, not a foreign one (silencing that would be the log-handler route).

One thing this surfaces: after the change `--force` looks unused — `config.force` is only set, never read. It also doesn't seem to have driven the "overwrite existing XDP program" behaviour from its help text: `xdp_program__attach()` is called with flags `0`, not `config.force`. So it may now be a no-op — happy to drop the flag or wire it into the XDP attach, whichever you'd prefer; I didn't want to fold that into this PR without checking.

Edit: Option 3 (Move to TCX attachment) is implemented in Pull Request #145.

tohojo:
Ah right, the force behaviour was from before pping started using libxdp to load XDP programs. With libxdp it's not really needed because that supports loading multiple programs. So I think it's fine to drop it...

tobiaspal:
Deprecated it in a third commit: `--force`/`-f` is still accepted but only prints a warning, and the `config.force` field is gone. A hard removal would make existing invocations fail with "unrecognized option", so the flag stays a no-op for now and can be removed in a later major version.

tohojo (approved):

PR #145, opened 2026-07-02
pping: attach the tc programs with TCX links on supported kernels

tobiaspal:
This implements the "Move to TCX attachment" option from the #144 discussion.

pping now attaches its tc programs through TCX bpf links (kernel 6.6+) instead of the netlink clsact hook. That removes the shared-qdisc bookkeeping on current kernels: no qdisc is created on the interface, attachments do not linger after a crash (the kernel detaches the programs when the link fds close, even on SIGKILL), and the -EEXIST/"Exclusivity flag on" issue does not arise. Kernels without TCX support reject the link creation with -EINVAL — I verified this against the v6.1 and v6.6 kernel sources (`link_create()` had no SCHED_CLS case before 6.6, and the `CONFIG_NET_XGRESS=n` stub also returns -EINVAL) — and pping then falls back on the legacy clsact path, which is unchanged (including the #144 fixes).

Two design choices worth calling out:

- I used the low-level `bpf_link_create()` instead of `bpf_program__attach_tcx()`. It takes the attach type explicitly, so `pping_kern.c` keeps its plain `SEC("tc")` sections (the object stays attachable through both mechanisms, and through iproute2), and the fallback probe on pre-6.6 kernels does not trigger a libbpf `pr_warn()` — avoiding the kind of log noise #144 removes.
- `headers/linux/bpf.h` was still the v5.11-rc7 snapshot, which predates the TCX attach types, so the second commit updates it from kernel v6.6 following the convention in headers/README.md (including the `__LINUX_BPF_H__` include-guard rename that `make headers_install` applies). To check the shared header update against the rest of the repository, I built all SUBDIRS examples with it (Ubuntu 24.04, gcc 13, clang 18) — all compile.

Tested on the same 3-uplink PPPoE router as the previous PRs (kernel 7.0): `bpftool net show` lists a tcx/ingress + tcx/egress pair per interface and `tc qdisc show` stays empty; RTTs report as before; SIGKILL detaches both programs and the restart re-attaches them; a clean stop leaves no attachments behind. What I could not test live is the legacy fallback on a real pre-6.6 kernel — it does not run on my systems, though the legacy attach code itself is the path previous versions used.

The first commit is a small preparatory cleanup: `config->ingress_prog_id`/`egress_prog_id` were only written, and the TCX commit would otherwise have had to store link fds into misnamed fields.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -153,8 +153,6 @@ struct pping_config {
 	char *event_map;
 	int ifindex;
 	struct xdp_program *xdp_prog;
-	int ingress_prog_id;
-	int egress_prog_id;
 	char ifname[IF_NAMESIZE];
 	char filename[PATH_MAX];
 	enum pping_output_format format;
@@ -2027,7 +2025,6 @@ static int load_attach_bpfprogs(struct bpf_object **obj,
 		err = tc_attach(*obj, config->ifindex, BPF_TC_INGRESS,
 				config->ingress_prog, &config->tc_ingress_opts,
 				true, &config->created_tc_hook);
-		config->ingress_prog_id = err;
 	}
 	if (err < 0) {
 		fprintf(stderr,
@@ -2037,17 +2034,15 @@ static int load_attach_bpfprogs(struct bpf_object **obj,
 	}
 
 	// Attach egress prog; create the clsact only when ingress used XDP
-	config->egress_prog_id = tc_attach(
+	err = tc_attach(
 		*obj, config->ifindex, BPF_TC_EGRESS, config->egress_prog,
 		&config->tc_egress_opts,
 		strcmp(config->ingress_prog, PROG_INGRESS_XDP) == 0,
 		config->created_tc_hook ? NULL : &config->created_tc_hook);
-	if (config->egress_prog_id < 0) {
+	if (err < 0) {
 		fprintf(stderr,
 			"Failed attaching egress BPF program on interface %s: %s\n",
-			config->ifname,
-			get_libbpf_strerror(config->egress_prog_id));
-		err = config->egress_prog_id;
+			config->ifname, get_libbpf_strerror(err));
 		goto egress_err;
 	}
 
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -153,6 +153,8 @@ struct pping_config {
 	char *event_map;
 	int ifindex;
 	struct xdp_program *xdp_prog;
+	int tcx_ingress_link_fd;
+	int tcx_egress_link_fd;
 	char ifname[IF_NAMESIZE];
 	char filename[PATH_MAX];
 	enum pping_output_format format;
@@ -598,7 +600,7 @@ static int xdp_detach(struct xdp_program *prog, int ifindex,
 }
 
 /*
- * Will attach program prog_name in obj to ifindex at attach_point.
+ * Will attach the program with fd prog_fd to ifindex at attach_point.
  * The ingress and egress programs share a single clsact qdisc (it carries both
  * hook points), so only the first attach for an interface should create it:
  * pass create_hook=true for that one and false for the other. This avoids
@@ -608,13 +610,11 @@ static int xdp_detach(struct xdp_program *prog, int ifindex,
  * if it created a new hook or not, and return the id of the attached program.
  * On failure it will return a negative error code.
  */
-static int tc_attach(struct bpf_object *obj, int ifindex,
-		     enum bpf_tc_attach_point attach_point,
-		     const char *prog_name, struct bpf_tc_opts *opts,
+static int tc_attach(int ifindex, enum bpf_tc_attach_point attach_point,
+		     int prog_fd, struct bpf_tc_opts *opts,
 		     bool create_hook, bool *new_hook)
 {
 	int err;
-	int prog_fd;
 	bool created_hook = false;
 	DECLARE_LIBBPF_OPTS(bpf_tc_hook, hook, .ifindex = ifindex,
 			    .attach_point = attach_point);
@@ -627,13 +627,6 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
 			return err;
 		else
 			created_hook = true;
-	}
-
-	prog_fd = bpf_program__fd(
-		bpf_object__find_program_by_name(obj, prog_name));
-	if (prog_fd < 0) {
-		err = prog_fd;
-		goto err_after_hook;
 	}
 
 	opts->prog_fd = prog_fd;
@@ -656,6 +649,64 @@ err_after_hook:
 		hook.attach_point = BPF_TC_INGRESS | BPF_TC_EGRESS;
 		bpf_tc_hook_destroy(&hook);
 	}
+	return err;
+}
+
+static int tcx_or_tc_attach(struct bpf_object *obj, struct pping_config *config,
+			    enum bpf_tc_attach_point attach_point)
+{
+	bool create_hook, new_hook = false;
+	enum bpf_attach_type link_type;
+	struct bpf_tc_opts *tc_opts;
+	const char *prog_name;
+	int *tcx_link_fd;
+	int prog_fd, err;
+
+	if (attach_point == BPF_TC_INGRESS) {
+		prog_name = config->ingress_prog;
+		tcx_link_fd = &config->tcx_ingress_link_fd;
+		tc_opts = &config->tc_ingress_opts;
+		link_type = BPF_TCX_INGRESS;
+		create_hook = true;
+	} else {
+		prog_name = config->egress_prog;
+		tcx_link_fd = &config->tcx_egress_link_fd;
+		tc_opts = &config->tc_egress_opts;
+		link_type = BPF_TCX_EGRESS;
+		/*
+		 * The ingress attachment creates the shared clsact qdisc,
+		 * so egress only creates it when ingress uses XDP.
+		 */
+		create_hook = strcmp(config->ingress_prog,
+				     PROG_INGRESS_XDP) == 0;
+	}
+
+	prog_fd = bpf_program__fd(
+		bpf_object__find_program_by_name(obj, prog_name));
+	if (prog_fd < 0)
+		return prog_fd;
+
+	err = bpf_link_create(prog_fd, config->ifindex, link_type, NULL);
+	if (err >= 0) {
+		*tcx_link_fd = err;
+		return err;
+	}
+	/*
+	 * Kernels without TCX support (pre-6.6) reject the link creation
+	 * with -EINVAL, fall back on the legacy clsact hook.
+	 */
+	if (err != -EINVAL)
+		return err;
+
+	err = tc_attach(
+		config->ifindex,
+		attach_point,
+		prog_fd,
+		tc_opts,
+		create_hook,
+		&new_hook);
+	if (new_hook)
+		config->created_tc_hook = true;
 	return err;
 }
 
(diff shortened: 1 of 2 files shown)


pping/pping.c
```diff
@@ -660,6 +662,63 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
 	return err;
 }
 
+/*
+ * Will attach program prog_name in obj to ifindex as a TCX link (available
+ * since kernel 6.6) and return the link fd. Closing the fd detaches the
+ * program again. Kernels without TCX support reject the link creation with
+ * -EINVAL.
+ */
```
tohojo: This comment just restates what the function does; please drop.

pping/pping.c
```diff
@@ -660,6 +662,63 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
...
+
+/*
+ * Will attach program prog_name in obj to ifindex at attach_point.
+ * When try_tcx is set, first attempts to attach the program as a TCX link,
+ * storing the link fd in *tcx_link_fd on success. When the kernel does not
+ * support TCX, or when try_tcx is false, attaches on the legacy clsact hook
+ * through tc_attach() instead, passing opts, create_hook and new_hook along.
+ * On success returns the TCX link fd or the legacy program id, on failure a
+ * negative error code.
+ */
```
tohojo: Same here

pping/pping.c
```diff
@@ -660,6 +662,63 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
...
+static int tcx_or_tc_attach(struct bpf_object *obj, int ifindex,
+			    enum bpf_tc_attach_point attach_point,
+			    const char *prog_name, struct bpf_tc_opts *opts,
+			    bool create_hook, bool *new_hook, bool try_tcx,
+			    int *tcx_link_fd)
+{
```
tohojo: With this many parameters I think it's better to just pass in the config struct and have the function pull out the members it needs.

pping/pping.c
```diff
@@ -660,6 +662,63 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
...
+		err = tcx_attach(obj, ifindex,
+				 attach_point == BPF_TC_INGRESS ?
+					 BPF_TCX_INGRESS :
+					 BPF_TCX_EGRESS,
+				 prog_name);
+		if (err >= 0) {
+			*tcx_link_fd = err;
+			return err;
+		}
+		if (err != -EINVAL)
```
tohojo: A comment here indicating that EINVAL means "not supported" might be appropriate.

pping/pping.c
```diff
@@ -660,6 +662,63 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
...
+ */
+static int tcx_or_tc_attach(struct bpf_object *obj, int ifindex,
+			    enum bpf_tc_attach_point attach_point,
+			    const char *prog_name, struct bpf_tc_opts *opts,
+			    bool create_hook, bool *new_hook, bool try_tcx,
+			    int *tcx_link_fd)
+{
+	int err;
+
+	if (try_tcx) {
```
tohojo: I don't think there's any way we could end up mixing TCX and TC (either TCX works and we'll succeed both times, or it doesn't and we'll fail twice; both are fine). So let's just drop this parameter.

pping/pping.c
```diff
@@ -2054,6 +2124,10 @@ static int load_attach_bpfprogs(struct bpf_object **obj,
 		detach_err = xdp_detach(config->xdp_prog, config->ifindex,
 					config->xdp_mode);
 		config->xdp_prog = NULL;
+	} else if (config->tcx_ingress_link_fd >= 0) {
+		/* Closing the TCX link fd detaches its program */
+		detach_err = close(config->tcx_ingress_link_fd) ? -errno : 0;
+		config->tcx_ingress_link_fd = -1;
```
tohojo: This detach branching is duplicated here and below (with slight differences). It's now  at the point where the complexity warrants factoring it into a helper function.

pping/pping.c
```diff
@@ -2788,13 +2864,17 @@ int main(int argc, char *argv[])
...
+		detach_err = close(config.tcx_egress_link_fd) ? -errno : 0;
+	} else {
+		/*
+		 * Remove the shared clsact qdisc if we created it, so it does
+		 * not linger and make the next run's create fail with -EEXIST.
+		 */
+		detach_err = tc_detach(config.ifindex, BPF_TC_EGRESS,
+				       &config.tc_egress_opts,
+				       config.created_tc_hook);
+	}
```
tohojo: We could have both directions of detach in that same helper function, there should be enough state in the config object to detect whether we need to detach both or not...

pping/pping.c
```diff
@@ -660,6 +662,63 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
...
+ */
+static int tcx_attach(struct bpf_object *obj, int ifindex,
+		      enum bpf_attach_type attach_type, const char *prog_name)
+{
+	int prog_fd;
+
+	prog_fd = bpf_program__fd(
+		bpf_object__find_program_by_name(obj, prog_name));
+	if (prog_fd < 0)
+		return prog_fd;
```
tohojo, line 687: This bit is identical between tc and tcx, pull it out to the wrapper.

tohojo (changes requested):

tobiaspal:
Pushed v2 addressing the review comments, and rebased onto the updated #144 branch (which now deprecates `--force`).

pping/pping.c
tobiaspal: Dropped — tcx_attach() is gone entirely: after hoisting the prog fd lookup into the wrapper, the remaining bpf_link_create() call moved there too.

pping/pping.c
tobiaspal: Dropped.

pping/pping.c
tobiaspal: Done — tcx_or_tc_attach() now takes (obj, config, attach_point) and pulls out the members it needs.

pping/pping.c
tobiaspal: Added.

pping/pping.c
tobiaspal: Dropped — both directions now try TCX first. On kernels without TCX the fallback notice prints once per direction (so twice with a tc ingress); say the word if you'd rather have it once or not at all.

pping/pping.c
tobiaspal: Done — detach_bpfprogs().

pping/pping.c
tobiaspal: Done — detach_bpfprogs(config) covers both directions for both callers (the attach error path and the shutdown cleanup). It keys off xdp_prog, the tcx link fds and tc_*_opts.prog_id to see what is attached, and destroys the clsact on the last legacy detach: normally the egress one, or the ingress one when egress never attached.

pping/pping.c
tobiaspal: Done — the wrapper looks up the fd once and passes it to both bpf_link_create() and tc_attach(), which now takes a prog_fd instead of obj + prog_name.

pping/pping.c
tohojo: Yeah, let's get rid of it. It's not really actionable for the user, and if the fallback doesn't work that will print a separate error...

tobiaspal: Removed.

pping/pping.c
```diff
@@ -659,6 +652,49 @@ static int tc_attach(struct bpf_object *obj, int ifindex,
 	return err;
 }
 
+static int tcx_or_tc_attach(struct bpf_object *obj, struct pping_config *config,
+			    enum bpf_tc_attach_point attach_point)
+{
+	bool ingress = attach_point == BPF_TC_INGRESS;
+	int *tcx_link_fd = ingress ? &config->tcx_ingress_link_fd :
+				     &config->tcx_egress_link_fd;
```
tohojo: All these ternary operator assignments combine to make this function harder to read than necessary. I think it's better to split up the egress/ingress assignments into one block at the top, something like:

```c
	bool create_hook, new_hook = false;
	enum bpf_attach_type link_type;
	struct bpf_tc_opts *tc_opts;
	const char *prog_name;
	int *tcx_link_fd;
	int prog_fd, err;

	if (attach_point == BPF_TC_INGRESS) {
		prog_name = config->ingress_prog;
		tcx_link_fd = &config->tcx_ingress_link_fd;
		tc_opts = &config->tc_ingress_opts;
		link_type = BPF_TCX_INGRESS;
		create_hook = true;
	} else {
		prog_name = config->egress_prog;
		tcx_link_fd = &config->tcx_egress_link_fd;
		tc_opts = &config->tc_egress_opts;
		link_type = BPF_TCX_EGRESS;
		create_hook = strcmp(config->ingress_prog, PROG_INGRESS_XDP) == 0;
      }
```

pping/pping.c
```diff
@@ -696,6 +732,55 @@ static int tc_detach(int ifindex, enum bpf_tc_attach_point attach_point,
...
+	} else if (config->tc_ingress_opts.prog_id) {
+		/*
+		 * Remove the shared clsact qdisc if we created it and no
+		 * egress detach follows, so it does not linger and make the
+		 * next run's create fail with -EEXIST.
+		 */
+		ingress_err = tc_detach(config->ifindex, BPF_TC_INGRESS,
+					&config->tc_ingress_opts,
+					config->created_tc_hook &&
+						!egress_attached);
```
tohojo: With these many-argument functions, please keep one argument per line, especially when one of them is a compound boolean like here.

pping/pping.c
```diff
@@ -696,6 +732,55 @@ static int tc_detach(int ifindex, enum bpf_tc_attach_point attach_point,
...
+			config->ifname, get_libbpf_strerror(ingress_err));
+
+	if (config->tcx_egress_link_fd >= 0) {
+		egress_err = close(config->tcx_egress_link_fd) ? -errno : 0;
+		config->tcx_egress_link_fd = -1;
+	} else if (config->tc_egress_opts.prog_id) {
+		/* Same here: remove the clsact if we created it */
+		egress_err = tc_detach(config->ifindex, BPF_TC_EGRESS,
+				       &config->tc_egress_opts,
+				       config->created_tc_hook);
```
tohojo: Same

pping/pping.c
```diff
@@ -696,6 +732,55 @@ static int tc_detach(int ifindex, enum bpf_tc_attach_point attach_point,
...
+		config->tc_egress_opts.prog_id = 0;
+	}
+	if (egress_err)
+		fprintf(stderr,
+			"Failed removing egress program from interface %s: %s\n",
+			config->ifname, get_libbpf_strerror(egress_err));
+
+	return ingress_err ? ingress_err : egress_err;
```
tohojo: This can just be `return ingress_err ?: egress_err;`

tohojo (changes requested):
A few more style nits, otherwise this looks good :)

pping/pping.c
tobiaspal: Done — easier to read this way.

pping/pping.c
tobiaspal: Done — also applied to the tc_attach call in tcx_or_tc_attach().

pping/pping.c
tobiaspal: Done.

pping/pping.c
tobiaspal: Done.

tohojo (approved):

PR #146, opened 2026-07-02
pping: fix errors in README, TODO and SAMPLING_DESIGN

tobiaspal:
Fixes for the pping documentation (README, TODO, SAMPLING_DESIGN). Three factual corrections:

- README described the packet capture as XDP on ingress and TC-BPF on egress, but the default ingress hook is tc and XDP is opt-in through `--ingress-hook`, so it now reads "XDP or TC-BPF (on ingress)".
- TODO listed "Use libxdp to load XDP program" as open; it is done (78b45bd), so it moved to the Done section.
- SAMPLING_DESIGN said the first packet with a matching identifier marks the timestamp as used "on egress"; matching happens on ingress.

The rest is spelling, grammar and markdown fixes (including a broken image tag and a misnumbered list) without changing the meaning. I also checked the numeric claims against the code (ICMP flow timeout, timestamp lifetime relative to sRTT, cleanup rate, per-CPU counter maps) and the external links; those are correct and untouched. One thing I did not touch: the `rec_bytes` value in the README's RTT-event example (37 bytes for 5922 received packets) looks off, but I cannot produce a verified replacement.

A few features are not covered by the documentation. I left them out to keep this PR to fixes — would you like any of them added in a follow-up?

- the `--aggregate` options
- the systemd-files/ and scripts/ directories
- TCX attachment (#145)

### DIFF
--- a/pping/README.md
+++ b/pping/README.md
@@ -1,7 +1,7 @@
 # PPing using XDP and TC-BPF
 A re-implementation of [Kathie Nichols' passive ping
-(pping)](https://github.com/pollere/pping) utility using XDP (on ingress) and
-TC-BPF (on egress) for the packet capture logic.
+(pping)](https://github.com/pollere/pping) utility using XDP or TC-BPF (on
+ingress) and TC-BPF (on egress) for the packet capture logic.
 
 ## Simple description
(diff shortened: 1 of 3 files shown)


tohojo (approved):
Looks good, thanks. And yeah, feel free to follow up with the other items :)

PR #148, opened 2026-07-03
pping: document aggregation, the systemd units and TCX attachment

tobiaspal:
The follow-up from #146: documents the `--aggregate` options (with examples captured on the PPPoE router from the previous PRs, IP prefixes replaced with documentation ranges), the systemd-files/ and scripts/ directories, TCX attachment in the pping.c file description, and the aggregation maps.

One thing that surfaced while writing it: `-a` disables all individual events (flow open/close/timeout as well as RTT), while its source comment says "Aggregated output currently disables individual RTT". The new section documents the actual behaviour.


tohojo (approved):
Looks good, thanks! :)

PR #149, opened 2026-07-03
pping: exit when the monitored interface is removed

tobiaspal:
When a PPPoE session re-establishes, the network interface is recreated. The kernel destroys the attached BPF programs together with the interface, and `pping` keeps running without receiving events.

With this PR `pping` subscribes to rtnetlink link messages through the existing epoll loop, and shuts down with an error when RTM_DELLINK arrives for the monitored ifindex: the shutdown happens as the interface disappears. A supervisor can then restart pping on the recreated interface; with systemd that is `Restart=on-failure`. If the netlink socket overflows (ENOBUFS), the handler falls back on comparing `if_nametoindex()` against the stored ifindex, so a lost RTM_DELLINK does not leave pping running blind. The detach step on shutdown is skipped when the interface is already gone, as its attachments no longer exist (a legacy tc detach would otherwise report a failure for programs the kernel already removed).

Tested on kernel 7.0, TCX attachment: pping attached to a veth, deleting the veth prints "Interface veth-pping was removed, shutting down" and exits with code 1 immediately, with no detach errors. On a PPPoE reconnect, systemd restarted `pping` after `RestartSec`, and the new instance attached to the recreated ppp interface.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -7,6 +7,7 @@ static const char *__doc__ =
 #include <linux/if_link.h>
 #include <net/if.h> // For if_nametoindex
 #include <net/if_arp.h> // For ARPHRD_PPP/ARPHRD_NONE (detecting L3 interfaces)
+#include <linux/rtnetlink.h> // For detecting when the interface is removed
 #include <sys/ioctl.h> // For ioctl()
 #include <linux/sockios.h> // For SIOCGIFHWADDR
 #include <sys/socket.h> // For socket()
@@ -59,9 +60,11 @@ static const char *__doc__ =
 #define PPING_EPEVENT_TYPE_SIGNAL (1ULL << 62)
 #define PPING_EPEVENT_TYPE_PIPE (1ULL << 61)
 #define PPING_EPEVENT_TYPE_AGGTIMER (1ULL << 60)
+#define PPING_EPEVENT_TYPE_NETLINK (1ULL << 59)
 #define PPING_EPEVENT_MASK                                                     \
 	(~(PPING_EPEVENT_TYPE_PERFBUF | PPING_EPEVENT_TYPE_SIGNAL |            \
-	   PPING_EPEVENT_TYPE_PIPE | PPING_EPEVENT_TYPE_AGGTIMER))
+	   PPING_EPEVENT_TYPE_PIPE | PPING_EPEVENT_TYPE_AGGTIMER |             \
+	   PPING_EPEVENT_TYPE_NETLINK))
 
 #define AGG_BATCH_SIZE 64 // Batch size for fetching aggregation maps (bpf_map_lookup_batch)
 
@@ -752,6 +755,26 @@ static int detach_bpfprogs(struct pping_config *config)
 	bool egress_attached = config->tcx_egress_link_fd >= 0 ||
 			       config->tc_egress_opts.prog_id;
 	int ingress_err = 0, egress_err = 0;
+
+	/*
+	 * If the interface is gone, so are its attachments - just release
+	 * the resources tracking them.
+	 */
+	if (if_nametoindex(config->ifname) != (unsigned int)config->ifindex) {
+		if (config->xdp_prog) {
+			xdp_program__close(config->xdp_prog);
+			config->xdp_prog = NULL;
+		}
+		if (config->tcx_ingress_link_fd >= 0) {
+			close(config->tcx_ingress_link_fd);
+			config->tcx_ingress_link_fd = -1;
+		}
+		if (config->tcx_egress_link_fd >= 0) {
+			close(config->tcx_egress_link_fd);
+			config->tcx_egress_link_fd = -1;
+		}
+		return 0;
+	}
 
 	if (config->xdp_prog) {
 		ingress_err = xdp_detach(config->xdp_prog, config->ifindex,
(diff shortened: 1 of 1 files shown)


tobiaspal:
This is the last PR of the series. With it, all the changes I made to `pping` are submitted. @tohojo thank you for your patience reviewing my vibe-coded changes. Your programming skills are way superior to Fable 5's (and to mine, of course).

I'm grateful to have contributed to such a cool project. If you'd like me to look at any of the TODO items or the open issues, I'm happy to help.

pping/pping.c
```diff
@@ -753,6 +756,10 @@ static int detach_bpfprogs(struct pping_config *config)
 			       config->tc_egress_opts.prog_id;
 	int ingress_err = 0, egress_err = 0;
 
+	/* If the interface is gone, so are its attachments */
+	if (if_nametoindex(config->ifname) != (unsigned int)config->ifindex)
+		return 0;
```
tohojo, line 776: The interface may be gone, but the resources we're tracking are not. Before returning this branch should clear out xdp_prog (with `xdp_program__close()`) and the link_fds if they exist.

tohojo (changes requested):
Cool idea; just one comment on the cleanup...

tohojo:
> This is the last PR of the series. With it, all the changes I made to `pping` are submitted. @tohojo thank you for your patience reviewing my vibe-coded changes. Your programming skills are way superior to Fable 5's (and to mine, of course).

Cool, thanks for taking the time to submit it all (and for massaging the LLM output into something usable :))!

> I'm grateful to have contributed to such a cool project. If you'd like me to look at any of the TODO items or the open issues, I'm happy to help.

I'll let @simosund chime in on that when he gets back from vacation. I'm about to head out too, so won't have any time to review things before I get back anyway :)

Out of curiosity, what are you doing with the data you're capturing from pping?

pping/pping.c
tobiaspal: Done.

tobiaspal:
> Out of curiosity, what are you doing with the data you're capturing from pping?

I'm trying to choose the "best" link for every connection.

In a multihomed network it's not trivial to choose which link to send a packet out on. Classic BGP does this with AS path length, which often picks a poor link. I could measure speeds to the most used networks and set up static routes, but that becomes outdated the moment a link goes down somewhere or an uplink gets saturated.

So the idea is: for every destination network, observe the latency over all uplinks and route future connections over the link with the lowest latency. Occasionally send a connection out on a "wrong" link to check if the latencies changed and the selection needs updating. If a link gets saturated its latency rises, another link becomes the "best", and new connections avoid the saturated one. ("Best" in quotes because there are many possible definitions; I'm simplifying and for my purposes best is the lowest latency.)

(With real BGP multihoming the return path is also a problem; at home I don't have BGP, the source IP determines the return path, so I'm sweeping that under the rug for now.)

`pping` gives me these per-network latency measurements from real traffic, passively. Could measure latency with `ping`, but ICMP is often put in a different QoS class; speed tests only measure their own probe traffic. No other solution is as good as `pping`.

tohojo (approved):

tohojo:
> > Out of curiosity, what are you doing with the data you're capturing from pping?
> 
> I'm trying to choose the "best" link for every connection.
> 
> In a multihomed network it's not trivial to choose which link to send a packet out on. Classic BGP does this with AS path length, which often picks a poor link. I could measure speeds to the most used networks and set up static routes, but that becomes outdated the moment a link goes down somewhere or an uplink gets saturated.
> 
> So the idea is: for every destination network, observe the latency over all uplinks and route future connections over the link with the lowest latency. Occasionally send a connection out on a "wrong" link to check if the latencies changed and the selection needs updating. If a link gets saturated its latency rises, another link becomes the "best", and new connections avoid the saturated one. ("Best" in quotes because there are many possible definitions; I'm simplifying and for my purposes best is the lowest latency.)
> 
> (With real BGP multihoming the return path is also a problem; at home I don't have BGP, the source IP determines the return path, so I'm sweeping that under the rug for now.)
> 
> `pping` gives me these per-network latency measurements from real traffic, passively. Could measure latency with `ping`, but ICMP is often put in a different QoS class; speed tests only measure their own probe traffic. No other solution is as good as `pping`.

Very cool! If you share the code you use to do this, we can definitely link it from here. As you note, this sounds like an excellent fit for pping's functionality :)

When you say "source IP determines the return path", that means you are NAT'ing things? So how do you avoid breaking connections in the middle if you switch uplink for a given prefix?

tobiaspal:
> When you say "source IP determines the return path", that means you are NAT'ing things? So how do you avoid breaking connections in the middle if you switch uplink for a given prefix?

Yes, at home have to NAT, don't have BGP at home yet.
Can't switch existing connections. Can choose the source IP only for new connections.
It's not really possible to have all connections on the best uplink. I'm aiming for "most" connections.

simosund:
>> I'm grateful to have contributed to such a cool project. If you'd like me to look at any of the TODO items or the open issues, I'm happy to help.
>
> I'll let @simosund chime in on that when he gets back from vacation. I'm about to head out too, so won't have any time to review things before I get back anyway :)

Hi, back from my vacation.

Thanks @tobiaspal for your contributions! Your use case sounds really cool, essentially supporting latency-based routing + load balancing for outgoing connections if I understand it correctly. As @tohojo mentioned, feel free to link to it if it's publicly available. Could potentially go under "Similar projects" (which should perhaps be called "Related projects") in the README.

As for working on issues or TODO items, feel welcome to address any item you find interesting or useful (or add other features/fixes as you've done here). As you may have noticed, I'm not actively working on this project myself anymore. However, I'm more than happy to discuss whatever ideas you may have for it.

simosund:
Related to the TODO list, I can also mention that the reason I never got around to adding support for the QUIC protocol is that turned out that virtually no QUIC clients set the spin bit. So while QUIC support would be cool, it would probably not give you very many RTT samples in the wild.

PR #151, opened 2026-08-01
pping: only print usage for invalid arguments

tobiaspal:
Follow-up to #149: when systemd restarts pping while the monitored interface is still gone, every failed attempt printed the full usage text into the journal after "Could not get index of interface X: No such device". Now only genuine argument errors print the usage text.

Verified on the router running this change:

```
# pping -i nonexistent0
Could not get index of interface nonexistent0: No such device
Failed parsing arguments: No such device
```

and an unknown option still prints the usage text.

LLM assistance was used for the investigation, the patch and this description.

### DIFF
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -2836,9 +2836,11 @@ int main(int argc, char *argv[])
 
 	err = parse_arguments(argc, argv, &config);
 	if (err) {
-		fprintf(stderr, "Failed parsing arguments:  %s\n",
+		fprintf(stderr, "Failed parsing arguments: %s\n",
 			get_libbpf_strerror(err));
-		print_usage(argv);
+		/* Usage text does not help with e.g. a missing interface */
+		if (err == -EINVAL)
+			print_usage(argv);
 		return EXIT_FAILURE;
 	}
 
--- a/pping/pping.c
+++ b/pping/pping.c
@@ -332,7 +332,7 @@ static int parse_arguments(int argc, char *argv[], struct pping_config *config)
 			len = strlen(optarg);
 			if (len >= IF_NAMESIZE) {
 				fprintf(stderr, "interface name too long\n");
-				return -EINVAL;
+				return -ENAMETOOLONG;
 			}
 			memcpy(config->ifname, optarg, len);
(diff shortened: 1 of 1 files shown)


pping/pping.c
```diff
@@ -2836,9 +2836,11 @@ int main(int argc, char *argv[])
 
 	err = parse_arguments(argc, argv, &config);
 	if (err) {
-		fprintf(stderr, "Failed parsing arguments:  %s\n",
+		fprintf(stderr, "Failed parsing arguments: %s\n",
 			get_libbpf_strerror(err));
-		print_usage(argv);
+		/* Usage text does not help with e.g. a missing interface */
+		if (err == -EINVAL)
```
simosund, line 2842: Hmm, as far as I can tell the only times `parse_arguments()` will currently return an error code other than `EINVAL` is either if the interface does not exist (the case mentioned by the PR/commit message) or if the filename given to `--write` is too long.

For both of these cases it seems fine to skip the usage message. However, the question is if we should also think about if there are other cases where it makes sense to skip the usage message (and change their error codes appropriately) while we're at it.

simosund (commented):
I would be fine with merging the PR as is (although I do not have the power to do so), as I do not see any case where it would currently be problematic.

However, before this PR the returned error code from `parse_arguments()` did not matter much (only affected the printed error after "Failed parsing arguments"), and I clearly did not give too much thought about what error code to return for different types of errors. Now that we tie the usage message to `EINVAL` specifically, we should perhaps consider what error code to return from `parse_arguments()` more carefully.

simosund:
From my understanding, @tohojo is now on vacation, so just be aware that it may take some time before your current PR's can be addressed.

tobiaspal:
Good point. Surveying parse_arguments(): everything else returns -EINVAL — unknown option, missing -i, out-of-range or unparseable option values (those already print a specific message first), and invalid enum values. For all of those the usage text seems reasonable.

One inconsistency found: an over-long interface name returned -EINVAL while an over-long --write filename returns -ENAMETOOLONG. Pushed a commit aligning the interface-name case to -ENAMETOOLONG, so the usage text no longer prints there either.

